PR #2, opened 2021-02-22
Add automation for JupyterHub notebook spawn and shutdown

LaVLaS:
Adds support for
* JupyterHub spawner web ui
* notebook startup/shutdown
* Debug breakpoints


vasukulkarni:
Thanks @LaVLaS

PR #14, opened 2021-03-10
Feature/odh dashboard

LaVLaS:
* Add the robot keywords for navigating the ODH dashboard
* Add command line switches to the `run_robot_test.sh` script to configure robot options

### DIFF
--- a/run_robot_test.sh
+++ b/run_robot_test.sh
@@ -1,6 +1,34 @@
 #/bin/bash
 
-if [[ ! -f "test-variables.yml" ]]; then
+TEST_CASE_FILE=tests/Tests/test.robot
+TEST_VARIABLES_FILE=test-variables.yml
+TEST_VARIABLES=""
+
+while [ "$#" -gt 0 ]; do
+  case $1 in
+    --test-variable)
+      shift
+      TEST_VARIABLES="${TEST_VARIABLES} --variable $1"
+      shift
+      ;;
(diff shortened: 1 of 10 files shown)


run_robot_test.sh
```diff
@@ -41,6 +69,6 @@ ${VENV_ROOT}/bin/pip install -r requirements.txt
 tmp_dir=$(mktemp -d -t ods-ci-$(date +%Y-%m-%d-%H-%M)-XXXXXXXXXX)
 #TODO: Configure the "tmp_dir" creation so that we can have a "latest" link
 mkdir $tmp_dir
-./venv/bin/robot -d $tmp_dir -x xunit_test_result.xml -r test_report.html --variablefile test-variables.yml tests/Tests/test.robot
+./venv/bin/robot -d $tmp_dir -x xunit_test_result.xml -r test_report.html ${TEST_VARIABLES} --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE}
```
vasukulkarni: can we also keep the original --variablefile option(via if/else)? This helps running the test locally or else we have to set it in cli everytime to run.

LaVLaS: The previous functionality is still preserved based on the default values for each of the new bash variables. If the is no cli override then the script will default to `--variablefile test-variables.yml tests/Tests/test.robot` once the bash variables are processed.

PR #32, opened 2021-04-22
Remove chromedriver from the repo

LaVLaS:


vasukulkarni:
Just waiting for our jenkins worker nodes to catchup with the changes so that sanity runs are not affected.

LaVLaS:
I've updated the PSI nodes so that they have the correct `chromedriver` installed

PR #37, opened 2021-05-11
Add Dockerfile for running framework in a container

LaVLaS:
Add a dockerfile for running the robot framework in a web browser

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,0 +1,26 @@
+FROM centos:centos8
+
+# Use this build arg to set any default test script arguments
+ARG RUN_SCRIPT_ARGS=
+ENV RUN_SCRIPT_ARGS=${RUN_SCRIPT_ARGS}
+
+ENV HOME /tmp
+
+RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm &&\
+    dnf install -y git python3 unzip chromium chromedriver &&\
+    dnf clean all
+
+RUN mkdir $HOME/ods-ci
+# Change the WORKDIR so the run script references any files/folders from the root of the repo
+WORKDIR $HOME/ods-ci
+
+COPY tests tests/
+COPY requirements.txt run_robot_test.sh setup.py .
+RUN python3 -m venv venv && source venv/bin/activate && venv/bin/pip3 install -r requirements.txt
+
+# Set the group ownership so non-root users can write to /tmp
+RUN chgrp -R 0 /tmp && chmod -R g=u /tmp
+
+USER 1001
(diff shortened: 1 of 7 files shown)


build/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM centos:centos8
+
+ARG CHROMEDRIVER_VER=90.0.4430.24
```
lugi0, line 3: I think doing something like this would be better:

```
RUN CHROMEVER=$(google-chrome-stable --product-version | grep -o "[^\.]*\.[^\.]*\.[^\.]*") && \
         DRIVERVER=$(curl -s "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$CHROMEVER") && \
         wget -q --continue -P /chromedriver "http://chromedriver.storage.googleapis.com/$DRIVERVER/chromedriver_linux64.zip" && \
         unzip /chromedriver/chromedriver* -d /chromedriver
```

jgarciao: chromedriver 89.x.x.x seems to be [present in the centos epel repository,](https://pkgs.org/search/?q=chromedriver) so if this version suits your needs you could just run `dnf install chromedriver`.

LaVLaS: That would simplify the setup but we would have to switch to the `chromium-browser` to use the `chromedriver` package in the epel repo since it appears that those two versions are synchronized for updates.

jgarciao: chrome and chromium are mostly the same so if it simplifies our testing setup I'll go for it if nobody have any concern

LaVLaS:
While this ran without issue with `podman run`, I encountered permission errors when running this on OpenShift.  I'm going to modify the Dockerfile to use a ubi8 python base image and properly support running as a non-root user.  

After our offline discussion, I don't think we need the ability to `git clone` the repo and can just `COPY` the required repo files (run script, tests/, requirements, ...) into image at build time

build/Dockerfile
LaVLaS: Updated to pull `chromium-browser` & `chromedriver` from the `epel` repo

LaVLaS:
> While this ran without issue with `podman run`, I encountered permission errors when running this on OpenShift. I'm going to modify the Dockerfile to use a ubi8 python base image and properly support running as a non-root user.
> 
> After our offline discussion, I don't think we need the ability to `git clone` the repo and can just `COPY` the required repo files (run script, tests/, requirements, ...) into image at build time

Permissions issues on OCP should be resolved with the cento8 image by setting the appropriate permissions and `USER` uid.

vasukulkarni (approved):

PR #41, opened 2021-05-12
Image iteration in single test case

lugi0:

@vasukulkarni is this what you had in mind when you said you wanted to iterate over all available images in one test case? I can't think of another way to achieve this, but I don't particularly like this execution either.
Pinging @LaVLaS as well for any potential ideas on how to improve this.

### DIFF
--- a/tests/Tests/image-iteration.robot
+++ b/tests/Tests/image-iteration.robot
@@ -1,0 +1,23 @@
+*** Settings ***
+Resource         ../Resources/ODS.robot
+Resource         ../Resources/Common.robot
+Library          DebugLibrary
+Library          JupyterLibrary
+Suite Setup      Begin Web Test
+Suite Teardown   End Web Test
+
+*** Variables ***
+@{IMAGES}  s2i-minimal-notebook  s2i-generic-data-science-notebook
+
+
+*** Test Cases ***
+Open ODH Dashboard
+  [Tags]  Sanity
+  Login To ODH Dashboard  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
+  Wait for ODH Dashboard to Load
+
+Iterative Testing
+  [Tags]  Sanity
+  FOR  ${image}  IN  @{IMAGES}
+    Iterative Image Test  ${image}
+  END
--- a/tests/Tests/image-iteration.robot
+++ b/tests/Tests/image-iteration.robot
@@ -7,7 +7,7 @@ Suite Setup      Begin Web Test
 Suite Teardown   End Web Test
 
 *** Variables ***
-@{IMAGES}  s2i-minimal-notebook  s2i-generic-data-science-notebook
+@{IMAGES}  s2i-generic-data-science-notebook  s2i-minimal-notebook
 
 
 *** Test Cases ***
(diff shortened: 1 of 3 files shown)


tests/Tests/image-iteration.robot
```diff
@@ -0,0 +1,23 @@
...
+
+*** Test Cases ***
+Open ODH Dashboard
+  [Tags]  Sanity
+  Login To ODH Dashboard  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
+  Wait for ODH Dashboard to Load
+
+Iterative Testing
+  [Tags]  Sanity
+  FOR  ${image}  IN  @{IMAGES}
```
vasukulkarni, line 21: Absolutely this is what I had in mind!, Neat and Clean! there will be problems with polarion ids and we can think of how to fix that.

LaVLaS:
> Signed-off-by: Luca Giorgi [lgiorgi@redhat.com](mailto:lgiorgi@redhat.com)
> 
> @vasukulkarni is this what you had in mind when you said you wanted to iterate over all available images in one test case? I can't think of another way to achieve this, but I don't particularly like this execution either.
> Pinging @LaVLaS as well for any potential ideas on how to improve this.

Yes. We can use the [Test Case Templates](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#test-templates) that are provided by the Robot Framework to standardize a test case to to spawn a notebook, clone repo, run a notebook and check for errors

lugi0:
@LaVLaS  is this what you meant with the test template? I don't quite see the advantage vs. using the Keyword in the loop.

LaVLaS:
> @LaVLaS is this what you meant with the test template? I don't quite see the advantage vs. using the Keyword in the loop.

You are correct that it is  for this specific use case. The only difference is the presentation of the test case and the data that it iterates over.  This would be the same test without the for loop iteration over the `{IMAGES}` array

```robot
*** Test Case ***
Iterative Testing
  [Template]  Iterative Image Test
  s2i-generic-data-science-notebook  https://github.com/lugi0/minimal-nb-image-test   minimal-nb-image-test/minimal-nb.ipynb
  s2i-minimal-notebook  https://github.com/lugi0/minimal-nb-image-test   minimal-nb-image-test/minimal-nb.ipynb

  # You can extend this out for unique repos/notebook for each notebook image
  s2i-minimal-notebook  https://github.com/foobar/nb-image-test   minimal-image-test/pip-install-generic-nb.ipynb
  s2i-generic-data-science-notebook  https://github.com/foobar/nb-image-test   generic-nb-image-test/generic-nb.ipynb
  s2i-generic-data-science-notebook  https://github.com/foobar/nb-image-test   generic-nb-image-test/pip-install-tf-nb.ipynb
  s2i-tensorflow-notebook  https://github.com/foobar/nb-image-test  some-tf-test/foobar-tf.ipynb
```

The choice comes down to the standard you want to enforce for these types of test cases

lugi0:
@vasukulkarni @LaVLaS We are now using a list of lists for the workloads to run, and since I had it ready I also added the new repository with the clustering workloads. I'd prefer this to be reviewed/merged before we add too much functionality on top of it.

tests/Tests/image-iteration.robot
```diff
@@ -0,0 +1,69 @@
...
+
+*** Test Cases ***
+Open ODH Dashboard
+  [Tags]  Sanity
+  Login To ODH Dashboard  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
+  Wait for ODH Dashboard to Load
+
+Iterative Testing
+  [Template]  Iterative Image Test
+  [Tags]  Sanity
```
vasukulkarni, line 30: I think i have missed something here, I was thinking lines: 28 to 33 will be repeated with different tags.

lugi0: There's no need to do that if we take the "list of lists" approach for the image/repo combo.
We CAN still do it if we want to have the different tags, but then the list of lists makes no sense and we need to change it.

lugi0:
@vasukulkarni @LaVLaS Switched over to a dictionary in order to be able to replicate the test case multiple times with different tags for future Polarion integration

vasukulkarni (approved):

PR #42, opened 2021-05-13
WIP: Refactor/test hierarchy

lugi0:
First try at refactoring the folder structure.
I have used 00/10 etc. so that we have space to slot new folders/files in the middle if need be.
The placeholder files are there just to be able to track the new (otherwise empty) folders.


jgarciao:
Sorry Luca, I should have told you before. I think the folder 50_Billing it's not necessary because billing will be tested inside 40_Metrics

jgarciao:
You can merge it like this and I'll remove the 50__Billing folder when I add my tests

lugi0:
I have used the folders @vasukulkarni proposed in the Jira issue. Regardless, I think it could make sense to split metrics from billing, there will likely be some metrics (i.e. internal ones for SREs) that won't impact billing, right?

PR #55, opened 2021-06-14
WIP: Add Polarion IDs for Sanity tests

lugi0:

### DIFF
--- a/tests/Tests/image-iteration.robot
+++ b/tests/Tests/image-iteration.robot
@@ -24,14 +24,14 @@ Open ODH Dashboard
   Wait for ODH Dashboard to Load
 
 Iterative Testing Classifiers
-  [Tags]  Sanity  POLARION-ID-Classifiers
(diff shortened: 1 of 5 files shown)


aloganat (commented):
If we have same polarion id tagged to more than one test scripts, there is a chance of overriding the test result. For example, if test1 with ods-1 tag failed, and if test2 with same ods-1 tag passed.. test2 result will overwrite the test1 result.  
I think we have to restructure in such a way that the test steps that is common across multiple scripts should be written in a library(or keyword) and call them in each script.

lugi0:
After talking with Arthy, I believe that in order to have unique Polarion IDs across all test cases/test suites, we will need to:

- Refactor the whole code base in ods-ci (Each test case that is repeated more than once should become a keyword. It has to be run once by itself with the correct Polarion ID as a tag, and used as a keyword without an explicit tag in all other cases)
- Restructure the test cases in Polarion to not only track authored test cases, but also Robot Framework test cases; this will enable us to track failures/successes of test cases that only rely on other keywords and wouldn't have their own Polarion IDs, e.g. the two Iterative test cases.

vasukulkarni:
Closing based on discussion.

PR #65, opened 2021-07-09
Long running test

lugi0:
This is still somewhat flaky and more work needs to be done, but I'd like some preliminary reviews on it.

### DIFF
--- a/tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
+++ b/tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
@@ -18,6 +18,7 @@ Launch JupyterLab
   Run Keyword If  ${authorization_required}  Authorize jupyterhub service account
   Wait Until Page Contains Element  xpath://span[@id='jupyterhub-logo']
   Select Notebook Image  s2i-generic-data-science-notebook
+  Select Container Size  Small
   Sleep  1
   ${ID} =  Spawner Environment Variable Exists  AWS_ACCESS_KEY_ID
   ${PW} =  Spawner Environment Variable Exists  AWS_SECRET_ACCESS_KEY
@@ -43,6 +44,8 @@ Long Running Test Case
   Run Repo and Clean  https://github.com/lugi0/clustering-notebook  clustering-notebook/CCFraud-clustering.ipynb
   Run Repo and Clean  https://github.com/lugi0/clustering-notebook  clustering-notebook/customer-segmentation-k-means-analysis.ipynb
   Run Repo and Clean  https://github.com/lugi0/clustering-notebook  clustering-notebook/CCFraud-clustering-S3.ipynb
+  Run Repo and Clean  https://github.com/red-hat-data-services/notebook-benchmarks  notebook-benchmarks/pytorch/PyTorch-MNIST-Minimal.ipynb
+  Run Repo and Clean  https://github.com/red-hat-data-services/notebook-benchmarks  notebook-benchmarks/tensorflow/TensorFlow-MNIST-Minimal.ipynb 
 
 *** Keywords ***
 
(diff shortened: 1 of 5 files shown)


tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
```diff
@@ -0,0 +1,76 @@
...
+
+  ${kernel_or_server_restarting} =  Run Keyword And Return Status  Page Should Not Contain Element  xpath:/html/body/div[8]/div/div[2]
+  IF  ${kernel_or_server_restarting} == False
+    ${is_server_down} =  Run Keyword And Return Status  Page Should Not Contain Element  xpath:/html/body/div[8]/div/div[2]/button[2]
+    IF  ${is_server_down} == False
+        Click Button  xpath:/html/body/div[8]/div/div[2]/button[2]
+    ELSE
+        Click Button  xpath:/html/body/div[8]/div/div[2]/button
+    END
+  END
```
LaVLaS, line 66: There is no expectation that the "kernel is restarting" is standard behavior for a normal notebook so I would recommend not including a workaround for it in a `Sanity` test. 

For this long running test and without looking at the logs, this issue is possibly the pod failing due to lack of resources, OOMKilled.  You should select a non-Defualt notebook pod size in the Spawner UI for this test since the `Default` size has a limit of 1 CPU & 2Gi of memory

lugi0: Hey @LaVLaS , I agree with you that this is non-standard behaviour and possibly due to resource limits, however without this workaround the test case will never pass. 
From what I have been able to see, the kernel seems to be restarting when we approach memory limits (although I've never seen the usage go above or even near the 2Gi). The pod doesn't actually fail during these runs, we only get the popups in Jupyter and as long as we dismiss them we can keep executing notebooks. 

In the pod logs, the only thing that I've been able to find is:
```
[I 2021-07-08 15:36:32.583 SingleUserLabApp restarter:68] AsyncIOLoopKernelRestarter: restarting kernel (1/5), keep random ports
kernel d2f42578-2ef7-4e6e-9937-ef15234a7754 restarted
[I 2021-07-08 15:36:32.712 SingleUserLabApp kernelmanager:268] Starting buffering for d2f42578-2ef7-4e6e-9937-ef15234a7754:e8b2a517-466a-4132-928a-bda63d33c0d9
[I 2021-07-08 15:36:33.051 SingleUserLabApp log:181] 101 GET /user/htpasswd-admin1/api/kernels/d2f42578-2ef7-4e6e-9937-ef15234a7754/channels?session_id=e8b2a517-466a-4132-928a-bda63d33c0d9 (htpasswd-admin1@::ffff:10.130.2.39) 3.08ms
[I 2021-07-08 15:36:33.052 SingleUserLabApp handlers:371] Restoring connection for d2f42578-2ef7-4e6e-9937-ef15234a7754:e8b2a517-466a-4132-928a-bda63d33c0d9
```

All of this being said, I do agree that we need to test this PR in a bigger pod size, however both on PSI and AWS this is currently not possible.

jgarciao: Luca, it's possible to easily add and delete machines to existing ODS clusters, and they can be more powerful.

For example, right now rhods-aws-qe has 3 nodes with 4 vCPU 16 GiB RAM (m5.xlarge) but we could temporally add an extra node with 8 vCPU 32 GiB RAM (m5.2xlarge)

lugi0: Tested this in a medium sized nb server (thanks @jgarciao) without this kernel restarting workaround and the test case passed twice without a problem. 
Still, we either provision a cluster that can always spawn a medium sized server or we keep the workaround in, otherwise this test case will always be flaky.
@LaVLaS @vasukulkarni

lugi0:
I believe this test is now finalized, using a Small server we don't encounter the kernel issue anymore.
Added the notebooks suggested by @LaVLaS as additional workloads and now the run time in my PSI cluster is ~20 minutes give or take a few seconds.

I will need to open a PR to add warning suppression in those notebooks, otherwise our own tests gets tripped up by the warning and thinks that the execution has failed at some point. I would hold off on merging until then.

@vasukulkarni

vasukulkarni:
LGTM in general

vasukulkarni (approved):

lugi0:
Updated to pull notebooks from my own fork
PR test will be at https://opendatascience-jenkins-csb-rhods.apps.ocp4.prod.psi.redhat.com/job/rhods-ci-pr-test/78/

lugi0:
Unexpected failures in previous run, started another one: https://opendatascience-jenkins-csb-rhods.apps.ocp4.prod.psi.redhat.com/job/rhods-ci-pr-test/79/

vasukulkarni:
There are quite a bit of failed test cases in the PR , need to check for infra or test issues.

lugi0:
Latest Jenkins run at https://opendatascience-jenkins-csb-rhods.apps.ocp4.prod.psi.redhat.com/job/rhods-ci-pr-test/90/

PR #73, opened 2021-08-10
Verify services available in ODS Dashboard

jgarciao:
* Adds keywords to check if services are available in ODS Dashboard in the Explore Page
* Adds keywords to check if services are enabled
* Adds one test case to check RHOSAK using this keywords


jgarciao:
This is being tested right now in https://opendatascience-jenkins-csb-rhods.apps.ocp4.prod.psi.redhat.com/job/rhods-ci-pr-test/106/

jgarciao:
Testing again at https://opendatascience-jenkins-csb-rhods.apps.ocp4.prod.psi.redhat.com/job/rhods-ci-pr-test/107/

pablofelix:
The conflict you are having with Menu keywords is because the qualified name only works with libraries and not with just any file name. The Menu.robot file contains keywords that can be used with any menu. It is done  this way to add one more level of abstraction in order to reuse code.  So to solve the conflict we can use those keywords as the base and wrap them with specific keywords for each menu that uses them, as this was their purpose, and maybe changing the Menu.robot file to another folder. This way both the new tests and the existing tests will work.

jgarciao:
In the end I solved the conflict adding the filename to the keyword calls (for example **Menu.Navigate To Page**, or **ODHDashboard.Navigate To Page**)

As you said, Menu.robot can work for OpenShift Console and RHODS Dashboard because they are similar applications. 

We could move this file to  Page/Common/Menu.robot and share the code. On the other side, this Menu.robot won't have any utility for other applications, like JupyterHub, Prometheus, ...

I think that as the file is only a few lines of code and it's quite simple I will leave duplicated so each app can have its own page object dedicated to menu.

jgarciao:
Following Pablo's suggestion, I've moved the Menu.robot file to Components/Menu.robot  and I have removed the now not necessary ODHDashboardMenu

A new OCPMenu.robot file is created to define specific keywords for OpenShift's console

jgarciao:
Don't merge it yet. I have still to add tests for OpenVino and Seldon and run all in jenkins again

tests/Resources/Page/OCPDashboard/OperatorHub/InstallODH.robot
```diff
@@ -1,10 +1,12 @@
...
+Resource   ../../Components/Components.resource
+Resource   ../../LoginPage.robot
+Library    SeleniumLibrary
 
 *** Keywords ***
 Open OperatorHub
-    Open OCP Console 
-    Login To Openshift
+    Open OCP Console
+    LoginPage.Login To Openshift  ${OCP_ADMIN_USER.USERNAME}  ${OCP_ADMIN_USER.PASSWORD}  ${OCP_ADMIN_USER.AUTH_TYPE}
```
vasukulkarni, line 9: Doesn't this modify for all Installs?

jgarciao:
Running here right now: https://opendatascience-jenkins-csb-rhods.apps.ocp4.prod.psi.redhat.com/job/rhods-ci-pr-test/118

lugi0:
@jgarciao this seems to be the same bug we spoke about earlier. It should be fixed if we move the 
`Go To ${ODH_DASHBOARD_URL}`
keyword at the beginning rather than end of the test

jgarciao:
I tried the change locally but still had problems and the test didn't pass. I think this errors should be handled in another PR

PR #92, opened 2021-09-15
Create a test for ODS-536

tmckayus:
This test launches a flask server in notebook and verifies that
it can be curled from another notebook.


tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
```diff
@@ -0,0 +1,86 @@
...
+  Sleep  10
+
+  # Run the curl command
+  Open With JupyterLab Menu  File  Open from Path…
+  Input Text  //div[.="Open Path"]/../div[contains(@class, "jp-Dialog-body")]//input  flask-notebook/curl_local.ipynb
+  Click Element  xpath://div[.="Open"]
+  Wait Until curl_local.ipynb JupyterLab Tab Is Selected
+  Open With JupyterLab Menu  Run  Run All Cells
+  Wait Until JupyterLab Code Cell Is Not Active
+  ${output} =  Get Text  (//div[contains(@class,"jp-OutputArea-output")])
```
tmckayus, line 75: This doesn't work. I can't seem to get the framework to read the output from curl in the second notebook, although it displays on screen. Need some help!

tmckayus: It just returns an empty string

tmckayus:
@LaVLaS I could use a hand here in getting that final output read and compare to work. It might be confused because there are multiple notebooks running? idk.

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
lugi0: Seems like you are missing the [last()] at the end of the xpath you get text from, this way I think it will get the output from the first code cell in the nb (which might not be the correct one)

lugi0: Since it is a one liner, you could also do 
`Run Cell And Check Output  !curl localhost:5000  {EXPECTED-OUTPUT}`

tmckayus: @lugi0 I'll retry  and confirm, I believe I tried both of those things.  Both returned empty string.  I'll check again ....

LaVLaS: tl;dr For this specific instance add `[last()]` to this line.  

`//div[contains(@class,"jp-OutputArea-output")]` is returning an array since multiple elements match across all open notebooks.  `Get Text` is returning the visible text for the first element which is the hidden notebook on the other tab which causing the empty string

tmckayus: I added [last()], the result is the same.

What can replace [last()] ? Is there a way to loop over output results, maybe it's in there somewhere?

tmckayus:
@LaVLaS @lugi0 
Turns out this was failing because a newline on the flask server output is required for robot to recognize the output.
So the code here is correct currently.

I'll make sure none of the sleeps are necessary and update.

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
```diff
@@ -0,0 +1,85 @@
...
+  Click Element  xpath://div[.="CLONE"]
+
+  # Run the flask server
+  Open With JupyterLab Menu  File  Open from Path…
+  Input Text  //div[.="Open Path"]/../div[contains(@class, "jp-Dialog-body")]//input  flask-notebook/run_flask.ipynb
+  Click Element  xpath://div[.="Open"]
+  Wait Until run_flask.ipynb JupyterLab Tab Is Selected
+  Close Other JupyterLab Tabs
+  Open With JupyterLab Menu  Run  Run All Cells
+  Sleep  15
```
tmckayus, line 66: Sleep is necessary because the server never ends, but it needs time to get up and running

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
```diff
@@ -0,0 +1,85 @@
...
+
+  # Run the curl command
+  Open With JupyterLab Menu  File  Open from Path…
+  Input Text  //div[.="Open Path"]/../div[contains(@class, "jp-Dialog-body")]//input  flask-notebook/curl_local.ipynb
+  Click Element  xpath://div[.="Open"]
+  Wait Until curl_local.ipynb JupyterLab Tab Is Selected
+  Open With JupyterLab Menu  Run  Run All Cells
+  Sleep  15
```
tmckayus, line 74: Sleep is necessary because "Wait Until JupyterLab Code Cell Is Not Active" doesn't seem to work here. Without the sleep, robot goes immediately to gathering input.

run_robot_test.sh
```diff
@@ -104,6 +104,6 @@ fi
 TEST_ARTIFACT_DIR=$(mktemp -d -p ${TEST_ARTIFACT_DIR} -t ods-ci-$(date +%Y-%m-%d-%H-%M)-XXXXXXXXXX)
 
 #run tests
-./venv/bin/robot -d ${TEST_ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html ${TEST_VARIABLES} --variablefile ${TEST_VARIABLES_FILE} --exclude TBC ${EXTRA_ROBOT_ARGS} ${TEST_CASE_FILE}
+./venv/bin/robot --loglevel DEBUG -d ${TEST_ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html ${TEST_VARIABLES} --variablefile ${TEST_VARIABLES_FILE} --exclude TBC ${EXTRA_ROBOT_ARGS} ${TEST_CASE_FILE}
```
tmckayus, line 107: oops, I'll remove this

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
pablofelix: I didn't try but maybe  'wait until keyword succeeds' could avoid the sleep keyword http://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Wait%20Until%20Keyword%20Succeeds

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
```diff
@@ -0,0 +1,85 @@
...
+  Should Match  ${output}  Hello World!
+  Close JupyterLab Selected Tab
+
+  # Shutdown flask server
+  Wait Until run_flask.ipynb JupyterLab Tab Is Selected
+  Open With JupyterLab Menu  Kernel  Interrupt Kernel
+  Wait Until JupyterLab Code Cell Is Not Active
+
+  Close JupyterLab Selected Tab
+  Logout JupyterLab
```
lugi0, line 85: This should not be needed since the Teardown will close the server

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
```diff
@@ -0,0 +1,85 @@
...
+Resource         ../../Resources/ODS.robot
+Resource         ../../Resources/Common.robot
+Resource         ../../Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+Library          DebugLibrary
+Suite Setup      Begin Web Test
+Suite Teardown   End Web Test
+
+Resource         ../../Resources/ODS.robot
+Resource         ../../Resources/Common.robot
+Library          DebugLibrary
```
lugi0, line 11: Duplicate imports

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
```diff
@@ -0,0 +1,85 @@
...
+Library          JupyterLibrary
+
+
+*** Variables ***
+
+
+*** Test Cases ***
+Open RHODS Dashboard
+  [Tags]  Sanity
```
lugi0, line 20: Is this test case supposed to be in Sanity? @vasukulkarni

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
LaVLaS: Can you add that comment in the code for future reference?

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
LaVLaS: Can you add that comment in the code for future reference?

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
tmckayus: I tried removing pieces of the shutdown code, anything less than this for some reason seems prone to failures in the teardown suite.  The screen never switches from displaying the activated shutdown button to having only "Start My Server".

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
tmckayus: @lugi0 I changed the tags all to "Tier2"

vasukulkarni:
Jenkins issue: https://opendatascience-jenkins-csb-rhods.apps.ocp4.prod.psi.redhat.com/job/rhods-ci-pr-test/159/console

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
lugi0: The only thing I'm not comfortable with is that Logout, because the Teardown then will not be able to go to the Hub Control Panel to shut down the server once you've logged out, so it will just close the browser without stopping the running server itself. If the test runs alone it's fine, but when there are multiple test cases being run in sequence this might cause some weird issues. Could you try removing it and checking if it still works?

tmckayus: okay I removed the logout. The trouble is that the shutdown procedure seems to have trouble pressing the "stop server" button so the server never shuts down (it seems to if I press it myself for the robot). I'll keep playing.

tests/Tests/500__jupyterhub/test-jupyterlab-flask-notebook.robot
```diff
@@ -0,0 +1,83 @@
...
+Library          JupyterLibrary
+Suite Setup      Begin Web Test
+Suite Teardown   End Web Test
+
+*** Variables ***
+
+
+*** Test Cases ***
+Open RHODS Dashboard
+  [Tags]  Tier2
```
vasukulkarni, line 15: @lugi0 Please add Polarion link later when we get to it.

vasukulkarni (approved):

PR #132, opened 2021-11-29
Added Keyword in Teardown to reset Setting for ACe using OpenShiftCLi

tarukumar:
Signed-off-by: Tarun <takumar@redhat.com>

### DIFF
--- a/tests/Resources/Page/OCPDashboard/Builds/Builds.robot
+++ b/tests/Resources/Page/OCPDashboard/Builds/Builds.robot
@@ -1,4 +1,5 @@
 *** Settings ***
+Library    OpenShiftCLI
 Resource   ../../OCPDashboard/Page.robot
 Resource   ../../ODH/ODHDashboard/ODHDashboard.robot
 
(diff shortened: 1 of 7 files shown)


tarukumar:
Verified using Jenkins CI [PR-216](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/216). Multiple TCs failed due to conflict with Get keyword  between Request and Openshiftcli library, Python version which is not related above code changes

bdattoma:
it looks good to me.
Only one thought about something that could be worth to refactor. It seems to me that there is the possibility of merging all the keywords for deleting and check existence of Anaconda objects in two common keywords with a couple of arguments. This would avoid code redundancy. What do you think?

bdattoma (dismissed):

tarukumar:
> 



> 

I have thought about it.There s a catch Some keyword check based on the name and some based on label selector.That's why i prepare separably  for each object. I understand fro where you are making the point for refactor , it absolutely make sense for longer run. May be what i can do is create a common key word  for check and have the if block inside it based on input, what do you think?

bdattoma:
> I have thought about it.There is a catch Some keyword check based on the name and some based on label selector.That's why i prepare separably for each object. I understand fro where you are making the point for refactor , it absolutely make sense for longer run. May be what i can do is create a common key word for check and have the if block inside it based on input, what do you think?

there could be different solutions. You could create two different common keywords which differs for selector type (still, there's redundancy but reduced), or you can use branching inside the keyword based on which argument has been passed, or maybe there is a way to pass arguments like *args and **kwargs and dynamically use the "name" or "label selector". 
For me, we could handle this improvement in another PR after this one is merged

tarukumar:
> 

what i meant was same like below example
Screenshot from 2021-12-07 22-17-39
.

lugi0:
>maybe there is a way to pass arguments like *args and **kwargs and dynamically use the "name" or "label selector".

It should definitely be possible to use kwargs and use the selectors dinamically.

>For me, we could handle this improvement in another PR after this one is merged

Agree with this, the PR is already meaty as it is

lugi0 (dismissed):

bdattoma (approved):

lugi0 (approved):
lgtm

PR #137, opened 2021-11-30
version compare keyword

lugi0:


setup.py
```diff
@@ -29,6 +29,7 @@
         "requests",
         "robotframework-requests",
         "escapism",
+        "semver>=2,<3"
```
vasukulkarni, line 32: is there a reason for version b/w 2 and 3?

setup.py
```diff
@@ -20,7 +20,7 @@
         "robotframework-requests",
         "robotframework-seleniumlibrary",
         "robotframework-jupyterlibrary>=0.3.1",
-        "robotframework-OpenShiftCLI",
+        "robotframework-OpenShiftCLI>=1.0.1",
```
vasukulkarni, line 23: can we pin to 1.0.1 ?  use `==`

setup.py
lugi0: v3 has breaking changes but I think it's still not released, my pip install got me version 2 by default

setup.py
lugi0: sure

vasukulkarni (approved):

vasukulkarni:
@jgarciao please cherrypick and let me know if it worked

jgarciao:
I integrated the commits in #138

PR #141, opened 2021-12-02
Automation of RHOSAK UseCase

bdattoma:
This PR wants to automate this set of Polarion IDs: ODS-248  ODS-247  ODS-246  ODS-245  ODS-243  ODS-241  ODS-239.
It follows the entire use case of RHOSAK as described in the QuickStart and as reported in the Polarion TCs:
1. enable rhosak from dashboard
2. connect to HybridCloudConsole 
3. create a Kafka stream
4. configure the stream (topic, service account and permissions)
5. run Kafka Producer and Consumer python code in JypyterLab to check if the communication with the stream is working.

The clean up code (remove rhosak, streams and configuration) is included as Test Teardown

It also brings some common keywords to JupyterLabLauncher.robot to handle different open notebook tabs.

### DIFF
--- a/tests/Resources/Page/HybridCloudConsole/Rhosak.robot
+++ b/tests/Resources/Page/HybridCloudConsole/Rhosak.robot
@@ -30,8 +30,8 @@ Create Kafka Stream Instance
   Wait Until Page Does Not Contain Element    xpath=//div[@id='modalCreateKafka']  timeout=10
 
 Check Stream Status
-  [Arguments]  ${target_status}
-  ${status}=  Get Text    xpath://tr[@tabindex='0']/td[@data-label='Status']
+  [Arguments]  ${target_status}  ${target_stream}
+  ${status}=  Get Text    xpath=//tr[td[@data-label='Name' and *[text()='${target_stream}']]]/td[@data-label='Status']
   Should Be Equal    ${status}    ${target_status}
 
 Check Stream Creation
@@ -47,8 +47,10 @@ Delete Kafka Stream Instance
   Wait Until Element Is Enabled    xpath=//button[text()='Delete']
   Capture Page Screenshot  3.png
   Click Button    Delete
+  Wait Until Page Does Not Contain Element    xpath=//div[@class='pf-l-bullseye']
+  Sleep  3
   Wait Until Page Contains    Create Kafka instance
-  Wait Until Keyword Succeeds    300  1  Page Should Not Contain    xpath=//tr/td[@data-label='Name' and //*[text()='${stream_name}']]
+  Wait Until Keyword Succeeds    300  1  Page Should Not Contain Element    xpath=//tr/td[@data-label='Name' and (text()='${stream_name}' or *[text()='${stream_name}'])]
   # Wait Until Keyword Succeeds    300  1  Page Should Contain    No results found
 
 Create Topic
@@ -169,7 +171,7 @@ Delete Stream Topic
   Wait Until Page Does Not Contains HCC Generic Modal
   Capture Page Screenshot  deleting_after_modal.png
   Wait Until Page Contains    Create topic
-  Wait Until Keyword Succeeds    300  1  Page Should Not Contain    xpath=//tr/td[@data-label='Name' and *[text()='${topic_to_delete}']]
+  Wait Until Keyword Succeeds    300  1  Page Should Not Contain Element    xpath=//tr/td[@data-label='Name' and *[text()='${topic_to_delete}']]
   Capture Page Screenshot  deletion_topic.png
 
 Delete Service Account By Client ID
@@ -179,7 +181,7 @@ Delete Service Account By Client ID
   Click Button  Delete
   Wait Until Page Does Not Contains HCC Generic Modal
   Wait Until Page Contains    Create service account
-  Wait Until Keyword Succeeds    300  1  Page Should Not Contain    xpath=//tr[td[@data-label='Client ID' and text()='${client_id_delete}']]
+  Wait Until Keyword Succeeds    300  1  Page Should Not Contain Element    xpath=//tr/td[@data-label='Client ID' and text()='${client_id_delete}']
 
 
 Clean Up RHOSAK
--- a/tests/Resources/Page/HybridCloudConsole/Rhosak.robot
+++ b/tests/Resources/Page/HybridCloudConsole/Rhosak.robot
@@ -186,6 +186,8 @@ Delete Service Account By Client ID
 
 Clean Up RHOSAK
   [Arguments]  ${stream_to_delete}  ${topic_to_delete}  ${sa_clientid_to_delete}
+  OpenShiftCLI.Delete      kind=ConfigMap  name=rhosak-validation-result  namespace=redhat-ods-applications
+  Switch Window  title:Red Hat OpenShift Streams for Apache Kafka
   Menu.Navigate To Page    Streams for Apache Kafka   Kafka Instances
   Enter Stream  stream_name=${stream_to_delete}
   Enter Stream Topics Section
@@ -197,6 +199,3 @@ Clean Up RHOSAK
   Capture Page Screenshot  after deleting_stream.png
   Click Link  Service Accounts
   Delete Service Account By Client ID  client_id_delete=${sa_clientid_to_delete}
-  OpenShiftCLI.Delete      kind=ConfigMap  name=rhosak-validation-result  namespace=redhat-ods-applications
-
-
--- a/tests/Resources/Page/HybridCloudConsole/Rhosak.robot
+++ b/tests/Resources/Page/HybridCloudConsole/Rhosak.robot
@@ -18,7 +18,7 @@ Create Kafka Stream Instance
   END
   Input Text    xpath=//input[@id='form-instance-name']    ${stream_name}
   Click Element    xpath=//div[text()='${cloud_provider}']
-  Select From List By Value    id:cloud-region-select   ${stream_region}
+  Select From List By Value    id:form-cloud-region-option   ${stream_region}
   Click Button    Create instance
   Capture Page Screenshot  form.png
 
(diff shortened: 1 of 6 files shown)


vasukulkarni:
:100:

tests/Resources/Page/HybridCloudConsole/Rhosak.robot
```diff
@@ -0,0 +1,123 @@
...
+  Click Button  Create Kafka instance
+  Sleep  5
+  Maybe Accept Cookie Policy
+  Sleep  5
+  Maybe Agree RH Terms and Conditions
+  Wait Until Page Contains Element    xpath=//div[@id='modalCreateKafka']  timeout=10
+  ${warn_msg}=  Run Keyword And Return Status    Page Should Not Contain    To deploy a new instance, delete your existing one first
+  IF    ${warn_msg} == ${False}
+     Log  level=WARN  message=The next keywords are going to fail because you cannot create more than one stream at a time.
+  END
```
lugi0, line 18: You could skip the rest of the keyword?
https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Skip

tests/Resources/Page/ODH/AiApps/Rhosak.robot
```diff
@@ -0,0 +1,49 @@
...
+
+Load Kafka Notebooks From Git
+  [Arguments]  ${git_repo_http}
+  Open With JupyterLab Menu  File  New  Notebook
+  Sleep  1
+  Maybe Select Kernel
+  Sleep  3
+  Close Other JupyterLab Tabs
+  Clone Git Repository  REPO_URL=${git_repo_http}
+  Sleep  5
```
lugi0, line 23: Why not use the keywords we already have to clone from Git?

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
```diff
@@ -1,23 +1,149 @@
...
+
+
+  
+
+
+
+
+
+
+
```
lugi0, line 149: too many trailing lines

lugi0 (commented):
Looks good, left a couple of comments

tests/Resources/Page/ODH/AiApps/Rhosak.robot
bdattoma: "Clone Git Repository" is one of the already developed. But actually, I think I could remove the lines from 17 to 20 and remove this "Load Kafka Notebooks From Git". 
Thanks

tests/Resources/Page/HybridCloudConsole/Rhosak.robot
bdattoma: I will check this. My initial idea was to "skip" the rest of the TC, but then I thought that making the TC failing naturally (and giving just a warning) would implicitly test the limit on one kafka instance per user: if you get the warning message but the rest of the TC does not fail then there can be something wrong in the functionality.

What do you think?

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
bdattoma: thanks, I will remove them

tests/Resources/Page/HybridCloudConsole/Rhosak.robot
lugi0: Is the limit of one kafka instance per user something that we really want to test? That seems like a feature of another product, so I wouldn't want to spend time debugging issues that are not in rhods. 
I also assume this limit of one instance per user is temporary / can be increased. This might need to be revisited down the line anyways.

bdattoma: makes sense yeah. But my idea is that even if this is a functionality from another product it is still faced by RHODS users. So, given that this would be implicitly tested by our TC just without skipping a line which would fail anyway, it is costless for us to test it.

tarukumar: I think we can throw error instead of failure or skip it will not affect our overall test suite status also when people will go thought the logs they can see the error.We should just skip the whole block of code below to stop the failure?

bdattoma: @tarukumar when you say "throw error", you mean setting the log level to ERROR or doing something like "raise ValueError" in python?

tarukumar: i was suggesting to set the log level to error.

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
```diff
@@ -1,23 +1,149 @@
...
 Library         SeleniumLibrary
+Library         OpenShiftCLI
+Resource        ../../../Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+Resource        ../../../Resources/Page/HybridCloudConsole/HCCLogin.robot
+Resource        ../../../Resources/Page/HybridCloudConsole/Rhosak.robot
+Resource        ../../../Resources/Page/ODH/JupyterHub/LoginJupyterHub.robot
+Resource        ../../../Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+Resource        ../../../Resources/Page/ODH/JupyterHub/JupyterLabSidebar.robot
+Resource        ../../../Resources/Page/ODH/AiApps/Rhosak.robot
 Suite Setup     Kafka Suite Setup
```
tarukumar, line 11: Just a suggestion ..I think it would be better not to call individual resource file(.robot)  instead we can call the one common resource file which has all these file already present(.resource). For example : We can remove all the Jupyterhub robot file from above and just call ODHJupyterhub.resource which has all those file imported there. What do you think

lugi0: I think this is a great suggestion and something that I never got around to doing, +1 from me

bdattoma: agree with this, thanks!

tests/Resources/Page/HybridCloudConsole/Rhosak.robot
bdattoma: I set the log level to ERROR

bdattoma:
tested using Jenkins: [JOB#300](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/300/TestReport/) .
Before merging, still need to apply changes based on some comments

bdattoma:
last two testing jobs: 

- only ODS-248 run (_success_) [JOB#308](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/308/) 

- entire PR testing suite run (_in progress_): [JOB#309](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/309/)

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
```diff
@@ -49,12 +57,79 @@ Verify User Is Able to Create And Delete a Kafka Stream
...
+  Wait for RHODS Dashboard to Load
+  Launch JupyterHub Spawner From Dashboard
+  Wait Until Page Contains Element  xpath://input[@name="Standard Data Science"]
+  &{notebook_envs}=  Create Dictionary  KAFKA_BOOTSTRAP_SERVER=${bootstrap_server}  KAFKA_USERNAME=${kafka_client_id}
+  ...                                   KAFKA_PASSWORD=${kafka_client_secret}  KAFKA_TOPIC=${topic_name_test}
+  ...                                   KAFKA_CONSUMER_GROUP=${consumer_group_test}
+  Spawn Notebook With Arguments  image=s2i-generic-data-science-notebook  envs=&{notebook_envs}
+  Wait for JupyterLab Splash Screen  timeout=60
+  Maybe Select Kernel
+  ## clone JL notebooks from git and run
```
lugi0, line 112: This should be handled by `Spawn Notebook With Arguments` after #172 is merged

lugi0 (commented):
Looks good, I would just be careful about the overlap with #172

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
bdattoma: > Looks good, I would just be careful about the overlap with #172

I can send a new PR to fix the overlap after both are merged. Do you think the overlap may create troubles during execution?

pablofelix (approved):
lgtm

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
lugi0: Yes, if both the keyword and the TC are waiting for `Wait for JupyterLab Splash Screen` the TC will fail because the keyword will give back control after the splash has already appeared

bdattoma: you're right. I would say, if this PR is merged before the #172, i will raise a PR for correction afterwords, otherwiise I will just commit it removing those 2 lines. What do you say?

lugi0: If it's ready to go right now just merge it and we can update it after #172

bdattoma: yes it is ready. Just missing an approval

lugi0: PR job for #172 is running, so up to you. I can give the approval if you want to merge now.

bdattoma: ok let's merge, I'll send a PR after that once 172 will be finished

lugi0 (approved):

PR #145, opened 2021-12-03
enhancement of some keywords to fix anaconda TC

bdattoma:
This PR implements two improvements to the "Select Project By Name" and "Get Pod Logs Manual Version" keyword:
1. compatibility with OCP 4.8 for both the keywords
2. enable the "show default projects" in OCP 4.9 if not already selected (select project by name)
3. change name (old name: Get Pod Logs) to avoid conflict with OpenshiftCLI lib
4. New keyword to handle "Server not running" page (which is intermittent) after "Fix Spawner Status" in Ananconda TC 
5. Add a missing resource import in JupyterSpawnerHub.robot

Moreover, it makes the TC using it compatible with OCP 4.8 (e.g., Anaconda TCs)

### DIFF
--- a/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
+++ b/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
@@ -1,17 +1,18 @@
 *** Settings ***
 Resource   ../../OCPDashboard/Page.robot
 Resource   ../../ODH/ODHDashboard/ODHDashboard.robot
-Library         ../../../../libs/Helpers.py
+Library    ../../../../libs/Helpers.py
 
 *** Keywords ***
 Get Pod Logs Manual Version
   [Arguments]  ${namespace}  ${pod_search_term}
   Navigate To Page    Workloads    Pods
   Search Last Item Instance By Title in OpenShift Table  search_term=${pod_search_term}  namespace=${namespace}
-  Click Link    xpath://tr[@data-key='0-0']/td[@id='name']/span/a
+  Click Link    xpath://tr[@data-key='0-0']/td/span/a
   Click Link    Logs
   Sleep  2
   Capture Page Screenshot  logs_page.png
   ${logs_text}=  Get Text    xpath://div[@class='log-window__lines']
   ${log_rows}=  Text To List  ${logs_text}
+
   [Return]  ${log_rows}
--- a/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
+++ b/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
@@ -4,11 +4,11 @@ Resource   ../../ODH/ODHDashboard/ODHDashboard.robot
 Library         ../../../../libs/Helpers.py
 
 *** Keywords ***
-Get Pod Logs
+Get Pod Logs Manual Version
   [Arguments]  ${namespace}  ${pod_search_term}
   Navigate To Page    Workloads    Pods
   Search Last Item Instance By Title in OpenShift Table  search_term=${pod_search_term}  namespace=${namespace}
-  Click Link    xpath://tr[@data-id='0-0']/td[@id='name']/*/a
+  Click Link    xpath://tr[@data-key='0-0']/td[@id='name']/span/a
   Click Link    Logs
   Sleep  2
   Capture Page Screenshot  logs_page.png
--- a/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
+++ b/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
@@ -4,7 +4,7 @@ Resource   ../../ODH/ODHDashboard/ODHDashboard.robot
 Library    ../../../../libs/Helpers.py
 
 *** Keywords ***
-Get Pod Logs Manual Version
+Get Pod Logs From UI
   [Arguments]  ${namespace}  ${pod_search_term}
   Navigate To Page    Workloads    Pods
   Search Last Item Instance By Title in OpenShift Table  search_term=${pod_search_term}  namespace=${namespace}
@@ -14,5 +14,4 @@ Get Pod Logs Manual Version
   Capture Page Screenshot  logs_page.png
   ${logs_text}=  Get Text    xpath://div[@class='log-window__lines']
   ${log_rows}=  Text To List  ${logs_text}
-
   [Return]  ${log_rows}
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -225,7 +225,7 @@ Login Via Button
    Wait Until Page Contains  Log in with
 
 Maybe Handle Server Not Running Page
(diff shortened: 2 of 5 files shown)


bdattoma:
Jenkins job for smoke on OCP 4.8: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-smoke/260/ (it uses my fork of the repo to run the smoke using the branch with the fix). The changes listed in the description made the Anaconda TC pass the Smoke. There are some failures in the run but they are not caused by these changes. 

At the moment, the same Smoke job is running on OCP 4.9: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-smoke/261/parameters/

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
```diff
@@ -1,17 +1,18 @@
 *** Settings ***
 Resource   ../../OCPDashboard/Page.robot
 Resource   ../../ODH/ODHDashboard/ODHDashboard.robot
-Library         ../../../../libs/Helpers.py
+Library    ../../../../libs/Helpers.py
 
 *** Keywords ***
-Get Pod Logs
+Get Pod Logs Manual Version
```
lugi0, line 7: What do you mean by manual version?

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -222,3 +223,9 @@ Login Via Button
    ...  Keyword.
    Click Element  xpath://a[@id='login']
    Wait Until Page Contains  Log in with
+
+Maybe Handle Server Not Running Page
+  ${SNR_visible} =  Run Keyword and Return Status   Wait Until Page Contains    Server not running  timeout=15
+  IF  ${SNR_visible}==True
+         Handle Server Not Running
+  END
```
lugi0, line 231: We already have the Keywords `Server Not Running Is Visible` and `Handle Server Not Running`, why define a new keyword?
If you really need a new keyword, you should use `Server Not Running Is Visible` in it to check whether you are in that page or not IMHO; that way if the text changes or we want to modify the timeout we only have to modify one keyword instead of two.

bdattoma: I create another one because the original one use "Page Should Contains" but in this case the "Wait Until Page Contains" is needed, because it takes some time before appearing. 
As alternative, I could use the original one and adding a Sleep of 15sec before running the keyword. When the time is kind of variable, I prefer using the Wait until. Maybe it is something that could be changed in the original one.

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
bdattoma: I wanted to change it to avoid conflicts with the Openshift CLI lib. I called it "manual" because in this version the robot is going to "read" the logs by clicking in the openshift UI

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
lugi0: Then my suggestion would be to update `Server Not Running Is Visible` to use `Wait Until Page Contains`, with the timeout as an input variable. Then you can either reuse that keyword from within this one, or directly use a combination of `Server Not Running Is Visible` + `Handle Server Not Running` to achieve the same result.

lugi0: (Just keep in mind that if we are unsure about which page is being displayed, in the worst case scenario `Wait Until Page Contains` will "waste" `timeout` seconds waiting for something to appear.)

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
lugi0: Then perhaps `Get Pod Logs From UI` would be a better name?

bdattoma: that's the one! I buy it :) i will change it, thanks

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
bdattoma: I thought about changing the original one, but given that this PR would have been a "quick" one to fix the Anaconda TC in the smoke, I didn't want to change the already present. Same reason why I didn't include the "Server not running" keyword directly in the Fix Spawner Status.

But hey, I will do both the changes now to have a better code. Thanks

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
bdattoma: done in this commit 889625840993aa438c1a7d82f0ffa8fe10dc1298

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
bdattoma: done in commit https://github.com/red-hat-data-services/ods-ci/pull/145/commits/bb84fdd4d769dd8d59087076e85e69c6f329d9c2

Implemented in a way to avoid the waste of timeout at the beginning of "Fix Spawner Status" keyword.

lugi0 (dismissed):
lgtm

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
bdattoma: > Implemented in a way to avoid the waste of timeout at the beginning of "Fix Spawner Status" keyword.
I have to change this again because it appears to have a situation where it fails. This situation didn't show up in the Friday testing, but it did today. I'm working on it

bdattoma:
> I have to change this again because it appears to have a situation where it fails. This situation didn't show up in the Friday testing, but it did today. I'm working on it

Tested with the fix implemented in commit ecba8b4049bdaaca5ebb3d4fd3b71f2ea49462b3:
1. pr test job to test only ODS-272 (OCP 4.8): JOB#[267](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/267) 
2. pr test job to test only ODS-272 (OCP 4.9): JOB#[269](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/269)
3. Entire smoke test run on OCP 4.8: JOB#[266](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-smoke/266)
4. Entire smoke test run on OCP 4.9: JOB#[267](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-smoke/267)

In all of the four runs, everything related to this PR is successful (you can ignore the two anaconda failures in run 4 because the anaconda was enabled before running the smoke but I fixed it before the ods-272 was started)

lugi0 (approved):
LGTM

aloganat (approved):
Looks good to me

PR #149, opened 2021-12-13
Open RHODS apps using hyperlink instead of dropdown

samuelvl:
The PR #168 in odh-dashboard https://github.com/red-hat-data-services/odh-dashboard/pull/168 removed the option to open JupyterHub (and other RHODS apps) from the dropdown menu.

From 1.4.0-2 version, use the "Launch ${dashboard_app} From RHODS Dashboard Link" instead of "Launch ${dashboard_app} From RHODS Dashboard Dropdown" option.

### DIFF
--- a/tests/Resources/Common.robot
+++ b/tests/Resources/Common.robot
@@ -15,7 +15,12 @@ Begin Web Test
     Open Browser  ${ODH_DASHBOARD_URL}  browser=${BROWSER.NAME}  options=${BROWSER.OPTIONS}
     Login To RHODS Dashboard  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
     Wait for RHODS Dashboard to Load
-    Launch JupyterHub From RHODS Dashboard Dropdown
+    ${version-check} =  Is RHODS Version Greater Or Equal Than  1.4.0
+    IF  ${version-check}==True
+      Launch JupyterHub From RHODS Dashboard Link
+    ELSE
+      Launch JupyterHub From RHODS Dashboard Dropdown
+    END
     Login To Jupyterhub  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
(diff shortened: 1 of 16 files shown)


tests/Resources/Common.robot
```diff
@@ -15,7 +15,7 @@ Begin Web Test
     Open Browser  ${ODH_DASHBOARD_URL}  browser=${BROWSER.NAME}  options=${BROWSER.OPTIONS}
     Login To RHODS Dashboard  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
     Wait for RHODS Dashboard to Load
-    Launch JupyterHub From RHODS Dashboard Dropdown
+    Launch JupyterHub From RHODS Dashboard Link
```
jgarciao, line 18: In order to maintain compatibility of tests with the current version in production this changes should be done like this:

```
${version-check} =  Is RHODS Version Greater Or Equal Than  1.4.0
IF  ${version-check}==True
  Launch JupyterHub From RHODS Dashboard Link
ELSE
  Launch JupyterHub From RHODS Dashboard Dropdown
END
```

vasukulkarni: +1, need to maintain compatibility

samuelvl: Done!

lugi0 (approved):
lgtm

lugi0 (changes requested):
I think `Is RHODS Version Greater Or Equal Than  1.4.0` actually needs to be changed to `Is RHODS Version Greater Or Equal Than  1.4.0-2` due to how semver handles comparisons when -x is present.
This would currently result in 1.4.0-2 being < 1.4.0

lugi0 (approved):
nevermind, we have already merged an update to the comparison keyword that drops the `-x` from the comparison
https://github.com/red-hat-data-services/ods-ci/blob/master/tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot#L76

vasukulkarni (approved):

PR #154, opened 2021-12-16
fix select region html id for RHOSAK ODS-242

bdattoma:
in the RHOSAK UI one of the HTML element id is changed and our ODS-242 is failing because of that. This PR wants to fix the TC

Signed-off-by: Berto D'Attoma bdattoma@redhat.com

### DIFF
--- a/tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
+++ b/tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
@@ -101,7 +101,7 @@ Create Kafka Stream Instance
   END
   Input Text    xpath=//input[@id='form-instance-name']    ${stream_name}
   Click Element    xpath=//div[text()='${cloud_provider}']
-  Select From List By Value    id:cloud-region-select   ${stream_region}
+  Select From List By Value    id:form-cloud-region-option   ${stream_region}
   Click Button    Create instance
   Capture Page Screenshot  form.png
   Sleep    2


lugi0 (approved):
lgtm

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
```diff
@@ -101,7 +101,7 @@ Create Kafka Stream Instance
   END
   Input Text    xpath=//input[@id='form-instance-name']    ${stream_name}
   Click Element    xpath=//div[text()='${cloud_provider}']
-  Select From List By Value    id:cloud-region-select   ${stream_region}
+  Select From List By Value    id:form-cloud-region-option   ${stream_region}
```
tarukumar, line 104: Code looks fine . @bdattoma @lugi0 just a thought could we add multiple locator with "or" using conatin or something which will be dynamic so that in case wither id gets changed (edge case) or  anything gets changed in future it will not impact our test suites?

tarukumar (approved):

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
lugi0: Mh, that would work only in the case where the id (or other locator) is changed in such a way that part of the original is still included. I'm not sure it would solve most of these kinds of issues we have encountered?

bdattoma: In practice, it can be done. The only "issue" I can see doing this is the code will contains outdated pieces and we wouldn't know. It is just a thought

In general, it could be helpful when we must write some code to be compatible with two different platform version and no common locators can be found.

bdattoma:
PR test on Jenkins: [PR#283](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/283/).
The RHOSAK TCs are successful. The failures in the job are not related to this PR.
The changes are independent from OCP version because they are about RHOSAK console.

tests/Tests/600__ai_apps/650__rhosak/650__rhosak.robot
tarukumar: > Mh, that would work only in the case where the id (or other locator) is changed in such a way that part of the original is still included. I'm not sure it would solve most of these kinds of issues we have encountered?

hmm. May be union(|) will be helpful in this case

tarukumar: > In practice, it can be done. The only "issue" I can see doing this is the code will contains outdated pieces and we wouldn't know. It is just a thought
> 
> In general, it could be helpful when we must write some code to be compatible with two different platform version and no common locators can be found.

Totally Agree.There will be chance of having outdated xpath. yes, it would be helpful for platform compatibility..In some case it will help us not rewrite the whole block of code using If Else block

bdattoma: agree with that. Let's have a dedicated discussion on that when we have a case we want to solve

PR #160, opened 2021-12-29
Add support to create vm instances in openstack

aloganat:

### DIFF
--- a/utils/scripts/terraform/openstack/README.md
+++ b/utils/scripts/terraform/openstack/README.md
@@ -1,0 +1,131 @@
+# Provision vm in openstack
+Script to manage instances in openstack using terraform.
+Currently the script handles provisioning and deleting vm instances.
+
+## Prerequisites
+Python 3.x
+Install terraform in cli
+pip3 install python-terraform
+
+## How to Run:
+1. Change directory to the terraform/openstack directory.
+
+```
+# cd ods-ci/utils/scripts/terraform/openstack
+```
+
+2. Now run the script.
+
+## Usage
+
+```
+# python3 provision.py -h
+usage: provision.py [-h] {create_instance,delete_instance,set_config} ...
+
+Script to manage instances in Openstack
+
+optional arguments:
+  -h, --help            show this help message and exit
+
+Available sub commands:
+  {create_instance,delete_instance,set_config}
+                        sub-command help
+    create_instance     Creates an instance in OpenStack
+    delete_instance     Deletes an instance in OpenStack
+    set_config          Sets config for using Openstack with terraform
+
+# python3 provision.py create_instance -h
+usage: provision.py create_instance [-h] --cloud_name CLOUD_NAME --vm_name
+                                    VM_NAME --vm_user VM_USER --vm_private_key
+                                    VM_PRIVATE_KEY --key_pair KEY_PAIR
+                                    --network_name NETWORK_NAME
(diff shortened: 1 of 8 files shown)


utils/scripts/terraform/openstack/requirements.sh
```diff
@@ -0,0 +1,7 @@
+sudo dnf install -y https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
+sudo dnf install -y python3-devel git redhat-lsb jq podman yum-utils java-1.8.0-openjdk
+sudo yum-config-manager --add-repo https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo
```
jgarciao, line 3: I think yum-config-manager is being replaced by `dnf config-manager`:
`sudo dnf config-manager --add-repo https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo`

utils/scripts/terraform/openstack/README.md
```diff
@@ -0,0 +1,21 @@
+# Provision vm in openstack
+Script to provision vm in openstack using terraform
```
jgarciao, line 2: You can add an usage example, like adding a node for X

utils/scripts/terraform/openstack/README.md
```diff
@@ -0,0 +1,21 @@
...
+```
+# cd ods-ci/utils/scripts/terraform/openstack
+```
+
+2. Now run the script.
+
+## Usage
+
+```
+# python3 provision.py -h
```
jgarciao, line 21: Can you add an example with all the required parameters?

utils/scripts/terraform/openstack/requirements.sh
aloganat: incorporated the comments.

utils/scripts/terraform/openstack/README.md
aloganat: Done.

utils/scripts/terraform/openstack/README.md
aloganat: Done.

utils/scripts/terraform/openstack/requirements.sh
```diff
@@ -0,0 +1,7 @@
+sudo dnf install -y https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
```
lugi0, line 1: Is this going to run ods-ci tests? Do we also need to pull down the chromedriver here?

lugi0 (dismissed):
Looks good to me, but I'm not an expert on terraform :)

utils/scripts/terraform/openstack/requirements.sh
jgarciao: Yes, the main [Readme.md](https://github.com/red-hat-data-services/ods-ci/blob/master/README.md) file says that chrome and chromedriver should have the same version number. In the [Dockerimage](https://github.com/red-hat-data-services/ods-ci/blob/master/build/Dockerfile) there is an example about how to install required packages

aloganat: Have installed chromedriver as mentioned in Dockerfile. We need this step in case if we want to use the vm as static worker node.

jgarciao (approved):
LGTM

vasukulkarni (approved):

PR #162, opened 2022-01-05
Added Tc for aikit to verify in RHODS dashboard,and install functiona…

tarukumar:
This Script verify the aikit tile is present in RHODS Dashboard(Explore tab). It also verify the installation of aikit non-market place operator and it functionality


tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
```diff
@@ -40,6 +40,7 @@ Search Installed Operator
 
 Is Operator Installed
   [Arguments]  ${operator}
+  Run Keyword and Return Status     Wait Until Element is Visible      //a[@data-test-operator-row="${operator}"]    timeout=10
```
bdattoma, line 43: isn't it redundant with the next line where you get the ${is_installed} variable?

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
```diff
@@ -104,3 +105,32 @@ Is RHODS Version Greater Or Equal Than
     ${comparison} =  GTE  ${ver}  ${target}
     [Return]  ${comparison}
 
+Move To Installed Operator instance
+    [Arguments]   ${operator_name}     ${tab_name}      ${namespace}
+    Switch To Administrator Perspective
+    Navigate to Installed Operators
+    Installed Operators Should Be Open
+    Log To Console   ${namespace}
+    Run Keyword If  "${namespace}" == "None"   Select All Projects
```
bdattoma, line 114: I think it is possible to use "Select Project By Name" (with the appropriate parameter set) instead of "Select All Projects".

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
```diff
@@ -104,3 +105,32 @@ Is RHODS Version Greater Or Equal Than
...
+    ...         ELSE   Select Project By Name   ${namespace}
+    #Select All Projects
+    Click On Searched Operator   ${operator_name}
+    Switch To New Tab       ${tab_name}
+    sleep    5
+
+Create Installed Operator instance
+    [Arguments]    ${operator_name}     ${tab_name}     ${namespace}=None
+    Move To Installed Operator instance    ${operator_name}     ${tab_name}    ${namespace}
```
bdattoma, line 123: this naming might appear as ambiguous, since the instance has not been created yet. it may be difficult for someone who's not familiar with this operation. e.g., "Go To Operator Config Page"

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
```diff
@@ -3,26 +3,31 @@ Library     SeleniumLibrary
 
 *** Keywords ***
 Install Operator
-    [Arguments]    ${operator}
+    [Arguments]    ${operator}      ${redhat_marketplace}=None
```
bdattoma, line 6: why not True/False?

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
```diff
@@ -3,26 +3,31 @@ Library     SeleniumLibrary
...
     Wait Until Installation Completes
 
-
 Search Operator
    [Arguments]    ${operator}
    Wait Until Element is Visible    //input[@data-test="search-operatorhub"]   timeout=150
    Input text    //input[@data-test="search-operatorhub"]   ${operator}
    Press keys    //input[@data-test="search-operatorhub"]   RETURN
 
+Select Non Marketplace Operator
```
bdattoma, line 22: Probably, we would need to have a more generic keyword to install operator from whatever catalog (marketplace, certified operators, community, ibm etc)

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
```diff
@@ -104,3 +105,32 @@ Is RHODS Version Greater Or Equal Than
...
+    Click Button     Create ${tab_name}
+    Wait Until Element is Visible     //button[contains(text(), "Create")]          timeout=10
+    Click Button      Create
```
bdattoma, line 126: Probably we may need a keyword to check the instance status like done in "Wait Until Installation Completes"

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
```diff
@@ -37,9 +42,16 @@ Confirm Show Operator
...
 Wait Until Installation Completes
-    Wait Until Page Contains    ready for use   timeout=150
+    Wait Until Page Contains    ready for use   timeout=300
 
 Operator Should Be Installed
     [Arguments]    ${operator}
     Page Should Contain    ${operator}
     Page Should Contain    ready for use
+
+Get List Of Operator Available
```
bdattoma, line 52: "Get List [..]" but it returns a numeric value

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
```diff
@@ -0,0 +1,68 @@
...
+    Login To RHODS Dashboard  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
+    Wait for RHODS Dashboard to Load
+    ${status}       Run keyword and Return Status           Verify Service Is Enabled          ${intel_aikit_container_name}
+    Run keyword If  ${status}    Launch JupyterHub Spawner From Dashboard
+    ...        ELSE   FAIL      ${intel_aikit_container_name}  tile  is not present in the Enable tab
+    Wait Until Page Contains Element  xpath://input[@name="oneAPI AI Analytics Toolkit"]
+    Wait Until Element Is Enabled    xpath://input[@name="oneAPI AI Analytics Toolkit"]   timeout=10
+    Spawn Notebook With Arguments  image=oneapi-aikit
+    Wait for JupyterLab Splash Screen  timeout=60
+    Maybe Select Kernel
```
bdattoma, line 51: What about checking that the directory with code examples has been created and/or we could check that the spawned image is actually the target one. it is possible to achieve by getting an env variable.
This code to show all the env variables
`import os
os.environ()`

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: it just a time taken  for operator to be appear. When we search the operator it take sometime to appear that's the reason i added for wait where as is installed fetch web element and return status. I tried to avoid usage of sleep

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
tarukumar: True and False have special meaning in robot Framework. I try to avoid that. If you want me to change i can do. IMHO i will stick with None.

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
tarukumar: good catch ..my intention is to get the number ...will modify it

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: By default project is all project ..that's why i kept it if user want to select project we have that in else statement below

bdattoma: I mean using `Select Project By Name  All Projects` instead of `Select All Projects`

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
bdattoma: mmmh I think if the parameter is a boolean, it would be better to limit its value range to 2 values. It's not a big deal to leave None

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
bdattoma: may help to returns the list of web elements?

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
bdattoma: okay so maybe you don't need the next lines. If this line succeeds, you have the "install" status of the operator. Correct?

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
tarukumar: do you mean to create another keyword ?

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
tarukumar: my intention to have this keyword to provide the no of operator present with same name so that we can verify it if somewhere in future if removes we will get to know

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: what i have seen is it doesn't take much time to get to status "Conditions: Initialized, Deployed". If it fails here we will get to know in the next step while launching the notebook. If you feel i can write a code for it

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
bdattoma: I mean having a keyword like "Select Operator" which takes as argument the operator name + the catalog name (marketplace or community for examples)

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
bdattoma: mmmh not sure honestly, let's see how it behaves with different operators in future

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
bdattoma: yeah I got the intention. It's fine

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: hmm..i have seen same atleast in openvino and aikit

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
tarukumar: hmm...i will modify the below keyword to handle the catlog value.I think this keyword should be there to handle non marketplace value

bdattoma: isn't possible to handle non-marketplace using the same keyword? At the end, when you say non-marketplace you need an operator from a catalog different from mrktpl.

tarukumar: you could do that ...however it will need nested If/else  block. It will be easy to understand and implement  if we have different keyword for both. Below is the code i am planning to add:

''''
Select Operator Without Respective Catalog Name Included
    [Arguments]    ${operator}      ${catalog}
    Wait Until Element is Visible    //a[contains(@data-test, "${operator}") and not (contains(@data-test," ${catalog}"))]   timeout=50
    Click Element    //a[contains(@data-test, "${operator}") and not (contains(@data-test," ${catalog}"))]

Select Operator with Catalog Name Included
    [Arguments]    ${operator}     ${catalog}
    Wait Until Element is Visible   //a[contains(@data-test, "${operator}") and (contains(@data-test," ${catalog}"))]  timeout=50
    Click Element    //a[contains(@data-test, "${operator}") and (contains(@data-test,"${catalog}"))]
'''

bdattoma: maybe, `Select Operator with Catalog Name Included` is enough to do the job because you always know the catalog the operator must be fetched from

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
tarukumar: added code for it

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: Done

tests/Resources/Page/OCPDashboard/OperatorHub/OperatorHub.robot
tarukumar: Done

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
tarukumar: Done

tarukumar:
To Check the spawn image i am currently check the whole path. There are other way f doing it just by name whihc we may include in future like below example:

    #Add and Run JupyterLab Code Cell in Active Notebook   print(os.environ['JUPYTER_IMAGE'])
    #${output} =  Get Text  (//div[contains(@class,"jp-OutputArea-output")])[last()]
    #${image_path_list}=  Split string     ${output}     /
    #Should Match  ${image_path_list}[-1]        oneapi-aikit:oneapi-aikit

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: yeah we actually do not need that..i have kept it as extra precautions nothing more

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: have changed it to ..hope it will clear the confusion

tarukumar:
Verified using Jenkins job:[311](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/311). Code works fine without any issue. There are some failure which are not related to this script.

bdattoma:
> #Should Match  ${image_path_list}[-1]        oneapi-aikit:oneapi-aikit

okay @tarukumar . FYI, for the future, we have a keyword to take the cell output.

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
bdattoma: I didn't mean the TC title but the name of "Move to installed operator instance" keyword

tarukumar:
> > #Should Match  ${image_path_list}[-1]        oneapi-aikit:oneapi-aikit
> 
> okay @tarukumar . FYI, for the future, we have a keyword to take the cell output.

tarukumar:
Mistakenly got closed. Reopened it.

bdattoma (approved):

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
```diff
@@ -0,0 +1,81 @@
...
+    Sleep  3
+    Close Other JupyterLab Tabs
+    ${is_launcher_selected} =  Run Keyword And Return Status  JupyterLab Launcher Tab Is Selected
+    Run Keyword If  not ${is_launcher_selected}  Open JupyterLab Launcher
+    Launch a new JupyterLab Document
+    Sleep  3
+    Maybe Select Kernel
+    Close Other JupyterLab Tabs
```
lugi0, line 61: If this is merged after my PR, this should be removed, these steps are now rolled into `Spawn Notebook With Arguments`

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
```diff
@@ -0,0 +1,81 @@
...
+    Add and Run JupyterLab Code Cell in Active Notebook  import os
+    Wait Until JupyterLab Code Cell Is Not Active
+    Run Cell And Check Output   print(os.environ['JUPYTER_IMAGE'])     ${image_path}/oneapi-aikit:oneapi-aikit
```
lugi0, line 64: This is also rolled into `Spawn Notebook With Arguments`

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
```diff
@@ -0,0 +1,81 @@
...
+    Maybe Skip Tour
+    Delete tabname Instance For Installed Operator     ${intel_aikit_operator_name}      AIKitContainer
+    Uninstall Operator      ${intel_aikit_operator_name} 
+
+***Keywords ***
+Intel_Aikit Suite Setup
+  Set Library Search Order  SeleniumLibrary
+
+Intel_Aikit Suite Teardown
+  Close All Browsers
```
lugi0, line 76: Should AiKit be uninstalled in teardown?

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
```diff
@@ -0,0 +1,81 @@
...
+
+Install intelaikit From the Operatorhub
+   [Tags]  ODS-760   Sanity
+    Open Installed Operators Page
+    Search Installed Operator  ${intel_aikit_operator_name} 
+    ${status}    Is Operator Installed     ${intel_aikit_operator_name}
+    IF  not ${status}
+         Open OperatorHub
+         ${no_of_operator}    Get The Number of Operator Available    ${intel_aikit_appname}
+         IF  ${no_of_operator} == ${2}
```
lugi0, line 33: I'm not sure I understand this? Why are you counting how many operators are available? What happens if this number changes in the future?

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
```diff
@@ -104,3 +105,31 @@ Is RHODS Version Greater Or Equal Than
...
+    Navigate to Installed Operators
+    Installed Operators Should Be Open
+    Log To Console   ${namespace}
+    Run Keyword If  "${namespace}" == "None"   Select Project By Name  All Projects
+    ...         ELSE   Select Project By Name   ${namespace}
+    Click On Searched Operator   ${operator_name}
+    Switch To New Tab       ${tab_name}
+    sleep    5
+
+Create tabname Instance For Installed Operator
```
lugi0, line 120: Is `tabname` supposed to be an expanded variable?

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
```diff
@@ -79,7 +80,7 @@ Switch To New Tab
 Click On Searched Operator
     [Arguments]   ${operator}
      Search Installed Operator          ${operator}
-     Sleep   1
+     Wait Until Element is Visible     xpath=//a[@data-test-operator-row="${operator}"]    timeout=10
      Click Element       xpath=//a[@data-test-operator-row="${operator}"]
      Wait until page contains           Description           timeout=10
```
lugi0, line 85: Shouldn't this be split in two keywords? `Search for Operator` and `Click on Operator`?

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
tarukumar: As of now we have two operator for aikit (marketplace and no marketplace each). Here i just wanted to make sure when we search there should be two operator. In case in future  the number gets changes we should be notified and make change. However IMHO it shouldn't change

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
tarukumar: I think no..when we have marketplace aikit Tc added. we want to uninstall in the Tc instead of tear down as it will move from one Tc to other

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: Nope

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
tarukumar: Guess then we should wait for your PR to merged and make changes to the script

tests/Tests/600__ai_apps/620__intel_aikit/620__intel_aikit.robot
tarukumar: same as above

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
lugi0: Then I would rename this keyword, because I don't understand what it's supposed to do

tarukumar: hmm...this keyword is basically travel to the tabs present when we click on any operator and try to create and instance based on the tab name.

bdattoma: I think one possible solution could be to something like:
1. rename the keyword `Move To Installed Operator instance`  to `Move To Installed Operator Page Tab`
2. rename the `Create tabname Instance For Installed Operator` to `Create Instance in Installed Operator Tab`

What do you guys think?

lugi0: Is the `tabname` the actual name of the browser tab that opens after clicking on an Operator? 
If so, I would avoid any reference to that. Maybe renaming to `Create Installed Operator Instance` would work? If it's supposed to be a general keyword to be used with any Operator, we don't really care about the `tabname` or the `operator` name, the keyword only knows that it's supposed to click on `Create`, and if it needs to be switched to a new tab before doing that it can be done either in the keyword or in an external keyword (like it is done now). 
`Switch To New Tab       ${tab_name}` isn't really necessary, as we know that a new tab will be opened and we can just switch to the newest one

bdattoma: I think the tabname here isn't referred to browser tab but it refers to the Tab in the operator page. Like
image

tarukumar: @lugi0 Comment mentioned by berto is right. One more thing i would like clarify is that "the keyword only knows that it's supposed to click on Create", it is just not Create in Operator we will have create+operator_tab_name

lugi0: Then I think I still do not understand what the keyword is doing. What is it creating exactly? And why does it need the name of one of those submenus?

jgarciao:
In my opinion the main test should be refactored into different keywords, so it's easier to understand what is doing with a quick view. Something like this:  
```
Verify Intel AiKit Operator Can Be Installed Using OpenShift Console
  [Tags]  ODS-760   Sanity
 
  Open Browser  ${OCP_CONSOLE_URL}  browser=${BROWSER.NAME}  options=${BROWSER.OPTIONS}
  LoginPage.Login To Openshift  ${OCP_ADMIN_USER.USERNAME}  ${OCP_ADMIN_USER.PASSWORD}  ${OCP_ADMIN_USER.AUTH_TYPE}
  Skip If Operator Is Already Installed   name=${intel_aikit_operator_name}  source=Certified
  Install Operator  name=${intel_aikit_operator_name} source=Certified
  	
  Verify Intel AiKit Operator Is Correctly Installed
  Verify JupyterHub Can Spawn Intel AiKit Notebook
  
  [Test Teardown]  Uninstall Intel AiKit Operator Operator
```

@tarukumar Let's have tomorrow a quick call to talk more about this

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: i am thinking to change the keyword names "Move To Installed Operator instance , Create tabname Instance For Installed Operator and Delete tabname Instance For Installed Operator" to "Move To Installed Operator Page Tab in Openshift,Create Notebook Instance For Installed Operator in Openshift and Delete Notebook Instance For Installed Operator in Openshift". Do you guys @lugi0 @bdattoma  think it will be still confusing or purposed name seems fine?

lugi0: @tarukumar I think that sounds good, I could not think of a better name for the keywords

bdattoma: the problem I see is that `Create Notebook` is something very specific which might be valid only for Openvino

aloganat:

tarukumar:
Deleted and created new PR-181

PR #167, opened 2022-01-12
to handle the failure occurs for smoke

tarukumar:

While running the [smoke]( https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-smoke/313/console) Arthy observed that  "Verify User Is Able to Activate Anaconda Commercial Edition" TC is getting failed. After going the Robot logs it looks like logs took more time than 2 second to get stream/appear in the log panel. Increase time to 4 sec to handle the case for now.

PS: I think this is  one off case and we can handle this in better way bu using other keyword like "Wait Until Keyword Succeeds" or something. We should avoid wherever possible to not hard code the wait/sleep time. We can do it in refactoring


aloganat:
lgtm

aloganat (approved):

bdattoma (approved):

bdattoma:
lgtm

PR #170, opened 2022-01-13
Add precise wait for spawner being ready

lugi0:

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -17,16 +17,15 @@ JupyterHub Spawner Is Visible
    ${spawner_visible} =  Run Keyword and Return Status  Page Should Contain  xpath:${JUPYTERHUB_SPAWNER_HEADER_XPATH}
    [return]  ${spawner_visible}
 
-JupyterHub Spawner Is Ready
+Wait Until JupyterHub Spawner Is Ready
    # Container size selector is usually the last element to load in the spawner page
    # Let's use it to check when the spawner has finished loading.
    ${version-check} =  Is RHODS Version Greater Or Equal Than  1.5.0
    IF  ${version-check}==True
-      ${spawner_ready} =    Run Keyword And Return Status   Wait Until Page Contains Element    xpath://div[contains(concat(' ',normalize-space(@class),' '),' jsp-spawner__size_options__select ')]
+      Wait Until Page Contains Element    xpath://div[contains(concat(' ',normalize-space(@class),' '),' jsp-spawner__size_options__select ')]
    ELSE
-      ${spawner_ready} =    Run Keyword And Return Status   Wait Until Page Contains Element    xpath:/html/body/div[1]/form/div/div/div[3]/div[3]/button
+      Wait Until Page Contains Element    xpath:/html/body/div[1]/form/div/div/div[3]/div[3]/button
    END
-   [Return]  ${spawner_ready}
 
 Select Notebook Image
    [Documentation]  Selects a notebook image based on a partial match of ${notebook_image} argument
@@ -134,7 +133,7 @@ Spawn Notebook With Arguments
    ...              By creating a dictionary beforehand, e.g. &{test-dict}  Create Dictionary  name=robot  password=secret
    [Arguments]  ${retries}=1  ${image}=s2i-generic-data-science-notebook  ${size}=Small  ${spawner_timeout}=600 seconds  &{envs}
    FOR  ${index}  IN RANGE  0  1+${retries}
-      ${spawner_ready} =    JupyterHub Spawner Is Ready
+      ${spawner_ready} =    Run Keyword and Return Status    Wait Until JupyterHub Spawner Is Ready
       IF  ${spawner_ready}==True
          Select Notebook Image  ${image}
          Select Container Size  ${size}
@@ -145,16 +144,7 @@ Spawn Notebook With Arguments
                Add Spawner Environment Variable  ${key}  ${value}
             END
          END
-         ${version-check} =  Is RHODS Version Greater Or Equal Than  1.5.0
-         IF  ${version-check}==True
-            Click Button  Start Server
-            Wait Until Page Contains  Starting server
-         ELSE
-            Click Button  Start server
-            Wait Until Page Contains  Your server is starting up
-         END
-         Wait Until Element is Visible  id:progress-bar
-         Run Keyword And Continue On Failure  Wait Until Page Does Not Contain Element  id:progress-bar  ${spawner_timeout}
+         Spawn Notebook
          ${spawn_fail} =  Has Spawn Failed
          Exit For Loop If  ${spawn_fail} == False
          Click Element  xpath://span[@id='jupyterhub-logo']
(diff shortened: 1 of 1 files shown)


lugi0:
Test run: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/321/

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -123,28 +134,33 @@ Spawn Notebook With Arguments
...
+         END
+         Wait Until Element is Visible  id:progress-bar
+         Run Keyword And Continue On Failure  Wait Until Page Does Not Contain Element  id:progress-bar  ${spawner_timeout}
+         ${spawn_fail} =  Has Spawn Failed
+         Exit For Loop If  ${spawn_fail} == False
+         Click Element  xpath://span[@id='jupyterhub-logo']
       ELSE
-         Click Button  Start server
-         Wait Until Page Contains  Your server is starting up
+         Click Element  xpath://span[@id='jupyterhub-logo']
```
bdattoma, line 162: If the spawner is not ready, should we wait for the other retry instead of clicking on the JH logo? I was thinking that if the JH is not still ready, maybe it is loading, then if it clicks on the logo it will restart loading and we could hit the same "not-ready" problem.

jgarciao: You can easily create a `Wait Until JupyterHub Spawner Is Ready` keyword using the built-in `Wait Until Keyword Succeeds` (which automatically retries running a keyword until a certain timeout is reached). 

I've implemented something similar [here](https://github.com/jgarciao/ods-ci/blob/3c77d3a6cd35453a0c3079b10d1be31ba1d13a5f/tests/Resources/Page/ODH/Prometheus/Prometheus.robot#L88)

jgarciao: I've looked again at `JupyterHub Spawner Is Ready` and I see that it's already waiting for the elements to be loaded. The name of the keyword is a little bit confusing because it doesn't reflect that. 

So, in case you add an additional `Wait Until JupyterHub Spawner Is Ready ` keyword, maybe `JupyterHub Spawner Is Ready` should just check for the elements to be ready but  not wait for them.

bdattoma: @lugi0 got it. Sounds good. Let's see in the long time and if we'll notice some failure due to slow loading we could also raise the timeout in the `JupyterHub Spawner Is Ready`. 

About the @jgarciao 's point,  calling the keyword `Wait Until JupyterHub Spawner Is Ready` might be more consistent with the naming we've adopted so far. correct me if I'm wrong

lugi0: I'm not sure I agree, the point of the keyword is not to wait for the elements, it is to return a `True`/`False` with which you can check if the spawner is ready or not. 
If you've opened the spawner page a few seconds earlier, that keyword won't really wait for anything (because the page is already loaded) and just return a `True`, with which you can do what you need to do; if for some reason the page is not ready yet, it will wait for a bit, and return a `False` in case something went wrong (which you can use to fail a TC with a more meaningful error message based on your TC/use case.

bdattoma: so basically, if the aim is just to check the presence of the element, you could do `Page Should Contain Element` instead of `Wait etc`

lugi0: In that case it might fail if run as soon as the spawner page is opened, or if there are any network issues, and to remove those false negatives I think it's better to have the `Wait` in the keyword rather than the `Page should contain`

jgarciao: That's what I meant before in my comment: in order to follow the naming you could modify `JupyterHub Spawner Is Ready` to use `Page Should Contain` and create an additional `Wait Until JupyterHub Spawner Is Ready` keyword

jgarciao: To me the difference is that `Wait Until JupyterHub Spawner Is Ready` fails if the timeout is reached, instead of returning a boolean. It can simplify the code in `User Is Allowed` and probably other places. Maybe we need just that one

lugi0: Honestly, I still don't agree. It would perhaps save a few lines of code, but I still think that having the boolean would give us more flexibility when dealing with a spawner that is not yet ready (wait some more, reload, fail etc.). In general no TC should really use this keyword directly, as it is more of an helper keyword for the `User is Allowed` and `Spawn Notebook` keywords, so when we actually need it directly in a TC I would expect the return value to be more valuable than fewer lines of code.
In general I would say that we should have almost every keyword returning something so that we can take action based on the return value, rather than letting the keyword fail without control.

jgarciao: I could say that if you need the flexibility create both keywords, but as you are the main user of this code, please do what you think is best and we can always change/enhance it in the future if needed. Thanks Luca!

jgarciao (approved):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -123,28 +134,33 @@ Spawn Notebook With Arguments
...
-         Click Button  Start Server
-         Wait Until Page Contains  Starting server
+         ${version-check} =  Is RHODS Version Greater Or Equal Than  1.5.0
+         IF  ${version-check}==True
+            Click Button  Start Server
+            Wait Until Page Contains  Starting server
+         ELSE
+            Click Button  Start server
+            Wait Until Page Contains  Your server is starting up
+         END
```
tarukumar, line 155: do you think it will be good idea to move line 148~155 in some keyword and call that keyword? I know it will not make much difference just for the easy readability

lugi0: Yeah, that might be better

lugi0:
Updated PR: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/326/

lugi0:
@jgarciao @tarukumar @bdattoma I see in the test run some failures in metrics, plugin verification and rhosak, can you take a look? other than that the run looks good to me.

bdattoma:
> @jgarciao @tarukumar @bdattoma I see in the test run some failures in metrics, plugin verification and rhosak, can you take a look? other than that the run looks good to me.

looked at RHOSAK ones. The problem is that RHOSAK was already enabled in the cluster. I'm going to remove from the suite the TC ODS-392 and ODS-242 because they are now covered by PR #141

jgarciao (approved):
LGTM

aloganat:

bdattoma (approved):

PR #139, opened 2021-12-02
RHODS/ODH keywords compatibility changes

vsokolovskii:
The idea was to created two separate resource files [ODH|RHODS]Variables.resource, each file will contain platform specific variables. The tester can switch between these two files just by setting the ODS_PROJECT variable, which is in the main variable file `test-variables.yml.` I adjusted the keyword for controlling the main dashboard title, because it's the only conflicting keyword I could find so far but if more conflicting stuff will occur it can be easily handled.

PR #175, opened 2022-01-19
Catch ValueError exception in GTE keyword on invalid SemVer

LaVLaS:
Catch the ValueError exception thrown by the `GTE` keyword when the `version` argument is invalid or null.

### DIFF
--- a/libs/Helpers.py
+++ b/libs/Helpers.py
@@ -12,8 +12,14 @@ class Helpers:
 
     @keyword
     def gte(self, version, target):
-        version=VersionInfo.parse(version)
-        target=VersionInfo.parse(target)
-        #version=tuple(version.translate(str.maketrans('', '', string.punctuation)))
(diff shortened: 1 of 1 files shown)


crobby (approved):
looks reasonable to me

lugi0 (approved):
lgtm as a quick fix for the problem, ideally I would like to see this handled differently, I'm not sure we always want an invalid SemVer failing silently and trying to use the old UI in the TCs.

LaVLaS:
> lgtm as a quick fix for the problem, ideally I would like to see this handled differently, I'm not sure we always want an invalid SemVer failing silently and trying to use the old UI in the TCs.

I agree. Ideally there we would be a check for a valid version before sending it to `GTE` if the test is running in RHODS. Since this PR was a reaction to failing tests for ODH, WE hope to follow this up with additional PRs, like #139, to allow ODH test to re-use common automation

LaVLaS:
rhods-ci-pr-test/331 PR CI test rand successfully

PR #178, opened 2022-01-20
ignore teardown failure on Splash Page

bdattoma:
Teardown in ODS-248 may fail if the Splash (loading) page doesn't appear making the entire TC failing (but all the TC steps are successful). So this PR wants to address the problem by adding a `Run Keyword And Ignore Error`

### DIFF
--- a/tests/Resources/Page/HybridCloudConsole/Rhosak.robot
+++ b/tests/Resources/Page/HybridCloudConsole/Rhosak.robot
@@ -183,7 +183,7 @@ Clean Up RHOSAK
   Enter Stream Topics Section
   Delete Stream Topic  topic_to_delete=${topic_to_delete}
   Menu.Navigate To Page    Streams for Apache Kafka   Kafka Instances
-  Wait For HCC Splash Page
+  Run Keyword And Ignore Error    Wait For HCC Splash Page
   Wait Until Page Contains    Create Kafka instance
   Delete Kafka Stream Instance  stream_name=${stream_to_delete}
   Capture Page Screenshot  after deleting_stream.png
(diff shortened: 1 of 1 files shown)


jgarciao (dismissed):
LGTM

tests/Resources/Page/HybridCloudConsole/Rhosak.robot
```diff
@@ -183,7 +183,7 @@ Clean Up RHOSAK
   Enter Stream Topics Section
   Delete Stream Topic  topic_to_delete=${topic_to_delete}
   Menu.Navigate To Page    Streams for Apache Kafka   Kafka Instances
-  Wait For HCC Splash Page
+  Run Keyword And Ignore Error    Wait For HCC Splash Page
```
lugi0, line 186: Maybe Continue on Failure would be better?

bdattoma: `Continue on Failure` would make the TC fail

lugi0: Yes we would keep seeing the failure but at least we have a reminder that something wrong is going on there while still executing the TC

bdattoma: well, as it is done now, it continue on failure automatically (because it is a teardown). The point is that if the splash page does not appear, it doesn't mean the TC will fail in this case, but it can success anyway (as it is happening). If the splash page keyword fails when it shouldn't, then the next steps would fail making the TC correctly failing

aloganat:

bdattoma:
Jenkins Job (success): [PRJOB#350](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/350/)

lugi0 (approved):
lgtm

tarukumar (approved):
LGTM

PR #182, opened 2022-01-25
[WIP] Update library version checks to x.y

lugi0:

### DIFF
--- a/tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
+++ b/tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
@@ -26,6 +26,19 @@ Launch JupyterLab
   &{S3-credentials} =  Create Dictionary  AWS_ACCESS_KEY_ID=${S3.AWS_ACCESS_KEY_ID}  AWS_SECRET_ACCESS_KEY=${S3.AWS_SECRET_ACCESS_KEY}
   Spawn Notebook With Arguments  image=s2i-generic-data-science-notebook  envs=&{S3-credentials}
 
+Verify Installed Libraries in SDS
+  [Tags]  Sanity
+  Python Version Check
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show boto3 | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  1.17
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show kafka-python | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  2.0
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show matplotlib | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  3.4
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show numpy | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  1.19
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show pandas | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  1.2
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show scikit-learn | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  0.24
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show scipy | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  1.6
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show jupyterlab | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  3.2
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show notebook | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  6.4
+
 Long Running Test Case
   Run Repo and Clean  https://github.com/lugi0/minimal-nb-image-test  minimal-nb-image-test/minimal-nb.ipynb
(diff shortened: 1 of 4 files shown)


tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
```diff
@@ -26,6 +26,19 @@ Launch JupyterLab
   &{S3-credentials} =  Create Dictionary  AWS_ACCESS_KEY_ID=${S3.AWS_ACCESS_KEY_ID}  AWS_SECRET_ACCESS_KEY=${S3.AWS_SECRET_ACCESS_KEY}
   Spawn Notebook With Arguments  image=s2i-generic-data-science-notebook  envs=&{S3-credentials}
 
+Verify Installed Libraries in SDS
+  [Tags]  Sanity
```
jgarciao, line 30: Polarion ID missing here?

tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
```diff
@@ -26,6 +26,19 @@ Launch JupyterLab
...
+  Python Version Check
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show boto3 | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  1.17
```
jgarciao, line 32: For clarity you could create a `Python Package Should Be Installed` keyword and make the call like this:
`Run Keyword And Continue On Failure     Python Package Should Be Installed    boto3   1.17`

Why do you `Run Keyword And Continue On Failure `. Don't you want to fail if requirements are not met?

tests/Tests/500__jupyterhub/minimal-tensorflow-test.robot
```diff
@@ -44,24 +44,24 @@ Verify Installed Libraries in Tensorflow
...
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show matplotlib | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  3.4
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show numpy | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  1.19
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show pandas | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  1.2
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show scikit-learn | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  0.24
+  Run Keyword And Continue On Failure  Run Cell And Check Output  !pip show scipy | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  1.6
   ${version-check} =  Is RHODS Version Greater Or Equal Than  1.4.0
```
jgarciao, line 54: In another PR (something related to the refactoring effort) we should find a way to specify the requirements in a file for each version (for example requirements-rhods-1.5.0-tensorflow.txt) and simplify the test to use it 

Related links:
- https://stackoverflow.com/questions/16294819/check-if-my-python-has-all-required-packages
- https://github.com/pypa/pip/issues/2733#issuecomment-257340871

tests/Tests/500__jupyterhub/test-minimal-image.robot
```diff
@@ -34,6 +34,11 @@ Can Spawn Notebook
   Fix Spawner Status
   Spawn Notebook With Arguments  image=s2i-minimal-notebook
 
+Verify Installed Libraries in Minimal
```
jgarciao, line 37: Missing Tags (Polarion ID, ...)

tests/Tests/500__jupyterhub/minimal-tensorflow-test.robot
lugi0: I am actually preparing another PR to run all these checks in a single TC rather than in 4 different ones. In that case I am grabbing the packages and versions from the spawner page, so we will dinamically be checking if that information and the installed libraries match. That should help in this regard

tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
lugi0: For point #1 see the other comment, point #2 no I don't want the TC to fail because of that, at least in this setup. There is other testing being done after the package version checks and it's not worth it to skip it entirely if this check fails. The failure is still shown in the run logs so we are still notified of any problems with the packages

tests/Tests/500__jupyterhub/long-running-test-generic-ds.robot
lugi0: Right now there are 4 TCs for 1 polarion ID, so I'm not sure how to handle the many-to-one relationship

bdattoma: I think it's not a problem now if we have more than one TC for 1 polarion ID (if each TC implement a part of that). If it is urgent to have this PR merged, I think you can assign the same polarion ID tag. After that, we'll have this fixed by the PR you're preparing. 
Would it work for you?

lugi0: I think the problem with that will be that the last test executed will overwrite the results in Polarion for all the others, so if the last one fails the Polarion TC is marked as a Fail (correct), but if it passes it will be marked as a Pass, which might not be correct depending on the other three TCs

bdattoma: right, then use a prefix/suffix. e.g., `ODS-123-part1`.

bdattoma: right. we could create different copies of the same TC in polarion. temporary solution

lugi0: We could but I don't like it that much as an option to be honest with you. I should have the other PR ready by today, so we could just hold off on this one

jgarciao: Mapping of PolarionIDs with ods-ci tests is one of the things to fix in our upcoming refactoring effort so if you want to merge it like this now and fix it in the future is fine for me

bdattoma: @lugi0 I don't like it to too. I meant that as temporary solution if there is urgency to merge this PR. If the other PR is ready in short time, than we could close this and use the new coming one

jgarciao (approved):
LGTM as temporal change and to be improved in future PRs

lugi0:
Do not merge for now, thanks

lugi0:
Closed in favor of #184

PR #184, opened 2022-01-27
Unify version check TC

lugi0:

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
@@ -341,18 +341,31 @@ Verify Installed Library Version
     [Arguments]  ${lib}  ${ver}
     ${status}  ${value} =  Run Keyword And Warn On Failure  Run Cell And Check Output  !pip show ${lib} | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  ${ver}
     Run Keyword If  '${status}' == 'FAIL'  Log  "Expected ${lib} at version ${ver}, but ${value}"
+    [Return]    ${status}
 
 Check Versions In JupyterLab
-    [Arguments]  ${list}
-    FOR  ${str}  IN  @{list}
-        @{info} =  Split String  ${str}  ${SPACE}v
-        IF  "${info}[0]" == "TensorFlow"
-            Verify Installed Library Version  tensorflow-gpu  ${info}[1]
-        ELSE IF  "${info}[0]" == "PyTorch"
-            Verify Installed Library Version  torch  ${info}[1]
-        ELSE IF  "${info}[0]" == "Python"
-            Python Version Check  ${info}[1]
+    [Arguments]  ${libraries-to-check}
+    ${return_status} =    Set Variable    PASS
+    FOR  ${libString}  IN  @{libraries-to-check}
+        # libString = LibName vX.Y -> libDetail= [libName, X.Y]
+        @{libDetail} =  Split String  ${libString}  ${SPACE}v
+        IF  "${libDetail}[0]" == "TensorFlow"
+            ${status} =  Verify Installed Library Version  tensorflow-gpu  ${libDetail}[1]
+            IF  '${status}' == 'FAIL'
+              ${return_status} =    Set Variable    FAIL
+            END
+        ELSE IF  "${libDetail}[0]" == "PyTorch"
+            ${status} =  Verify Installed Library Version  torch  ${libDetail}[1]
+            IF  '${status}' == 'FAIL'
+              ${return_status} =    Set Variable    FAIL
+            END
+        ELSE IF  "${libDetail}[0]" == "Python"
+            ${status} =  Python Version Check  ${libDetail}[1]
         ELSE
-            Verify Installed Library Version  ${info}[0]  ${info}[1]
+            ${status} =  Verify Installed Library Version  ${libDetail}[0]  ${libDetail}[1]
+            IF  '${status}' == 'FAIL'
+              ${return_status} =    Set Variable    FAIL
+            END
         END
     END
+    [Return]  ${return_status}
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
@@ -340,7 +340,7 @@ Select ${filename} Tab
 Verify Installed Library Version
     [Arguments]  ${lib}  ${ver}
     ${status}  ${value} =  Run Keyword And Warn On Failure  Run Cell And Check Output  !pip show ${lib} | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  ${ver}
-    Run Keyword If  '${status}' == 'FAIL'  Log To Console  "Expected ${lib} at version ${ver}, but ${value}"
+    Run Keyword If  '${status}' == 'FAIL'  Log  "Expected ${lib} at version ${ver}, but ${value}"
 
 Check Versions In JupyterLab
     [Arguments]  ${list}
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
@@ -143,16 +143,10 @@ Clean Up Server
   Sleep  1
   Maybe Close Popup
   Open With JupyterLab Menu  File  New  Notebook
-  Sleep  1
+  Sleep  2
   Maybe Close Popup
-  Open With JupyterLab Menu  File  Open from Path…
-  Input Text  xpath=//input[@placeholder="/path/relative/to/jlab/root"]  Untitled.ipynb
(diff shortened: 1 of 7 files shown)


tests/Tests/500__jupyterhub/test-versions.robot
```diff
@@ -0,0 +1,48 @@
...
+        ${list} =  Combine Lists  ${list}  ${tmp}
+        Log  ${list}
+        Spawn Notebook With Arguments  image=${img}
+        Check Versions In JupyterLab  ${list}
+        Clean Up Server
+        Stop JupyterLab Notebook Server
+        Go To  ${ODH_DASHBOARD_URL}
+        Wait for RHODS Dashboard to Load
+        Launch JupyterHub From RHODS Dashboard Link
+        Wait Until JupyterHub Spawner Is Ready
```
tarukumar, line 47: Do we need to have line 44~47. should we just start JL server after it has stopped. it will help use not to go back to RHODS dashboard and lunch JH from there

tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
```diff
@@ -143,16 +143,10 @@ Clean Up Server
   Sleep  1
   Maybe Close Popup
   Open With JupyterLab Menu  File  New  Notebook
-  Sleep  1
-  Maybe Close Popup
-  Open With JupyterLab Menu  File  Open from Path…
-  Input Text  xpath=//input[@placeholder="/path/relative/to/jlab/root"]  Untitled.ipynb
-  Click Element  xpath://div[.="Open"]
+  Sleep  2
```
bdattoma, line 146: should the Sleep be part of the "Maybe Close Popup" keyword since it seems to be needed everytime the "Maybe.." is called

tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
```diff
@@ -342,3 +336,23 @@ Get JupyterLab Code Output In a Given Tab
...
+Verify Installed Library Version
+    [Arguments]  ${lib}  ${ver}
+    ${status}  ${value} =  Run Keyword And Warn On Failure  Run Cell And Check Output  !pip show ${lib} | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s", b[1], b[2]}'  ${ver}
+    Run Keyword If  '${status}' == 'FAIL'  Log  "Expected ${lib} at version ${ver}, but ${value}"
+
+Check Versions In JupyterLab
+    [Arguments]  ${list}
+    FOR  ${str}  IN  @{list}
+        @{info} =  Split String  ${str}  ${SPACE}v
+        IF  "${info}[0]" == "TensorFlow"
```
bdattoma, line 349: mmh maybe you can give more self-explanatory names to these variables (i.e., list, info)

tests/Tests/500__jupyterhub/test-versions.robot
```diff
@@ -0,0 +1,48 @@
...
+
+*** Variables ***
+@{notebook_images}  s2i-minimal-notebook  s2i-generic-data-science-notebook  tensorflow  pytorch
+
+*** Test Cases ***
+Verify Installed Libraries
+  [Tags]  Sanity
+  ...     ODS-340  ODS-695  ODS-204  ODS-205  ODS-206  ODS-207  ODS-215  ODS-216  ODS-217  ODS-218
+  Wait for RHODS Dashboard to Load
+  ${version-check} =  Is RHODS Version Greater Or Equal Than  1.4.0
```
bdattoma, line 19: it might be worth to have a common keyword containing lines from 19 to 24

tests/Tests/500__jupyterhub/test-versions.robot
bdattoma: The keyword `Fix Spawner Status` could be used to replace from 42 to 47

tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
lugi0: It could be, but I also don't want this PR to change more than it needs to; that can be done as part of a refactoring effort.

tests/Tests/500__jupyterhub/test-versions.robot
lugi0: Same as above, it could be done but I'd like to avoid changing more than what needs to be changed with this PR. Further changes can be applied during refactoring.

tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
lugi0: Agreed, will do

tests/Tests/500__jupyterhub/test-versions.robot
bdattoma: ok

tests/Tests/500__jupyterhub/test-versions.robot
```diff
@@ -9,40 +9,69 @@ Suite Setup      Begin Web Test
...
+    Append To List  ${status_list}  ${status}
+
+Verify Libraries in Tensorflow Image
+    [Tags]  ODS-204  ODS-205  ODS-206  ODS-207
+    @{additional_libs} =    Create List
+    Append To List    ${additional_libs}    JupyterLab v3.2    Notebook v6.4
+    ${status} =  Verify Libraries In Base Image    tensorflow    ${additional_libs}
+    Append To List  ${status_list}  ${status}
+
+Verify All Images And Spawner
```
bdattoma, line 54: why creating a TC to check the results from the previous ones? Isn't possible to do the status check in each TC?

lugi0: Done like this to have the Polarion ID in a single TC; the test needs to check that all spawner descriptions match with all installed versions.

bdattoma: hmm yeah I got it..an alternative maybe would have been to turn all the single TCs (e.g., `Verify Libraries in Tensorflow Image`) into keywords, and write `Verify All Images And Spawner` as a list of checks (e.g., `Run Keyword and Continue on Failure`). I'm saying this because otherwise we have in the logs many TCs (one for each image) performing the checks, and then an additional TC which repeats the results of the signle Tcs, without adding value.

But, just one opinion. I would still agree with the current structure. And if in the future we'll see it doesn't fit, we could change

lugi0: If I understood your suggestion correctly, I don't think it would work because we would end up doing all the checks (i.e. calling all the keywords) from one TC, which would bring us back to the previous problem I was trying to solve (all Polarion IDs linked to a single TC). If that's not what you meant, I think the other option would be to create different keywords and different TCs (where each TC only calls one keyword) but that's essentially the same as it is now, the only difference is that the "logic" is in the keyword rather than the TC. 

Let me know if I have misunderstood anything, otherwise I think this is the better approach for now

bdattoma: yeah you got what I meant. But I thought the initial problem was to have the same Polarion ID assigned to multiple TCs. As of now we have different case where each TC has more than once Polarion IDs. 
IMO, the real solution to the problem is to review all the polarion TC to remove redundancy. 

With the current implementation here, I'm not sure about the sense of having the same check done in two different TCs. And you can have that when the check fails, the single-image TC (which only returns the status) is still SUCCESS, while the cumulative TC is FAILED. hmmm correct me if I'm wrong

lugi0: Fixed the last point in the newest commit

lugi0:
Tested locally and works, Jenkins currently unavailable to launch a PR test

lugi0:
PR job: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/365/

bdattoma (approved):

bdattoma:
> PR job: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/365/

the `Verify Libraries in Tensorflow Image` is failing because of a missing element. Is that expected?

lugi0:
It failed because it could not find `Element with locator '//input[contains(@id, "tensorflow")]/../span' not found.`, seems like a timing issue or similar

lugi0:
Additional PR test: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/371

bdattoma (approved):

bdattoma:
lgtm

tests/Tests/500__jupyterhub/test-versions.robot
```diff
@@ -0,0 +1,83 @@
+*** Settings ***
+Force Tags       Sanity
+Resource         ../../Resources/ODS.robot
+Resource         ../../Resources/Common.robot
+Resource         ../../Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+Resource         ../../Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
```
tarukumar, line 6: could you replace above two line with .resource file or would you like to do that during refactoring?

tarukumar (approved):
LGTM

jgarciao (approved):
LGTM

PR #187, opened 2022-02-01
Hotfix patch for rhods operator

tarukumar:
Added missing colon for the error during command execution.

### DIFF
--- a/tests/Tests/100__deploy/130__operators/130__rhods_operator/135__rhods_operator_logs_verification.robot
+++ b/tests/Tests/100__deploy/130__operators/130__rhods_operator/135__rhods_operator_logs_verification.robot
@@ -23,7 +23,7 @@ Verify RHODS Operator log
    #Capture the logs based on containers
    ${val}        Run   oc logs --tail=1000000 ${data[0]['metadata']['name']} -n ${namespace} -c rhods-operator
    #To check if command has been suessfully executed and the logs has been captured
-   Run Keyword IF    len($val)==${0} or "error:" in $val     FAIL   Either OC command has not been executed suessfully or Logs is not present
+   Run Keyword IF    len($val)==${0} or "error:" in $val     FAIL   Either OC command has not been executed sucessfully or Logs is not present
    #Filter the error msg from the log captured
    ${match_list} 	 Get Regexp Matches   ${val}     ${regex_pattern}
    #Remove if any duplicate entry are present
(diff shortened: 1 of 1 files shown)


tests/Tests/100__deploy/130__operators/130__rhods_operator/135__rhods_operator_logs_verification.robot
```diff
@@ -23,7 +23,7 @@ Verify RHODS Operator log
    #Capture the logs based on containers
    ${val}        Run   oc logs --tail=1000000 ${data[0]['metadata']['name']} -n ${namespace} -c rhods-operator
    #To check if command has been suessfully executed and the logs has been captured
-   Run Keyword IF    len($val)==${0} or "error" in $val     FAIL   Either OC command has not been executed suessfully or Logs is not present
+   Run Keyword IF    len($val)==${0} or "error:" in $val     FAIL   Either OC command has not been executed suessfully or Logs is not present
```
milindw96, line 26: is it a good idea to add regex example "error*" so that in the future we may not face the same issue

tarukumar: i don't think so...i we use error or error* we will be capturing every error present in the sentences as well  which might be not be the actual error. the idea to have here is that when we run oc command it succeed without any error. When we use wrong command we get error like below:

ex: error:*

bdattoma (dismissed):

tests/Tests/100__deploy/130__operators/130__rhods_operator/135__rhods_operator_logs_verification.robot
jgarciao, line 26: There is a typo in "successfully"

LGTM as a hotfix. If you want to improve it after maybe you can check the return status differently using the keyword  
 `Run Command` :

```
${result} =  Run Process    oc logs --tail=1000000 ${data[0]['metadata']['name']} -n ${namespace} -c rhods-operator
Should Be Equal As Integers  ${result.rc}  0
```
More info at  https://robotframework.org/robotframework/latest/libraries/Process.html

jgarciao (approved):

bdattoma (approved):

tests/Tests/100__deploy/130__operators/130__rhods_operator/135__rhods_operator_logs_verification.robot
tarukumar: Done...For now i have kept the same ..will see in future if we need to use Process library

tarukumar:
Verifies using jenkins [Jobid-356](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/356/console)..TC is working as exepected and it has failed because it has found error entry in the operator logs(RHODS-2857).There are some other failure which are not caused by his code change

lugi0 (approved):
lgtm

PR #192, opened 2022-02-03
update test-variables example file

bdattoma:

### DIFF
--- a/test-variables.yml.example
+++ b/test-variables.yml.example
@@ -19,3 +19,15 @@ OCP_ADMIN_USER:
 S3:
   AWS_ACCESS_KEY_ID: ID-value
   AWS_SECRET_ACCESS_KEY: Secret-Key
+ANACONDA_CE:
+  ACTIVATION_KEY: license-key
+SSO:
+  USERNAME: rh-sso-account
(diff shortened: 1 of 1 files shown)


tarukumar (approved):

test-variables.yml.example
```diff
@@ -19,3 +19,15 @@ OCP_ADMIN_USER:
 S3:
   AWS_ACCESS_KEY_ID: ID-value
   AWS_SECRET_ACCESS_KEY: Secret-Key
+ANACONDA_CE:
+  ACTIVATION_KEY: license-key
+SSO:
+  USERNAME: rh-sso-account
+  PASSWORD: rh-sso-account-pw
+RHOSAK_CONFIG_TEST:
```
jgarciao, line 27: Is `_CONFIG_TEST `needed?    Otherwise LGTM

jgarciao (approved):

test-variables.yml.example
bdattoma: I thought it would have been more self-explanatory while calling it in the code. But yeah it's not extremely important

bdattoma: if you want I can change it, but IMO the effort is not worth now to do it

PR #193, opened 2022-02-03
request and limit size verification

tarukumar:
This TC Verify spawned server pod has the correct resource requests/limits
a. verification is done for small and medium size
b. It fetch the container size from JH spanwer page. Format it and then convert to dictionary
c. capture the container resource for notebook pod
d. Compare both of them


lugi0 (approved):
lgtm

lugi0 (commented):
Changing my approval to a comment, I was thinking about this PR and there are a couple of things that I think can be added to it:

1. Other than Small and Medium, also verify the same steps for the `Default` size. Along with that, I would also suggest checking the size when `Default` is set to, well, the default value, as well as after modifying it to make sure that the new requests/limits are picked up (This should cover another Polarion ID)

2. Verify that the spawner does not override the size selection: This can be done rather easily by selecting a size, refreshing the page, and spawning the server. If after refreshing the server is still spawned with the correct size, then the spawner is not overriding the selection upon page load. This is also another Polarion ID.

tarukumar:
Verified with Jenkins [job-380](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/380). one TC is failed due to insufficient node, however the implementation works fine. There are are some other failure whihc is not caused by this.

tarukumar:
Re-verified the [Jenkins job-389](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/389/) all the Tc passed. There are some failures not related to this PR.

lugi0 (dismissed):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -132,20 +132,24 @@ Spawn Notebook With Arguments
...
    ...              It will try again ${retries} times (Default: 1). Environment variables can be passed in as kwargs
    ...              By creating a dictionary beforehand, e.g. &{test-dict}  Create Dictionary  name=robot  password=secret
-   [Arguments]  ${retries}=1  ${image}=s2i-generic-data-science-notebook  ${size}=Small  ${spawner_timeout}=600 seconds  &{envs}
+   [Arguments]  ${retries}=1  ${image}=s2i-generic-data-science-notebook  ${size}=Small  ${spawner_timeout}=600 seconds  ${refresh}=${False}  &{envs}
    FOR  ${index}  IN RANGE  0  1+${retries}
       ${spawner_ready} =    Run Keyword and Return Status    Wait Until JupyterHub Spawner Is Ready
       IF  ${spawner_ready}==True
          Select Notebook Image  ${image}
          Select Container Size  ${size}
+         IF   ${refresh}
```
bdattoma, line 141: why refreshing?

tests/Tests/500__jupyterhub/notebook_size_verification.robot
```diff
@@ -0,0 +1,92 @@
...
+
+*** Test Cases ***
+Verify Spwaned Notebook Size
+    [Documentation]    Check the available container size spec
+    ...    with actual assign to spwaned notebook pod
+    [Tags]    Sanity    ODS-1072
+    Launch JupyterHub Spawner From Dashboard
+    Spawn Notebook And Verify Size
+
+Verify Custome Spwaned Notebook Size
```
bdattoma, line 38: there's a typo

tests/Tests/500__jupyterhub/notebook_size_verification.robot
```diff
@@ -0,0 +1,92 @@
...
+    [Documentation]    Modify and verify the default notebook conatiner size spec
+    ...    with spwaned notebook pod and set back to default size
+    [Tags]    Sanity    ODS-318
+    ...       ODS-1071
+    Launch JupyterHub Spawner From Dashboard
+    Modify Default Container Size
+    ${d_continer_size}    Create List    Default
+    Spawn Notebook And Verify Size    size=${CUSTOME_SIZE}    NOTEBOOK_SIZE=${d_continer_size}
```
bdattoma, line 46: typo in `continer_size`

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
tarukumar: we want to reload page after selecting the container size to check if after selecting it is it going back to default or getting modified

lugi0: to test that the size selection is not overwritten by the spawner I think

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -132,20 +132,24 @@ Spawn Notebook With Arguments
...
+         END
          IF  &{envs}
             Remove All Spawner Environment Variables
             FOR  ${key}  ${value}  IN  &{envs}[envs]
                Sleep  1
                Add Spawner Environment Variable  ${key}  ${value}
             END
          END
-         Spawn Notebook
+         Wait Until Keyword Succeeds    30    1    Spawn Notebook
```
lugi0, line 152: I think you need to specify what `30` is, otherwise it might be confusing.
You're also giving a 30s timeout to a keyword that has a 600s timeout by default: https://github.com/red-hat-data-services/ods-ci/blob/master/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot#L110

I think this is not needed, you can use the regular `Spawn Notebook` and increase the `retries` of this keyword if you think it can fail

tests/Tests/500__jupyterhub/notebook_size_verification.robot
tarukumar: Done

lugi0 (changes requested):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
tarukumar: I disagree this 30 second is for the whole keyword to succeed Even if i try to increase the retires to the keyword it doesn't help the purpose because retry only come into the picture while checking the progress bar. This keyword will fail before that exactly at click button start server. I agree to add 's' for more clarity

lugi0: >This keyword will fail before that exactly at click button start server

I have never seen it fail when clicking that button, but it might fail due to JH not spawning a server (either timeout or other failure), but for that you already have the 600s timeout in the `Spawn Notebook` keyword itself. I don't understand the need for an additional timeout on top of it?

tarukumar: In normal case it doesn't fail but when you reload page it take sometime to get reload if i don't give timeout, Start server button will not be available because page is still in reloading state. 

Coming back to 600s timeout if you look at the spawn keyword this 600s is only for the **id progress** which is the last line. If i don't give timeout it will fail in _start button_ (mentioned in italic)

Spawn Notebook
   ${version-check} =  Is RHODS Version Greater Or Equal Than  1.5.0
   IF  ${version-check}==True
      _Click Button  Start Server_
      Wait Until Page Contains  Starting server
   ELSE
      _Click Button  Start server_
      Wait Until Page Contains  Your server is starting up
   END
   Wait Until Element is Visible  id:progress-bar
   **Wait Until Page Does Not Contain Element  id:progress-bar  ${spawner_timeout}**

lugi0 (approved):
lgtm

bdattoma (approved):

PR #201, opened 2022-02-10
check if it is possible to filter items by enable state or application

rkamble5:
Filter resources by enable state on the resource page.

Signed-off-by: rkamble@redhat.com <rkamble@redhat.com>

### DIFF
--- a/tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
+++ b/tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
@@ -11,44 +11,6 @@ Test Teardown   Dashboard Test Teardown
 ${RHOSAK_REAL_APPNAME}=         rhosak
 ${RHOSAK_DISPLAYED_APPNAME}=    OpenShift Streams for Apache Kafka
 
-Check All Links Are working
-  [Documentation]  Verify the all links after appling a filter
-  ${link_elements}=  Get WebElements    //a[@class="odh-card__footer__link" and not(starts-with(@href, '#'))]
-  ${len}=  Get Length    ${link_elements}
-  Log To Console    ${len} Links found\n
-  FOR  ${idx}  ${ext_link}  IN ENUMERATE  @{link_elements}  start=1
-      ${href}=  Get Element Attribute    ${ext_link}    href
-      ${status}=  Get HTTP Status Code   ${href}
-      Log To Console    ${idx}. ${href} gets status code ${status}
-  END
-
-Check For enabled option
-  [Documentation]  Filter Resources by enabling "enabled" from enable state
-  ${link_elements}=  Get WebElements    //a[@class="odh-card__footer__link"]
-  ${len}=  Get Length    ${link_elements}
-  Select Checkbox  id=enabled-filter-checkbox--check-box
-  sleep  10s
-  ${link_elements}=  Get WebElements    //a[@class="odh-card__footer__link"]
-  ${lent}=  Get Length    ${link_elements}
-  Should Not Be Equal As Integers  ${len}  ${lent}
-  Check All Links Are working
-  Click Element  id=enabled-filter-checkbox--check-box
-  sleep  5s
-
-Check for not enabled option
-  [Documentation]  Filter Resources by enabling "not enabled" from enable state
-  ${link_elements}=  Get WebElements    //a[@class="odh-card__footer__link"]
-  ${len}=  Get Length    ${link_elements}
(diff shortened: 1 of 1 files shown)


tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
```diff
@@ -20,6 +20,14 @@ Verify Resource Link Http status code
         Log To Console    ${idx}. ${href} gets status code ${status}
     END
 
+Go to RHODS Dashboard, Resources And Filter By enabled state
+  [Documentation]  check if it is possible to filter items by enable state or application
+  [Tags]  ODS-489
+  Click Link    Resources
+  sleep  10s
```
bdattoma, line 27: would it be possible to replace the sleep with some other logic? maybe waiting for some page component to appear

tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
```diff
@@ -42,4 +49,49 @@ Dashboard Test Setup
   Wait for RHODS Dashboard to Load
 
 Dashboard Test Teardown
-  Close All Browsers
\ No newline at end of file
+  Close All Browsers
+  
+Get HTTP Status Code
```
bdattoma, line 54: there is `Check HTTP Status Code` keyword in tests/Resources/page/ODH/ODHDashboard/ODHDashboard.robot. I think it is better to use it to avoid duplication

tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
```diff
@@ -42,4 +49,49 @@ Dashboard Test Setup
...
+    [Arguments]  ${link_to_check}
+    ${response}=    RequestsLibrary.GET  ${link_to_check}   expected_status=any
+    Run Keyword And Continue On Failure  Status Should Be  200
+    [Return]  ${response.status_code}
+
+Check All Links Are working
+  [Documentation]  Verify the all links after appling a filter
+  ${link_elements}=  Get WebElements    //a[@class="odh-card__footer__link" and not(starts-with(@href, '#'))]
```
bdattoma, line 62: you don't need to check these links because there are already verified by `Verify Resource Link Http status code` test case in the same suite

tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
```diff
@@ -42,4 +49,49 @@ Dashboard Test Setup
...
+
+Check For enabled option
+  [Documentation]  Filter Resources by enabling "enabled" from enable state
+  ${link_elements}=  Get WebElements    //a[@class="odh-card__footer__link"]
+  ${len}=  Get Length    ${link_elements}
+  Select Checkbox  id=enabled-filter-checkbox--check-box
+  sleep  10s
+  ${link_elements}=  Get WebElements    //a[@class="odh-card__footer__link"]
+  ${lent}=  Get Length    ${link_elements}
+  Should Not Be Equal As Integers  ${len}  ${lent}
```
bdattoma, line 79: I think it may make sense to check also that after the filter has been applied, you get the expected items. So for example, if the only application enabled is Anaconda, I would expect to find only items about Anaconda when using this filter

tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
```diff
@@ -42,4 +49,49 @@ Dashboard Test Setup
...
+  Check All Links Are working
+  Click Element  id=enabled-filter-checkbox--check-box
+  sleep  5s
+
+Check for not enabled option
```
bdattoma, line 84: as per above comment, I would check here if selecting this filter I get the expected items

bdattoma:
I think the TC would also need to check other filters in the Resources page other than "enabled status". So, for example, filtering by application provider etc

tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
```diff
@@ -20,6 +20,14 @@ Verify Resource Link Http status code
         Log To Console    ${idx}. ${href} gets status code ${status}
     END
 
+Go to RHODS Dashboard, Resources And Filter By enabled state
```
bdattoma, line 23: I would change the test title with something more like `Verify Use Can Filter Resources By Enabled State`. 
In addition, I think it would be good to verify other filters other than Enable state

rkamble5:
Sorry I am closing this PR because it is one month older and I tried but it was not rebased by me, difficult resolve the conflicts

PR #166, opened 2022-01-11
Improvements to run better inside a container

erwangranger:
There are a lot of changes in this pull request. 

Mainly:
* getting the dockerfile to work again 
* by adding the `oc` binary to it
* doing an `oc login` based on the admin credentials available in the variable file. 
* some pieces only run if `yq` is available 

If this changes too many things at once, let me know, and I can make a PR with a narrower scope.

PR #205, opened 2022-02-11
[ods-ci container] fix test selection features and other fixes

bdattoma:
This PR wants to fix the issue reported in PR https://github.com/red-hat-data-services/ods-ci/pull/166#discussion_r782735501 

- _Test selection_: The idea is to use two different Env variables to handle args for  `run_robot_test` and `robot` commands
To select a subset of tests it is possible to run the container adding the env variables `ROBOT_EXTRA_ARGS` like this:
`podman run --rm -e ROBOT_EXTRA_ARGS='-i Smoke'  [...] `
- _Other fixes_: base image changed in dockerfile, intergated fix from #166 to connect to ocp cluster in run script. 

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -8,6 +8,10 @@ RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.n
     dnf install -y git python3 unzip chromium chromedriver redhat-lsb-core &&\
     dnf clean all
 
+## Install yq in the image
+RUN curl -L https://github.com/mikefarah/yq/releases/download/v4.16.2/yq_linux_amd64 -o /usr/bin/yq &&\
+    chmod +x /usr/bin/yq
+
 ## Install oc in the container
 COPY build/oc /usr/local/bin/oc
 # COPY oc /usr/local/bin/oc
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,8 +1,8 @@
 FROM centos:centos8
 
 # Use this build arg to set any default test script arguments
-ARG RUN_SCRIPT_ARGS=
 ENV RUN_SCRIPT_ARGS=${RUN_SCRIPT_ARGS}
+ENV ROBOT_EXTRA_ARGS=${ROBOT_EXTRA_ARGS}
 
 ENV HOME /tmp
 
@@ -23,4 +23,9 @@ RUN chgrp -R 0 /tmp && chmod -R g=u /tmp
 
 USER 1001
 
-CMD ./run_robot_test.sh --skip-pip-install ${RUN_SCRIPT_ARGS}
+CMD if [ -n "${RUN_SCRIPT_ARGS}" ]; then \
+	        ./run_robot_test.sh --skip-pip-install ${RUN_SCRIPT_ARGS} --extra-robot-args "'${ROBOT_EXTRA_ARGS}'" ;\
+     else \
+            ./run_robot_test.sh --skip-pip-install ${RUN_SCRIPT_ARGS};\
+     fi
+
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,5 @@
-FROM registry.access.redhat.com/ubi8/ubi:8.1
+FROM quay.io/centos/centos:stream8
+# FROM registry.access.redhat.com/ubi8/ubi:8.1
 # Use this build arg to set any default test script arguments
 ENV RUN_SCRIPT_ARGS=${RUN_SCRIPT_ARGS}
 ENV ROBOT_EXTRA_ARGS=${ROBOT_EXTRA_ARGS}
@@ -16,26 +17,23 @@ RUN curl -L https://github.com/mikefarah/yq/releases/download/v4.16.2/yq_linux_a
 COPY build/oc /usr/local/bin/oc
 # COPY oc /usr/local/bin/oc
 RUN chmod 755 /usr/local/bin/oc &&\
-         oc version --client &&\
-         oc config set-context loadtest
+         oc version --client
 
 RUN mkdir $HOME/ods-ci
 # Change the WORKDIR so the run script references any files/folders from the root of the repo
 WORKDIR $HOME/ods-ci
 
-COPY requirements.txt setup.py .
-RUN python3 -m venv venv && source venv/bin/activate &&  pip3 install --upgrade pip && venv/bin/pip3 install -r requirements.txt
-
 COPY tests tests/
 COPY libs libs/
 COPY run_robot_test.sh  .
+COPY requirements.txt setup.py .
+RUN python3 -m venv venv && source venv/bin/activate &&  pip3 install --upgrade pip && venv/bin/pip3 install -r requirements.txt
 
 # Set the group ownership so non-root users can write to /tmp
 RUN chgrp -R 0 /tmp && chmod -R g=u /tmp
 RUN chown -R 1001 /tmp
 
 USER 1001
-
 
 CMD if [ -n "${ROBOT_EXTRA_ARGS}" ]; then \
             echo --extra-robot-args "${ROBOT_EXTRA_ARGS}" &&\
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -17,9 +17,9 @@ RUN curl -L https://github.com/mikefarah/yq/releases/download/v4.16.2/yq_linux_a
     chmod +x /usr/bin/yq
 
 ## Install oc in the container
-RUN curl -L https://mirror.openshift.com/pub/openshift-v$(echo $OC_VERSION | cut -d'.' -f 1)/x86_64/clients/ocp/$OC_CHANNEL-$OC_VERSION/openshift-client-linux.tar.gz -o ./oc_client.tar.gz && \
-    tar xvf oc_client.tar.gz -C /usr/local/bin/ && \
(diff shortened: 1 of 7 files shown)


lugi0 (dismissed):
lgtm

build/Dockerfile
```diff
@@ -1,26 +1,46 @@
...
-    dnf install -y git python3 unzip chromium chromedriver &&\
+    dnf install -y git python3 unzip chromium chromedriver redhat-lsb-core &&\
     dnf clean all
 
+## Install yq in the image
+RUN curl -L https://github.com/mikefarah/yq/releases/download/v4.16.2/yq_linux_amd64 -o /usr/bin/yq &&\
+    chmod +x /usr/bin/yq
+
+## Install oc in the container
+COPY build/oc /usr/local/bin/oc
```
LaVLaS, line 16: This will fail unless you let the user has manually copied the `oc` binary into the `build/` directory.
I would recommend adding a note to the README.

build/Dockerfile
```diff
@@ -1,26 +1,46 @@
...
+# COPY oc /usr/local/bin/oc
+RUN chmod 755 /usr/local/bin/oc &&\
+         oc version --client &&\
```
LaVLaS, line 19: Is this line just to show the client version when building the container?

build/Dockerfile
```diff
@@ -1,26 +1,46 @@
...
 
-CMD ./run_robot_test.sh --skip-pip-install ${RUN_SCRIPT_ARGS}
+
+CMD if [ -n "${ROBOT_EXTRA_ARGS}" ]; then \
+            echo --extra-robot-args "${ROBOT_EXTRA_ARGS}" &&\
+	        ./run_robot_test.sh --skip-pip-install ${RUN_SCRIPT_ARGS} --extra-robot-args "${ROBOT_EXTRA_ARGS}" ;\
+     else \
+            echo execute all &&\
+            ./run_robot_test.sh --skip-pip-install ${RUN_SCRIPT_ARGS};\
+     fi
```
LaVLaS, line 46: Are the `echo` statements required?

Instead of embedding logic in the Dockerfile, you should just document that the user must include `--extra-robot-args` in the env var `ROBOT_EXTRA_ARGS`.  This would simplify this statement to
```
 ./run_robot_test.sh --skip-pip-install ${RUN_SCRIPT_ARGS} "${ROBOT_EXTRA_ARGS}"
```

run_robot_test.sh
```diff
@@ -114,6 +114,34 @@ case "$(uname -s)" in
         ;;
 esac
 
+## if we have yq installed
+if command -v yq &> /dev/null
```
LaVLaS, line 118: Can you make this conditional based on a command line argument instead? This would be preferable so that a developer running the script locally does not current oc context?

Example: If I am logged into a cluster as cluster-admin but the test admin user is a completely different user with a different permissions. Running this would switch my local login and require me to reset it back after each execution

run_robot_test.sh
```diff
@@ -114,6 +114,34 @@ case "$(uname -s)" in
...
+then
+    echo "INFO: we found a yq executable"
+
+    ## get the user, pass and API hostname for OpenShift
+    oc_user=$(yq  e '.OCP_ADMIN_USER.USERNAME' ${TEST_VARIABLES_FILE})
+    oc_pass=$(yq  e '.OCP_ADMIN_USER.PASSWORD' ${TEST_VARIABLES_FILE})
+    oc_host=$(yq  e '.OCP_API_URL' ${TEST_VARIABLES_FILE})
+
+    ## do an oc login here
+    oc login "${oc_host}" --username "${oc_user}" --password "${oc_pass}" --insecure-skip-tls-verify=true
```
LaVLaS, line 128: Is this required for test execution? I assume this is for the robot-openshiftcli use and, if so, shouldn't it perform it's own authentication during instantiation that would be contained to that running instance?

LaVLaS (changes requested):

build/Dockerfile
bdattoma: I can remove the echo statements, they were there for debugging while developing.

In origin, it was like you said, but it wasn't working as reported in a [comment](https://github.com/red-hat-data-services/ods-ci/pull/166#discussion_r782735501) in PR #166, that should be related to quotes escaping issue. In this way we workaround the issue.

run_robot_test.sh
bdattoma: yeah, it's not the best place for oc login. But we need it to make the test running. We'll remove it as soon as the robot-openshiftcli lib will handle the authentication itself

build/Dockerfile
bdattoma: yes, otherwise we could use a variable containing the oc path. Let's see

build/Dockerfile
```diff
@@ -1,26 +1,47 @@
...
+# FROM registry.access.redhat.com/ubi8/ubi:8.1
 # Use this build arg to set any default test script arguments
-ARG RUN_SCRIPT_ARGS=
 ENV RUN_SCRIPT_ARGS=${RUN_SCRIPT_ARGS}
-
+ENV ROBOT_EXTRA_ARGS=${ROBOT_EXTRA_ARGS}
 ENV HOME /tmp
 
+ARG oc_local_path=build/oc
+
```
tarukumar, line 9: Do you think it will be good idea to upload oc file to git repo(personal/ods-ci) and use ADD to directly download and move it to  /usr/local/bin/oc?

bdattoma: not sure. If we do that, we should also maintain it and update it. However, let's wait until we understand if this is something that should be implemented directly in openshiftcli library

run_robot_test.sh
jgarciao: In jenkins we are using `--kubeconfig=/home/jenkins/workspace/rhods-tier2/62/kconfig`  so the different jobs don't interfere each other's session. Could this  approach work here?

bdattoma: done by creating a param `--skip-oclogin` whose default value is `true`. If set to `false`, the script will try to oc login

build/Dockerfile
bdattoma: yes, it could be useful as log of the build, but if you think it's irrelevant we could remove, np

run_robot_test.sh
bdattoma: @jgarciao I'll investigate more on your suggestion to see if it could work, thank you!

build/Dockerfile
bdattoma: now the docker file will download the oc_client directly from `https://mirror.openshift.com/pub/`. The version and channel (e.g., candidate, stable, etc) to be downloaded can be selected by using `OC_VERSION` and `OC_CHANNEL` arguments. Does it work for you @LaVLaS ?

build/Dockerfile
bdattoma: starting from @jgarciao 's suggestion, I've implemented the download of oc  client  directly from `https://mirror.openshift.com/pub/`. The version and channel (e.g., candidate, stable, etc) to be downloaded can be selected by using OC_VERSION and OC_CHANNEL arguments.

LaVLaS (commented):
A few comments but I was able to build and run this with podman
```
 podman run --rm -v $PWD/test-variables.yml:/tmp/ods-ci/test-variables.yml:Z -v $PWD/test-output:/tmp/ods-ci/test-output:Z -e RUN_SCRIPT_ARGS="--skip-oclogin false --test-case tests/Tests/500__jupyterhub/test-jupyterlab-notebook.robot"
```

run_robot_test.sh
```diff
@@ -114,6 +128,61 @@ case "$(uname -s)" in
...
+        if ! ${SKIP_OC_LOGIN}
+            then
+                echo "INFO: OC Login enabled"
+
+                ## get the user, pass and API hostname for OpenShift
+                if ${SET_RHODS_URLS}
+                    then
+                        oc_host=${api_server}
+                    else
+                        oc_host=$(yq  e '.OCP_API_URL' ${TEST_VARIABLES_FILE})
```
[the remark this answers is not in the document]
bdattoma: sure thing

run_robot_test.sh
bdattoma: hm do you mean enable the login by default? I could just change the default value in the `run_robot_test.sh` script rather than setting it in the Docker file. Would it work for you?

run_robot_test.sh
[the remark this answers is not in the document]
bdattoma: done

build/README.md
```diff
@@ -1,7 +1,7 @@
 # ODS-CI Container Image
 
 A [Dockerfile](Dockerfile) is available for running tests in a container.
-The latest build can be downloaded from: https://quay.io/odsci/ods-ci:latest
+The latest build can be downloaded from: https://quay.io/odsci/ods-ci:latest (<em>this image has been deprecated, a new one will be available soon. Please build your own image using the instructions below</em>)
```
LaVLaS, line 4: OPTIONAL: I would delete this entire line if you don't plan to push an image.  That will remove any questions when or if the image is going to be updated

bdattoma: I think we will publish it. But anyway, yeah I can remove the line and when we publish the image we can restore it

LaVLaS (approved):
/lgtm

I was able to build the ods-ci container and run a single testcase.

pablofelix (approved):
lgtm

aloganat (approved):
LGTM

PR #208, opened 2022-02-16
ODS-293 - Verify if users can use custom rhods groups to access JH

bdattoma:
It wants to implement automation for ODS-293: check if users can use custom rhods group to access JupyterHub.
Before merging this PR I need to change the variable file (yml) to add the TEST_USER_x values. I'll do it after receiving some reviews on this, in case anyone disagree with the TC design.

PR #212, opened 2022-02-21
add three additional test users to test-variables for automating ODS-293

bdattoma:
to be merged for allowing #208 to run


lugi0:
Considering we already know the format of the usernames created in the cluster (i.e ldap-user/adminN), is this actually needed?

bdattoma:
> Considering we already know the format of the usernames created in the cluster (i.e ldap-user/adminN), is this actually needed?

I think so, IMO we shouldn't mention anywhere in the code the usernames. Besides, if we change naming logic, this way would be easier to update the code.

lugi0 (approved):
lgtm

jgarciao (approved):
LGTM

PR #213, opened 2022-02-21
[Test/lib]capture and verify if kfdef event is streaming in openshift or not

tarukumar:
This script check if kfdef stream are getting stream in openshift or not 

1. Capture event for redhat-ods-applications
2. Check the number of event captured, Filter the captured data based on resource type
3. Verify if the status of capture event is Normal or not and message is captured as exepected

### DIFF
--- a/tests/Resources/Page/OCPDashboard/Events/Events.resource
+++ b/tests/Resources/Page/OCPDashboard/Events/Events.resource
@@ -18,7 +18,7 @@ Get Event Entry From Openshift
     END
     [Return]    ${res_data}
 
-Get Data Based On Object Kind
+_Get Data Based On Object Kind
     [Documentation]    Compare and filter data
     ...    from the provided payload based
     ...    on object kind and return the data
--- a/tests/Resources/Page/OCPDashboard/Events/Events.resource
+++ b/tests/Resources/Page/OCPDashboard/Events/Events.resource
@@ -8,22 +8,23 @@ Library             OpenShiftCLI
 *** Keywords ***
 Get Event Entry From Openshift
     [Documentation]    Capture openshift event based on namespace
-    ...    and return filtered data
-    [Arguments]    ${namespace}    ${object_kind}    ${expected_event}=1
+    ...    and return filtered data, fail if no events are capture
+    [Arguments]    ${namespace}    ${object_kind}    ${fail-if-no-events}=1
     ${data}    Run Keyword    OpenShiftCLI.Get    kind=Event    namespace=${namespace}
-    IF    len(${data}) >= ${expected_event}
+    IF    len(${data}) >= ${fail-if-no-events}
         ${res_data}    Get Data Based On Object Kind    ${data}    ${object_kind}
     ELSE
         FAIL    All the Events are not captured
     END
     [Return]    ${res_data}
 
-_Get Data Based On Object Kind
+Get Data Based On Object Kind
     [Documentation]    Compare and filter data
     ...    from the provided payload based
     ...    on object kind and return the data
     ...    in new list
     [Arguments]    ${data}    ${resource_kind}
+    [Tags]    Private
     @{tem_data}    Create List
     FOR    ${element}    IN    @{data}
         IF    '${element['involvedObject']['kind']}' == '${resource_kind}'
--- a/tests/Resources/Page/OCPDashboard/Events/Events.resource
+++ b/tests/Resources/Page/OCPDashboard/Events/Events.resource
@@ -6,12 +6,12 @@ Library             OpenShiftCLI
 
 
 *** Keywords ***
-Get Event Entry From Openshift
+Get Openshift Event
     [Documentation]    Capture openshift event based on namespace
     ...    and return filtered data, fail if no events are captured
-    [Arguments]    ${namespace}    ${object_kind}    ${fail-if-no-events}=1
+    [Arguments]    ${namespace}    ${object_kind}    ${expected-events-count}=1
     ${data}    Run Keyword    OpenShiftCLI.Get    kind=Event    namespace=${namespace}
-    IF    len(${data}) >= ${fail-if-no-events}
+    IF    len(${data}) >= ${expected-events-count}
         ${res_data}    Get Data Based On Object Kind    ${data}    ${object_kind}
     ELSE
         FAIL    All the Events are not captured
--- a/tests/Resources/Page/OCPDashboard/Events/Events.resource
+++ b/tests/Resources/Page/OCPDashboard/Events/Events.resource
@@ -9,7 +9,7 @@ Library             OpenShiftCLI
 Get Event Entry From Openshift
     [Documentation]    Capture openshift event based on namespace
     ...    and return filtered data
-    [Arguments]    ${namespace}    ${object_kind}    ${expected_event}=2
+    [Arguments]    ${namespace}    ${object_kind}    ${expected_event}=1
     ${data}    Run Keyword    OpenShiftCLI.Get    kind=Event    namespace=${namespace}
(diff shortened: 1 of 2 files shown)


lugi0 (approved):
lgtm

tests/Resources/Page/OCPDashboard/Events/Events.resource
```diff
@@ -0,0 +1,33 @@
...
+
+Library             Collections
+Library             OpenShiftCLI
+
+
+*** Keywords ***
+Get Event Entry From Openshift
+    [Documentation]    Capture openshift event based on namespace
+    ...    and return filtered data
+    [Arguments]    ${namespace}    ${object_kind}    ${expected_event}=2
```
jgarciao, line 12: ` ${expected_event}=2`

This parameter seems tied to your test and not needed in the library itself.  Why don't you just return an empty list if there are no events and check the list length in the test?

tests/Resources/Page/OCPDashboard/Events/Events.resource
```diff
@@ -0,0 +1,33 @@
...
+    ${data}    Run Keyword    OpenShiftCLI.Get    kind=Event    namespace=${namespace}
+    IF    len(${data}) >= ${expected_event}
+        ${res_data}    Get Data Based On Object Kind    ${data}    ${object_kind}
+    ELSE
+        FAIL    All the Events are not captured
+    END
+    [Return]    ${res_data}
+
+Get Data Based On Object Kind
```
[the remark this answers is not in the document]
tarukumar: I have added underscore, however i don't see any benefit of adding ' _' to a keyword   apart from getting to know if its is private keyword or not, in technical point of view it doesn't really work like private keyword

tests/Resources/Page/OCPDashboard/Events/Events.resource
tarukumar: i disagree, This parameter is not tied to my test it will be useful for other TC in future if we are dealing with events to make sure we get minimum number of event. If i don't get that i think it is good idea to Fail the Keyword.

tests/Tests/100__deploy/130__operators/130__rhods_operator/139__rhods_kfdef_event.robot
```diff
@@ -0,0 +1,36 @@
...
+Resource       ../../../../Resources/Page/OCPDashboard/Events/Events.resource
+
+
+*** Variables ***
+${NAMESPACE}           redhat-ods-applications
+${RESOURCE_KIND}       KfDef
+
+
+*** Test Cases ***
+Verify KFDEF Event Entry
```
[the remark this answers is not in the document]
tarukumar: The actual Tc is to check if kfdef streams are getting streamed or not. I have added extra steps to check the event status. If you like i can change let me know ?

tests/Resources/Page/OCPDashboard/Events/Events.resource
```diff
@@ -0,0 +1,33 @@
...
+    [Arguments]    ${namespace}    ${object_kind}    ${expected_event}=2
+    ${data}    Run Keyword    OpenShiftCLI.Get    kind=Event    namespace=${namespace}
+    IF    len(${data}) >= ${expected_event}
+        ${res_data}    Get Data Based On Object Kind    ${data}    ${object_kind}
+    ELSE
+        FAIL    All the Events are not captured
+    END
+    [Return]    ${res_data}
+
+_Get Data Based On Object Kind
```
lugi0, line 21: Why the `_` before the Keyword name?

tarukumar: There was a proposal from Jorge to have **_** for Private keyword.

lugi0: I don't like it that much but ok :D

lugi0 (approved):
lgtm

tests/Resources/Page/OCPDashboard/Events/Events.resource
tarukumar: Neither do i but i think it will be good practice :D

tests/Tests/100__deploy/130__operators/130__rhods_operator/139__rhods_kfdef_event.robot
```diff
@@ -0,0 +1,36 @@
...
+    [Documentation]    Perform kfdef event streaming
+    ...   in openshift after RHODS deployment
+    [Tags]    ODS-1005    Sanity
```
bdattoma, line 22: could you change it based on the agreed format?
```
[Tags] Sanity
...        ODS-XYZ
```

tests/Resources/Page/OCPDashboard/Events/Events.resource
bdattoma: @tarukumar why setting 2 as threshold?

tests/Tests/100__deploy/130__operators/130__rhods_operator/139__rhods_kfdef_event.robot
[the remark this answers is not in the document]
bdattoma: `Verify KFDEF Is Streaming Events` What do you think @tarukumar @jgarciao ? Would it be a good compromise?

tests/Tests/100__deploy/130__operators/130__rhods_operator/139__rhods_kfdef_event.robot
```diff
@@ -0,0 +1,36 @@
...
+
+
+*** Keywords ***
+Verify KFDEF Event Status
+    [Documentation]    Verify if kfdef event is good and status is Normal
+    [Arguments]     ${kfdef_data}
+    FOR    ${element}    IN    @{kfdef_data}
+        ${name}     Set Variable   ${element['involvedObject']['name']}
+        Should Be Equal As Strings    ${element['message']}
+        ...    KfDef instance ${name} created and deployed successfully
```
bdattoma, line 34: is this string present in all the event entries? or just the first one?

tarukumar: It will be present for all

tests/Resources/Page/OCPDashboard/Events/Events.resource
tarukumar: It is based on the manifest how the mapping is..Each resource is mapped to namespace as i am filtering based on namespace it is expected that we should get those number of events and it should be done. Creating one more keyword doesn't do anything. I think i would keep here only. Regarding setting the default value i wouldn't set to 0 if i set to 0 it means there is not a single event. It is default so if people have less or more they can do it when they call the keyword

tests/Tests/100__deploy/130__operators/130__rhods_operator/139__rhods_kfdef_event.robot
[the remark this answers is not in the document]
tarukumar: I have made changes as berto suggested. resolving the conv

tests/Resources/Page/OCPDashboard/Events/Events.resource
bdattoma: I kind of still disagree, but if you think this is the best way to do that let's keep it, no issues :)

tarukumar: I have changed it to 1

bdattoma (dismissed):

bdattoma (approved):

lugi0 (approved):
LGTM

PR #215, opened 2022-02-22
Fix failures in library version check TC

lugi0:

There was a corner case bug in the way we check that the spawner has loaded; this would cause our TC for version verification to sometimes fail. This PR aims to fix that bug by enforcing a stricter check on the spawner being ready before starting TC execution.


jgarciao (approved):
LGTM

aloganat (approved):
LGTM

PR #217, opened 2022-02-24
Fix for spawner failures in 1.7 automated runs

lugi0:

Potential fix for some of the failures seen in the automated runs for 1.7 (i.e. spawner page not being loaded). Will need to wait for rebuild to happen to test it out.

### DIFF
--- a/tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
+++ b/tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
@@ -40,7 +40,7 @@ Verify PyTorch Image Can Be Spawned With GPU
     ...     ODS-1145
     Clean Up Server
     Stop JupyterLab Notebook Server
-    Handle Start My Server
+    Fix Spawner Status
     Wait Until JupyterHub Spawner Is Ready
     Spawn Notebook With Arguments  image=${NOTEBOOK_IMAGE}  size=Default  gpus=1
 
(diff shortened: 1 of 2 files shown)


tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
```diff
@@ -40,7 +40,7 @@ Verify PyTorch Image Can Be Spawned With GPU
     ...     ODS-1145
     Clean Up Server
     Stop JupyterLab Notebook Server
-    Handle Start My Server
+    Fix Spawner Status
```
bdattoma, line 43: maybe lines 41-41-44 are not needed anymore since `Fix Spawner Status` is used, are they?

lugi0: Theoretically no, but to err on the side of caution I'd prefer to explicitly close the previously running server and then fixing any potential issues with `Fix Spawner Status` since we have seen some failures here

bdattoma: hm I don't like it much, but if you think it is necessary now to make the suite more robust it's okay to keep it.
However, I think we should revisit the Fix Spawner Status and try to make it more reliable to avoid a such duplication of code

lugi0: Yeah, agreed that `Fix Spawner Status` probably needs to be revisited to make it more robust. For now I'd prefer to keep it like this in order to fix during this test cycle.

lugi0:
PR test done, 5 failures but none related to the two test suites this PR changes.

bdattoma (approved):

jgarciao (approved):
lgtm

PR #230, opened 2022-03-04
Add support to install gpu addon and to create machine pool

aloganat:


utils/scripts/ocm/ocm.py
```diff
@@ -777,6 +804,52 @@ def wait_for_osd_cluster_to_get_deleted(self, timeout=3600):
...
+            action="store", dest="pool_instance_type", metavar="",
+            default="g4dn.xlarge")
+        optional_machinepool_cluster_parser.add_argument("--replicas",
+            help="Machine pool replicas",
+            action="store", dest="pool_replicas", metavar="",
+            default="1")
+        optional_machinepool_cluster_parser.add_argument("--taints",
+            help="Machine pool taints information",
+            action="store", dest="taints", metavar="",
+            default="nvidia.com/gpu=NONE:NoSchedule")
```
lugi0, line 846: Are you sure the `NONE` is going to be converted to an empty string for the value of the taint?

utils/scripts/ocm/ocm.py
```diff
@@ -777,6 +804,52 @@ def wait_for_osd_cluster_to_get_deleted(self, timeout=3600):
...
+            required=True)
+
+        optional_machinepool_cluster_parser.add_argument("--instance-type",
+            help="Machine pool instance type",
+            action="store", dest="pool_instance_type", metavar="",
+            default="g4dn.xlarge")
+        optional_machinepool_cluster_parser.add_argument("--replicas",
+            help="Machine pool replicas",
+            action="store", dest="pool_replicas", metavar="",
+            default="1")
```
lugi0, line 842: This is the node count rather than machine pool replicas, right?

aloganat: in API, the argument key is --replicas, so followed the same naming convention, but in UI it's showing as worker node count. Since, worker node count is more meaningful, have done the changes.

utils/scripts/ocm/ocm.py
lugi0: I've tested this manually. The `NONE` is not converted but the value of the taint doesn't matter, it could be empty or not and the taint would still work. I think we can keep it like this.

lugi0 (approved):
lgtm

tarukumar (approved):
LGTM

milindw96 (approved):

PR #233, opened 2022-03-07
Add test for backend of custom notebook images

lugi0:


milindw96 (approved):

tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -0,0 +1,43 @@
...
+
+*** Test Cases ***
+Verify ImageStream Can Be Created
+    [Documentation]    Applies the YAML and Gets the ImageStream
+    [Tags]    Sanity
+    ...       ODS-1208
+    ${apply_status} =    Run Keyword And Return Status    Run    oc apply -f ${YAML}
+    # OpenShiftCLI.Apply    kind=ImageStream    src="configs/resources/custom_image.yaml"
+    # ...    namespace=redhat-ods-applications
+    ${get_status} =    OpenShiftCLI.Get    kind=ImageStream    field_selector=metadata.name==custom-test-image
```
tarukumar, line 25: this one will not return status. Could we change the variable name to get_data or get_payload something like this? Also I think it would be good to have Common keyword for status so that people can  also use in Future.

tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -0,0 +1,43 @@
...
+*** Variables ***
+${YAML} =         configs/resources/custom_image.yaml
+
+
+*** Test Cases ***
+Verify ImageStream Can Be Created
+    [Documentation]    Applies the YAML and Gets the ImageStream
+    [Tags]    Sanity
+    ...       ODS-1208
+    ${apply_status} =    Run Keyword And Return Status    Run    oc apply -f ${YAML}
```
tarukumar, line 22: If possible could we also change the variable name?

tarukumar:
Overall looks good..have added small comment

configs/resources/custom_image.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: image.openshift.io/v1
+kind: ImageStream
+metadata:
+  name: custom-test-image
+  namespace: redhat-ods-applications
+  annotations:
+    opendatahub.io/notebook-image-name: "Custom Image - Testing Only"
```
jgarciao, line 7: This file should be included in the Docker image. I don't know if having it in configs/resources/ is the best place  (@bdattoma  might know better).   Maybe it should be in tests/Tests/500__jupyterhub/config

Other suggestions in case you have more custom images for testing in the future:
- rename file to custom_image_elyra.yaml
- Modify the annotation to "Custom Image - Elyra - Testing Only"

tests/Tests/500__jupyterhub/custom-image.robot
jgarciao: I also  think you  should create a keyword `Apply Configuration` (or similar), probably in tests/Resources/OCP.resource

configs/resources/custom_image.yaml
lugi0: I don't think we will need more YAML files for this specific functionality, so the name should be fine as is. @bdattoma opinions on where to put the file?

tests/Tests/500__jupyterhub/custom-image.robot
lugi0: Sure

tests/Tests/500__jupyterhub/custom-image.robot
lugi0: For applying the yaml we already have a keyword, which is `OpenShiftCLI.Apply` (but it was giving me issues). I don't think we need another keyword to do the same thing to be honest.
Agree with Tarun that the variable can be renamed and we can probably write a keyword to get the status from the data that `OpenShiftCLI.Get` returns.

tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -0,0 +1,46 @@
...
+    ${apply_result} =    Run Keyword And Return Status    Run    oc apply -f ${YAML}
+    Should Be Equal    "${apply_result}"    "True"
+    # OpenShiftCLI.Apply    kind=ImageStream    src="configs/resources/custom_image.yaml"
+    # ...    namespace=redhat-ods-applications
+    ${get_metadata} =    OpenShiftCLI.Get    kind=ImageStream    field_selector=metadata.name==${IMG_NAME}
+    ...    namespace=redhat-ods-applications
+    &{data} =    Set Variable    ${get_metadata}[0]
+    Should Be Equal    ${data.metadata.name}    ${IMG_NAME}
+
+Verify Custom Image Spawn
```
jgarciao, line 32: This tests depends on the execution of the previous one. One solution could be to create it in the Suite Setup and apply the pattern used in other jupyterhub tests

tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -0,0 +1,46 @@
...
+    [Documentation]    Tries spawning the custom image
+    [Tags]    Tier2
+    ...       ODS-1208
```
jgarciao, line 35: Both tests have ODS-1208 as PolarionID.  They should be merged in one test (or have different polarionIDs)

tests/Tests/500__jupyterhub/custom-image.robot
lugi0: Fixed in latest commit

tests/Tests/500__jupyterhub/custom-image.robot
lugi0: Fixed in latest commit

configs/resources/custom_image.yaml
bdattoma: hm good question. We don't have a place where to put this kind of files. I generally like to keep them near where they are used. So if this yaml is a resource of jupyter tests, then I'd store it in the jupyter resources folder. The risk is this is not the most tidy approach. 

Maybe we could create a subdirectory under `tests/Resources` called `Files` and store all those kind of resources there. What do you think?

tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -0,0 +1,42 @@
...
+${IMG_NAME} =     custom-test-image
+
+
+*** Test Cases ***
+Verify Custom Image Can Be Added
+    [Documentation]    Applies the YAML and Gets the ImageStream
+    ...                Then loads the spawner and tries using the custom img
+    [Tags]    Tier2
+    ...       ODS-1208
+    ${apply_result} =    Run Keyword And Return Status    Run    oc apply -f ${YAML}
```
bdattoma, line 24: I would create a keyword `Import Custom Image And Check Status` to wrap lines 24-25

tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -0,0 +1,42 @@
...
+    Should Be Equal    "${apply_result}"    "True"
+    # OpenShiftCLI.Apply    kind=ImageStream    src="configs/resources/custom_image.yaml"
+    # ...    namespace=redhat-ods-applications
+    ${get_metadata} =    OpenShiftCLI.Get    kind=ImageStream    field_selector=metadata.name==${IMG_NAME}
+    ...    namespace=redhat-ods-applications
+    &{data} =    Set Variable    ${get_metadata}[0]
+    Should Be Equal    ${data.metadata.name}    ${IMG_NAME}
```
bdattoma, line 31: I'd write a new keyword `Get Image Stream Metadata` which takes as args the field you want (name), the image stream, namespace and returns the metadata value. 

Or as alternative, a new keyword to wrap lines 28-30-32

lugi0: The problem with parametrizing `the field you want (name)` is that traversing the output is not straightforward and I usually need to check manually how to get to a specific field. I can hardcode the name filter, but it'd be hard to do it for any possible field.

lugi0: I believe @pablofelix has implemented a filter functionality in the latest version of openshiftcli which is able to traverse the output and give you any field you want, but we still haven't updated ods-ci to use. Perhaps it is time to move to the new version? Then this problem would be solved.

Will wrap this in a keyword for now

lugi0: Done in latest commit

tests/Tests/500__jupyterhub/custom-image.robot
lugi0: Done in latest commit

configs/resources/custom_image.yaml
lugi0: Done in latest commit

bdattoma (approved):
lgtm

jgarciao (approved):
lgtm

PR #236, opened 2022-03-08
Add tests to check the image builds alert (ods-717, ods-790)

rkamble5:
Signed-off-by: rkamble5 <rkamble@redhat.com>

### DIFF
--- a/tests/Resources/Page/OCPDashboard/Builds/Builds.robot
+++ b/tests/Resources/Page/OCPDashboard/Builds/Builds.robot
@@ -38,3 +38,42 @@ Check Image Build Status
   ...    Log   The '${build_name}' image match the expected target status
   ...  ELSE
   ...    Fail  The '${build_name}' image build status doesn't match the target status '${target_status}'
+
+Search Last Build
+    [Documentation]    Returns latest(sorted by creation time) build which match the ${build_name_includes}
+    [Arguments]    ${namespace}    ${build_name_includes}
+    ${build} =    Run    oc get builds --sort-by=.metadata.creationTimestamp -n ${namespace} | grep ${build_name_includes} | awk '{print $1;}'
+    @{builds} =  Split String  ${build}  \n
+    [Return]    ${builds}[-1]
+
+Delete Build
+    [Documentation]    Deletes the build ${build_name} in ${namespace}
+    [Arguments]    ${namespace}    ${build_name}
+    OpenShiftCLI.Delete    kind=Build    namespace=${namespace}    field_selector=metadata.name==${build_name}
+
+Start New Build
+    [Documentation]    Starts new build using ${buildconfig}
+    [Arguments]    ${namespace}    ${buildconfig}
+    ${name} =    Run    oc start-build ${buildconfig} -n ${namespace}
+    @{list_for} =    Split String    ${name}    ${SPACE}
+    @{name_list} =    Split String    ${list_for}[0]    /
+    [Return]    ${name_list}[1]
+
+Get Build Status From Oc
+    [Documentation]    Get Status of build using name
+    [Arguments]    ${namespace}    ${build_name}
+    ${status} =    Run    oc get builds -n ${namespace} -o json | jq '.items[] | select(.metadata.name == "${build_name}") | .status.phase'
+    ${status}    Replace String    ${status}    "    ${EMPTY}
+    [Return]    ${status}
+
+Build Status Should Be
+    [Documentation]    Get status and check with ${expected_status}
+    [Arguments]    ${namespace}    ${build_name}    ${expected_status}=Complete
+    ${status} =    Get Build Status From Oc    namespace=${namespace}    build_name=${build_name}
+    Should Be Equal    ${status}    ${expected_status}
+
+Wait Until Build Status Is
+    [Documentation]    Check status build with ${expected_status} for every min until is succeed or timeout
+    [Arguments]    ${namespace}    ${build_name}    ${expected_status}=Complete    ${timeout}=20 min
+    Wait Until Keyword Succeeds    ${timeout}    1 min
+    ...    Build Status Should Be    ${namespace}    ${build_name}    ${expected_status}
--- a/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
+++ b/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
@@ -132,8 +132,37 @@ Verify Container Image
     Should Be Equal  ${image}  ${expected-image-url}
 
 Search Pod
-    [Documentation]   Return list of pods   pod_start_with = here ypu have to provide starting of pod name
+    [Documentation]   Returns list pod  ${pod_start_with} = here ypu have to provide starting of pod name
     [Arguments]   ${namespace}  ${pod_start_with}
     ${pod} =  Run  oc get pods -n ${namespace} -o json | jq '.items[] | select(.metadata.name | startswith("${pod_start_with}")) | .metadata.name'
     @{list_pods} =  Split String  ${pod}  \n
+    [Return]  ${list_pods}
+
+Run Command In Container
+    [Documentation]    Executes a command in a container.
+    ...    If ${container_name} is omitted, the first container in the pod will be chosen
+    [Arguments]    ${namespace}    ${pod_name}    ${command}    ${container_name}=${EMPTY}
+    IF    "${container_name}" == "${EMPTY}"
+        ${output}    Run    oc exec ${pod_name} -n ${namespace} -- ${command}
+    ELSE
+        ${output}    Run    oc exec ${pod_name} -n ${namespace} -c ${container_name} -- ${command}
+    END
+    [Return]    ${output}
+
+Wait Until Container Exist
+    [Documentation]     Waits until container is exists
+    [Arguments]     ${namespace}    ${pod_name}    ${container_to_check}    ${timeout}=5 min
+    Wait Until Keyword Succeeds    ${timeout}    1 min
+    ...    Check Is Container Exist    namespace=${namespace}  pod_name=${pod_name}  container_to_check=${container_to_check}
+
+Check Is Container Exist
+    [Documentation]     Checks container is exist or if not then fails
+    [Arguments]     ${namespace}    ${pod_name}    ${container_to_check}
+    ${container_name} =  Run  oc get pod ${pod_name} -n ${namespace} -o json | jq '.spec.containers[] | select(.name == "${container_to_check}") | .name'
+    Should Be Equal    "${container_to_check}"    ${container_name}
+
+Find First Pod By Name
+    [Documentation]   Returns first occurred pod  ${pod_start_with} = here ypu have to provide starting of pod name
+    [Arguments]   ${namespace}  ${pod_start_with}
+    ${list_pods} =  Search Pod  namespace=${namespace}  pod_start_with=${pod_start_with}
     [Return]  ${list_pods}[0]
--- a/tests/Tests/100__deploy/100__installation/102__post_install.robot
+++ b/tests/Tests/100__deploy/100__installation/102__post_install.robot
@@ -6,7 +6,7 @@ Library             OperatingSystem
 Resource            ../../../Resources/OCP.resource
 Resource            ../../../Resources/Page/OCPDashboard/OCPDashboard.resource
 Resource            ../../../Resources/Page/ODH/JupyterHub/HighAvailability.robot
-
+Resource            ../../../Resources/Page/ODH/Prometheus/Prometheus.robot
 
 *** Test Cases ***
 Verify Dashboard Deployment
@@ -71,7 +71,7 @@ Verify That Prometheus Image Is A CPaaS Built Image
     [Tags]    Sanity   
     ...     Tier1
     ...     ODS-734    
-    ${pod} =    Search Pod    namespace=redhat-ods-monitoring    pod_start_with=prometheus-
+    ${pod} =    Find First Pod By Name    namespace=redhat-ods-monitoring    pod_start_with=prometheus-
     Verify Container Image    redhat-ods-monitoring    ${pod}    prometheus
     ...    "registry.redhat.io/openshift4/ose-prometheus"
     Verify Container Image    redhat-ods-monitoring    ${pod}    oauth-proxy
@@ -81,7 +81,7 @@ Verify That Grafana Image Is A Red Hat Built Image
     [Tags]    Sanity    
     ...     Tier1
     ...     ODS-736    
-    ${pod} =    Search Pod    namespace=redhat-ods-monitoring    pod_start_with=grafana-
+    ${pod} =    Find First Pod By Name    namespace=redhat-ods-monitoring    pod_start_with=grafana-
     Verify Container Image    redhat-ods-monitoring    ${pod}    grafana
     ...    "registry.redhat.io/rhel8/grafana:7"
     Verify Container Image    redhat-ods-monitoring    ${pod}    auth-proxy
@@ -91,7 +91,7 @@ Verify That Blackbox-exporter Image Is A CPaaS Built Image
     [Tags]    Sanity    
     ...     Tier1
     ...     ODS-735    
-    ${pod} =    Search Pod    namespace=redhat-ods-monitoring    pod_start_with=blackbox-exporter-
+    ${pod} =    Find First Pod By Name    namespace=redhat-ods-monitoring    pod_start_with=blackbox-exporter-
     Verify Container Image    redhat-ods-monitoring    ${pod}    blackbox-exporter
     ...    "quay.io/integreatly/prometheus-blackbox-exporter:v0.19.0"
 
@@ -99,6 +99,6 @@ Verify That Alert Manager Image Is A CPaaS Built Image
     [Tags]    Sanity    
     ...     Tier1
     ...     ODS-733    
-    ${pod} =    Search Pod    namespace=redhat-ods-monitoring    pod_start_with=prometheus-
+    ${pod} =    Find First Pod By Name    namespace=redhat-ods-monitoring    pod_start_with=prometheus-
     Verify Container Image    redhat-ods-monitoring    ${pod}    alertmanager
     ...    "registry.redhat.io/openshift4/ose-prometheus-alertmanager"
--- a/tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
+++ b/tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
@@ -3,8 +3,10 @@ Documentation       RHODS monitoring alerts test suite
 
 Resource            ../../../Resources/ODS.robot
 Resource            ../../../Resources/Common.robot
+Library             OperatingSystem
 Library             SeleniumLibrary
 Library             JupyterLibrary
+Library             OpenShiftCLI
 
 Suite Setup         Alerts Suite Setup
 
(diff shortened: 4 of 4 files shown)


tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,33 @@ Check Image Build Status
...
+Get Last Build
+    [Documentation]    Returns last build
+    [Arguments]    ${namespace}    ${word}
+    ${build} =    Run    oc get builds -n ${namespace} | grep ${word} | awk '{print $1;}'
+    [Return]    ${build}
+
+Delete Build Using Name
+    [Documentation]    Delete the build
+    [Arguments]    ${namespace}    ${build_name}
+    Run    oc delete builds ${build_name} -n ${namespace}
```
jgarciao, line 51: Can be done with the OpenShiftCLI library:
`OpenShiftCLI.Delete  kind=Build  namespace=${namespace}    field_selector=metadata.name==${build_name}`

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,33 @@ Check Image Build Status
...
+    Sleep    120s
```
jgarciao, line 52: I think this two lines are not needed. We expect the previous deletion  command  to work properly or fail if not

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,33 @@ Check Image Build Status
...
+
+Start New Build
+    [Documentation]    Start new Buid
+    [Arguments]    ${namespace}    ${buildconfig}
+    ${name} =    Run    oc start-build ${buildconfig} -n ${namespace}
+    @{list_for} =    Split String    ${name}    ${SPACE}
+    @{name_list} =    Split String    ${list_for}[0]    /
+    [Return]    ${name_list}[1]
+
+Get Build Status By Name
```
jgarciao, line 64: I think `Get Build Status` is enough

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,33 @@ Check Image Build Status
...
+    [Documentation]    Get Status of build using name
+    [Arguments]    ${namespace}    ${name}
```
jgarciao, line 66: ${build-name} is better

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
```diff
@@ -204,6 +205,41 @@ Verify Alert "RHODS Probe Success Burn Rate" Is Fired When RHODS Dashboard Is Do
...
+    ...    ${RHODS_PROMETHEUS_TOKEN}
+    ...    Builds
+    ...    JupyterHub image builds are failing
+    [Teardown]    Delete Build Using Name    namespace=redhat-ods-applications    build_name=${build_name_fail}
+
+JupyterHub Image Builds Are Failing And Fires Alert For 30 min
+    [Tags]    Sanity    
+    ...     Tier2    
+    ...     ODS-790
+    ${build_name_fail} =    Bring JupyterHub Image Builds Are In Failing State
```
jgarciao, line 229: ${failed-build-name} maybe it's better

jgarciao (commented):
Can you please rename this PR to "Add tests to check the image builds alert (ods-717, ods-790)"?

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
rkamble5: `Get Build Status` is already present in the build file
so that's why I used `Get Build Status By Name`

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,33 @@ Check Image Build Status
...
   ...    Log   The '${build_name}' image match the expected target status
   ...  ELSE
-  ...    Fail  The '${build_name}' image build status doesn't match the target status '${target_status}'
\ No newline at end of file
+  ...    Fail  The '${build_name}' image build status doesn't match the target status '${target_status}'
+
+Get Last Build
+    [Documentation]    Returns last build
+    [Arguments]    ${namespace}    ${word}
+    ${build} =    Run    oc get builds -n ${namespace} | grep ${word} | awk '{print $1;}'
```
lugi0, line 45: What is `${word}` supposed to be? can you give it a more meaningful name?

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
```diff
@@ -204,6 +205,41 @@ Verify Alert "RHODS Probe Success Burn Rate" Is Fired When RHODS Dashboard Is Do
...
+    ...    JupyterHub image builds are failing
+
+    Prometheus.Wait Until Alert Is Not Firing    ${RHODS_PROMETHEUS_URL}
+    ...    ${RHODS_PROMETHEUS_TOKEN}
+    ...    Builds
+    ...    JupyterHub image builds are failing
+    ...    32 min
+    ${build_name} =    Start New Build    namespace=redhat-ods-applications
+    ...    buildconfig=s2i-tensorflow-gpu-cuda-11.4.2-notebook
+    [Teardown]    Delete Build Using Name    namespace=redhat-ods-applications    build_name=${build_name_fail}
```
lugi0, line 242: Wouldn't this leave the cluster without a tensorflow build?

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
```diff
@@ -204,6 +205,41 @@ Verify Alert "RHODS Probe Success Burn Rate" Is Fired When RHODS Dashboard Is Do
...
+    ...    ${RHODS_PROMETHEUS_TOKEN}
+    ...    Builds
+    ...    JupyterHub image builds are failing
+    ${build_name} =    Start New Build    namespace=redhat-ods-applications
+    ...    buildconfig=s2i-tensorflow-gpu-cuda-11.4.2-notebook
+    Prometheus.Wait Until Alert Is Not Firing    ${RHODS_PROMETHEUS_URL}
+    ...    ${RHODS_PROMETHEUS_TOKEN}
+    ...    Builds
+    ...    JupyterHub image builds are failing
+    [Teardown]    Delete Build Using Name    namespace=redhat-ods-applications    build_name=${build_name_fail}
```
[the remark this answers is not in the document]
rkamble5: we are deleting failed builds and on the cluster, we are placing good one build

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
rkamble5: we are deleting failed builds and on the cluster, we are placing good one build

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,42 @@ Check Image Build Status
   Run Keyword If    "${build_status}" == "${target_status}"
   ...    Log   The '${build_name}' image match the expected target status
   ...  ELSE
-  ...    Fail  The '${build_name}' image build status doesn't match the target status '${target_status}'
\ No newline at end of file
+  ...    Fail  The '${build_name}' image build status doesn't match the target status '${target_status}'
+
+Search Last Build
+    [Documentation]    Returns last build
+    [Arguments]    ${namespace}    ${word-to-search-build}
```
jgarciao, line 44: What about ${build-name-includes}

jgarciao: Please, can you write in the documentation what happens if there are two builds matching ${build-name-includes}? Is it really giving you the latest one?

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,42 @@ Check Image Build Status
...
+    ${build} =    Run    oc get builds -n ${namespace} | grep ${word-to-search-build} | awk '{print $1;}'
+    [Return]    ${build}
+
+Delete Build
+    [Documentation]    Delete the build
```
jgarciao, line 49: The style guide says it should be in the 3rd form (Deletes instead of Delete): 
_Deletes the build ${build_name} in  ${namespace}_

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,42 @@ Check Image Build Status
...
+    [Arguments]    ${namespace}    ${build_name}
+    OpenShiftCLI.Delete    kind=Build    namespace=${namespace}    field_selector=metadata.name==${build_name}
+
+Start New Build
+    [Documentation]    Start new Buid
```
jgarciao, line 54: Same as before (Starts instead of Start)
_Starts new build using ${buildconfig}_

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,42 @@ Check Image Build Status
...
+    [Return]    ${name_list}[1]
+
+Get Build Status By Name
+    [Documentation]    Get Status of build using name
+    [Arguments]    ${namespace}    ${build-name}
+    ${build_info} =    Run    oc get builds -n ${namespace} | grep ${build-name}
+    @{status} =    Split String    ${build_info}    ${SPACE}
+    [Return]    ${status}[9]
+
+Check Build Status Is
```
jgarciao, line 68: `Build Status Should Be`

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
bdattoma: as alternative, there is `Search Last Item Instance By Title in OpenShift Table` keywords which search in an openshift table and sort the table by creation date (DESC order)

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
bdattoma: is there a reason to create a new keyword to get status if there's already one available?

rkamble5: Yes
The `get status` keyword uses a web browser.
But for our test case, we are using OpenshiftCLI or oc commands that's why we have to create another keyword that work with openshift or oc..

bdattoma: hm okay, then I'd change the name of the kw. I think the best approach would be to rename the already existent kw to `Get Build Status From UI` but in that case you should change accordingly all the TCs using that keyword (I think it should be only one).
As alternative, you could call you keyword `Get Build Status From Oc`. @jgarciao @rkamble5 what do you prefer?

rkamble5: `Get Build Status From Oc` I think this one is better/good

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
rkamble5: Done

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
rkamble5: Done,
Now it returns the latest build,I have sorted using the creation time of build.

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
```diff
@@ -284,6 +325,18 @@ Verify "RHODS Probe Success Burn Rate" Alerts Are Not Firing And Continue On Fai
...
     ...    ${pm_url}    ${pm_token}    ${rule_group}    ${alert}    ${alert-duration}
     Skip If    ${alert_is_firing}    msg=Test skiped because alert "${alert} ${alert-duration}" is already firing
+
+Provoke Tensorflow Image Build Failure
+    [Documentation]   Starts New Build after some time it kill the build and return  name of failed build
+    ${build} =    Search Last Build    namespace=redhat-ods-applications    build-name-includes=tensorflow
+    Deletes Build    namespace=redhat-ods-applications    build_name=${build}
+    ${failed_build_name} =    Starts New Build    namespace=redhat-ods-applications
+    ...    buildconfig=s2i-tensorflow-gpu-cuda-11.4.2-notebook
+    Sleep    60s
```
bdattoma, line 338: is there a reason for sleeping 1 minute?

rkamble5: Because when Start new build, first it takes place in `pending` state and then move to `running`
and also in the next step, we have to kill the pod so we are waiting for 1 min to pod place in running

jgarciao: It's better that you use here _Wait Until Build Status Is_  (with a timeout of 15 or 20 mins), as builds can be in Pending state for a long time if cluster resources are scarce

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,43 @@ Check Image Build Status
...
+  ...    Fail  The '${build_name}' image build status doesn't match the target status '${target_status}'
+
+Search Last Build
+    [Documentation]    Returns latest(sorted by creation time) build which match the ${build-name-includes}
+    [Arguments]    ${namespace}    ${build-name-includes}
+    ${build} =    Run    oc get builds --sort-by=.metadata.creationTimestamp -n ${namespace} | grep ${build-name-includes} | awk '{print $1;}'
+    @{builds} =  Split String  ${build}  \n
+    [Return]    ${builds}[-1]
+
+Deletes Build
```
jgarciao, line 49: The keyword name should be `Delete`. In the documentation section `Deletes` is correct

tests/Resources/Page/OCPDashboard/Builds/Builds.robot
```diff
@@ -37,4 +37,43 @@ Check Image Build Status
...
+    [Documentation]    Deletes the build ${build_name} in ${namespace}
+    [Arguments]    ${namespace}    ${build_name}
+    OpenShiftCLI.Delete    kind=Build    namespace=${namespace}    field_selector=metadata.name==${build_name}
+
+Starts New Build
```
jgarciao, line 54: The keyword name should be `Start New Build` (Start without s). In the documentation section Starts (with s) is correct

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
rkamble5: Ok
so I Have to pass `Running` state  ${expected-status} to `wait Until build state is`

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
```diff
@@ -132,8 +132,15 @@ Verify Container Image
...
     [Arguments]   ${namespace}  ${pod_start_with}
     ${pod} =  Run  oc get pods -n ${namespace} -o json | jq '.items[] | select(.metadata.name | startswith("${pod_start_with}")) | .metadata.name'
     @{list_pods} =  Split String  ${pod}  \n
     [Return]  ${list_pods}[0]
+
+Kill Pod
+    [Documentation]    Search and kills pod with name including ${pod-name-include} and state = Running
+    [Arguments]    ${namespace}    ${pod-name-includes}
+    ${pod}=    Run    oc get pods -n ${namespace} | grep ${pod-name-includes} | grep Running | awk '{print $1;}'
+    Run    oc exec ${pod} -n ${namespace} -- /bin/kill 1
```
lugi0, line 145: This might return multiple pods (e.g. with namespace=redhat-ods-applications and pod-name-includes=jupyter). What happens in that case? Are you killing all pods? only one pod? Which one in that case?

Also, why are you not using `oc delete pod` instead of executing kill PID1 in the container? Isn't that killing only 1 container of the pod? What if there are multiple running containers in the pod?

rkamble5: No it no returns multiple pods because we are passing full name of build (eg namespace=redhat-ods-applications and pod-name-includes=s2i-tensorflow-gpu-cuda-11.4.2-notebook-5) and pod name = build name + "-build" so evry time it will return just one name.
For `oc delete pod` I Have to talk with jorge about it.

lugi0: You might be passing a fully qualified name to the keyword, but this is sitting in `Pods.robot` and will surely be used by other tests sooner or later. I think you need to also understand what happens when you are returned multiple pods and how to handle that situation.

rkamble5: Ok I got it
So here we can, sort the pods using creation time,fetch latest one and it will delete the latest pod..

jgarciao: @rkamble5 is going to add the parameter ${container_name} to this keyword to choose the container to run the command in

We are killing the pod to provoke the build to be in Failed state. I believe using `oc delete pod` the resulting state was different (Cancelled?, I don't remember well)

lugi0: >sort the pods using creation time,fetch latest one and it will delete the latest pod

IMHO that's not a good strategy for a general keyword. If you go ahead with that you should probably rename the keyword to something like `Kill Youngest Pod` and clarify what you are doing in the documentation of the keyword.

>We are killing the pod to provoke the build to be in Failed state. I believe using oc delete pod the resulting state was different (Cancelled?, I don't remember well)

That makes sense then. However then the keyword name should probably not be `Kill Pod` but something like `Kill Container In Pod`, or `Make Pod Go In Failed State`

rkamble5:
Tested successfully here - https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/466/
But there are known issues for these test cases 
issue:- https://issues.redhat.com/browse/RHODS-2051

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
```diff
@@ -132,8 +132,21 @@ Verify Container Image
     Should Be Equal  ${image}  ${expected-image-url}
 
 Search Pod
-    [Documentation]   Return list of pods   pod_start_with = here ypu have to provide starting of pod name
+    [Documentation]   Returns list of pods   pod_start_with = here ypu have to provide starting of pod name
     [Arguments]   ${namespace}  ${pod_start_with}
     ${pod} =  Run  oc get pods -n ${namespace} -o json | jq '.items[] | select(.metadata.name | startswith("${pod_start_with}")) | .metadata.name'
     @{list_pods} =  Split String  ${pod}  \n
     [Return]  ${list_pods}[0]
```
lugi0, line 139: Is this actually returning a list of pods? Or is it only the first element of a list?

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
```diff
@@ -132,8 +132,21 @@ Verify Container Image
...
+
+Run Command In Container
+    [Documentation]    Executes a command in a container.
+    ...    If ${container_name} is omitted, the first container in the pod will be chosen
+    [Arguments]    ${namespace}    ${pod_name_start_with}    ${command}    ${container_name}=${EMPTY}
+    ${pod_name}    Search Pod    namespace=${namespace}    pod_start_with=${pod_name_start_with}
```
lugi0, line 145: I still don't understand if this works on a single pod or a list of pods. And if it works on a list of pods, on what container/pod is it actually running the command?

rkamble5: It works on a single pod(Which is provided by `Search Pod` )
if we provide the name of the container then it will exec the command on that container.
if it is not provided then it will exec command on first container or we can say on default container
I will share example with you in googlechat

lugi0 (dismissed):
I still don't like the fact that we are returning the first pod out of a list of multiple pods without any more control than that. If @jgarciao is fine with it we can merge this.

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
jgarciao: I think `Search Pod` returns just the first pod matching the name. Rushikesh has updated the [Documentation] tag to reflect that (he didn't change the behaviour)

rkamble5: Yes @jgarciao 
I have changed documentation,not behavior

lugi0: @jgarciao as I wrote in the other comment, I don't like the fact that `Search Pod` arbitrarily returns the first pod out of a list of an unknown number of pods; I feel like this is not a good functionality, and that sooner or later someone is going to try to reuse that keyword and it won't work. When that happens, we will either have to replace the keyword or update the internal logic, which might lead to this TC then breaking apart. 
IMHO we should write the keyword in such a way that `Search Pod` returns *exactly* the pod you want, not just the first one according to `oc`'s ordering of the list, but if you feel that this is ok I've already given the approval to the PR and we can proceed with the merge.

bdattoma:
> I still don't like the fact that we are returning the first pod out of a list of multiple pods without any more control than that. If @jgarciao is fine with it we can merge this.

maybe, a better approach would be to return the entire list of pod because there isn't a criterion to chose one. If you pass the full name of the pod, you will receive a list of one element. Would it work?

lugi0:
That would certainly be better

tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -72,33 +72,33 @@ Verify That Prometheus Image Is A CPaaS Built Image
     ...     Tier1
     ...     ODS-734    
     ${pod} =    Search Pod    namespace=redhat-ods-monitoring    pod_start_with=prometheus-
```
jgarciao, line 74: As pod is a list maybe it's better to call it ${pods}    (in all tests)
`${pods} =    Search Pod    namespace=redhat-ods-monitoring    pod_start_with=prometheus-`

Alternatively, you could create a new keyword  `Find First Pod By Name`

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
```diff
@@ -204,6 +206,76 @@ Verify Alert "RHODS Probe Success Burn Rate" Is Fired When RHODS Dashboard Is Do
...
+    ...    JupyterHub image builds are failing
+    [Teardown]    Delete Build    namespace=redhat-ods-applications    build_name=${failed_build_name}
+
+Verify Alert "JupyterHub Image Builds Are Failing" Fires Up To 30 Minutes When There Is An Image Build Error
+    [Documentation]     Verify that alert is firing up to 30 min and after thar resolve automatically
+    [Tags]    Sanity
+    ...       Tier2
+    ...       ODS-790
+    ...       KnownIssues
+    ${failed_build_name} =    Provoke Tensorflow Image Build Failure
```
jgarciao, line 254: @gmfrasca said that a way not to have the two build alert tests interfere each other could be to use tensorflow build in one test and pytorch build in the other

Could you add the build name as parameter and use different builds on each test?

```
Provoke Image Build Failure
[Arguments]    ${build_name_includes}
```

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
```diff
@@ -284,6 +356,29 @@ Verify "RHODS Probe Success Burn Rate" Alerts Are Not Firing And Continue On Fai
...
+Provoke Tensorflow Image Build Failure
+    [Documentation]    Starts New Build after some time it kill the build and return name of failed build
+    ${build} =    Search Last Build    namespace=redhat-ods-applications    build_name_includes=tensorflow
+    Delete Build    namespace=redhat-ods-applications    build_name=${build}
+    ${failed_build_name} =    Start New Build    namespace=redhat-ods-applications
+    ...    buildconfig=s2i-tensorflow-gpu-cuda-11.4.2-notebook
+    ${pod_name} =   Search Pod  namespace=redhat-ods-applications  pod_start_with=${failed_build_name}
+    Wait Until Build Status Is    namespace=redhat-ods-applications    build_name=${failed_build_name}
+    ...    expected_status=Running
+    Run Command In Container    namespace=redhat-ods-applications    pod_name=${pod_name}[0]
```
jgarciao, line 372: This could be a keyword

```
Kill Container In Pod
[Documentation]  Kills abruptly a running container in a pod executing `bin/kill 1` inside it 
[Arguments]  ${namespace}  ${pod}  ${container}
```

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
rkamble5: Done

tests/Tests/200__monitor_and_manage/210__alerts/210__alerts.robot
rkamble5: Done

jgarciao (approved):
lgtm

tests/Resources/Page/OCPDashboard/Pods/Pods.robot
```diff
@@ -132,7 +132,30 @@ Verify Container Image
...
+        ${output}    Run    oc exec ${pod_name} -n ${namespace} -c ${container_name} -- ${command}
+    END
+    [Return]    ${output}
+
+Kill Container In Pod
+    [Documentation]  Kills abruptly a running container in a pod executing `bin/kill 1` inside it 
+    [Arguments]  ${namespace}  ${pod_name}  ${container_name}
+    Run    oc exec ${pod_name} -n ${namespace} -c ${container_name} -- /bin/kill 1
+
+Find First Pod By Name
```
bdattoma, line 157: all the lines of this keyword are implemented by `Search Pod` keyword. I think you can reuse it and return only the list cell you need to return

lugi0 (dismissed):
LGTM other than berto's comment

lugi0 (dismissed):
lgtm

jgarciao (approved):
lgtm

jgarciao (approved):
lgtm

bdattoma (approved):
lgtm

PR #235, opened 2022-03-08
Fix keyword conflict for post-install

lugi0:
Signed-off-by: Luca Giorgi <lgiorgi@redhat.com>

PR #240, opened 2022-03-09
Verify versioning each release

prudhvi-165:


tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -68,37 +69,45 @@ Verify GPU Operator Deployment  # robocop: disable
...
     Verify Container Image    redhat-ods-monitoring    ${pod}    alertmanager
     ...    "registry.redhat.io/openshift4/ose-prometheus-alertmanager"
+
+Verify Versioning Each Release
+    [Documentation]    Verify versioning each release
+    [Tags]    Sanity
+    ...       ODS-478
+    @{list} =    Oc Get    kind=ClusterServiceVersion    label_selector=olm.copiedFrom=redhat-ods-operator
+    &{dict} =    Set Variable    ${list}[0]
+    Should Match Regexp    ${dict.spec.version}    ^[0-9]+\.[0-9]+\.[0-9]+\-[0-9]+$
```
lugi0, line 113: Mh, if I'm correct you're checking for a match for `x.y.z-build`, but I'm not sure the `-build` will always be present. Perhaps that last bit should be optional? Like `^[0-9]+\.[0-9]+\.[0-9]+\-*[0-9]*$`

tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -68,37 +69,45 @@ Verify GPU Operator Deployment  # robocop: disable
...
+    ...     ODS-733
     ${pod} =    Search Pod    namespace=redhat-ods-monitoring    pod_start_with=prometheus-
     Verify Container Image    redhat-ods-monitoring    ${pod}    alertmanager
     ...    "registry.redhat.io/openshift4/ose-prometheus-alertmanager"
+
+Verify Versioning Each Release
+    [Documentation]    Verify versioning each release
+    [Tags]    Sanity
+    ...       ODS-478
+    @{list} =    Oc Get    kind=ClusterServiceVersion    label_selector=olm.copiedFrom=redhat-ods-operator
```
bdattoma, line 111: wasn't possible to use `Get RHODS version` which is already implemented?

tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -68,37 +69,45 @@ Verify GPU Operator Deployment  # robocop: disable
...
+    [Tags]    Sanity
     ...     Tier1
-    ...     ODS-733    
+    ...     ODS-733
     ${pod} =    Search Pod    namespace=redhat-ods-monitoring    pod_start_with=prometheus-
     Verify Container Image    redhat-ods-monitoring    ${pod}    alertmanager
     ...    "registry.redhat.io/openshift4/ose-prometheus-alertmanager"
+
+Verify Versioning Each Release
+    [Documentation]    Verify versioning each release
```
bdattoma, line 108: could the doc be more specific? what "verify versioning" means?

tests/Tests/100__deploy/100__installation/102__post_install.robot
bdattoma: just checked and that keyword fetch only the x.y.z. However, i feel like we should have a common keyword to get the rhods version. You may change the already implemented one to retrieve also the full version format (x.y.z-build). Maybe by using an argument to pilot the keyword to return the full or partial version.

lugi0: I think the best approach would be to have `Get RHODS Version` return the full version, but then you'd also need to modify the keyword used for the version comparison to modify the returned value and drop the `-build`

tests/Tests/100__deploy/100__installation/102__post_install.robot
prudhvi-165: No. the description says we need to check the complete version i.e of the form a.b.c-d

lugi0: The description of what? we are the ones defining the test cases :)
My worry is that this will fail in the case in which there have been no rebuilds of a release, and the version would be x.y.z-0; in that case I *think* that the -0 would be dropped (as it is not communicating anything), but we have never encountered this situation.
In any case, `^[0-9]+\.[0-9]+\.[0-9]+\-*[0-9]*$` would match both `x.y.z-w` and `x.y.z`

tests/Tests/100__deploy/100__installation/102__post_install.robot
prudhvi-165: why drop -build , because in the description it is written we need to check complete version. https://issues.redhat.com/browse/RHODS-1695

tests/Tests/100__deploy/100__installation/102__post_install.robot
prudhvi-165: https://issues.redhat.com/browse/RHODS-1695

tests/Tests/100__deploy/100__installation/102__post_install.robot
prudhvi-165: It means Checking Build Version of Rhods, will update it.

tests/Tests/100__deploy/100__installation/102__post_install.robot
lugi0: What I mean is that if you modify `Get RHODS version` (https://github.com/red-hat-data-services/ods-ci/blob/master/tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot#L93) to return the full version, then you'd also need to modify `Is RHODS Version Greater Or Equal Than` (https://github.com/red-hat-data-services/ods-ci/blob/master/tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot#L102) to remove the `-build` in there.

So you'd have that `Get RHODS version` becomes:
```
Get RHODS version
    #@{list} =  OpenShiftCLI.Get  kind=ClusterServiceVersion  label_selector=olm.copiedFrom=redhat-ods-operator
    #&{dict} =  Set Variable  ${list}[0]
    #Log  ${dict.spec.version}
    ${ver} =  Run  oc get csv -n redhat-ods-operator | grep "rhods-operator" | awk '{print $1}' | sed 's/rhods-operator.//'
    [Return]  ${ver}
```

And `Is RHODS Version Greater Or Equal Than` becomes:
```
Is RHODS Version Greater Or Equal Than
    [Arguments]  ${target}
    ${ver} =  Get RHODS version
    ${ver} =  Fetch From Left  ${ver}  -
    ${comparison} =  GTE  ${ver}  ${target}
    [Return]  ${comparison}
```

tests/Tests/100__deploy/100__installation/102__post_install.robot
lugi0: Yes, the story mentions that the build number increases `for every new build that arrives with bug fixes. `. There might be a release where we don't ever need to rebuild to include bug fixes, and that would mean that its version would be `x.y.z-0`; I expect that in that case `-0` would not be included in the actual build number, as it does not convey any information. With the regex you have currently defined, this would fail the TC.

tests/Tests/100__deploy/100__installation/102__post_install.robot
bdattoma: I agree with @lugi0

tests/Tests/100__deploy/100__installation/102__post_install.robot
prudhvi-165: Updated.

tests/Tests/100__deploy/100__installation/102__post_install.robot
prudhvi-165: Updated.

tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -68,37 +68,44 @@ Verify GPU Operator Deployment  # robocop: disable
     # ...   nvidia-operator-validator DS
 
 Verify That Prometheus Image Is A CPaaS Built Image
-    [Tags]    Sanity   
+    [Tags]    Sanity
     ...     Tier1
```
lugi0, line 72: When did we start using `Tier1` as a tag for TCs?

lugi0 (dismissed):
lgtm in general, please test it by running the full sanity suite

prudhvi-165:
Jenkins Run : https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/463/ 
Result : 4 Tests failing , for which a PR has been raised [here](https://github.com/red-hat-data-services/ods-ci/pull/235/files)

jgarciao:
Please, can you run the whole sanity suite? Also, can you sign the PR with your @redhat.com email account?

Info here: https://github.com/red-hat-data-services/ods-ci/blob/master/CONTRIBUTING.md

prudhvi-165:
Jenkins Link for full suite : https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/482/

tests/Tests/100__deploy/100__installation/102__post_install.robot
jgarciao: You can rename the test to `Verify RHODS Release Version Number` and add it to` Tier1`

tests/Tests/100__deploy/100__installation/102__post_install.robot
jgarciao: @lugi0 we have to set TierX to all the tests (same as we do in Polarion)

lugi0: Since when?

jgarciao (approved):
lgtm

bdattoma (approved):
lgtm

PR #247, opened 2022-03-14
Added test case to verify if IBM resources have been removed

risusing:


lugi0:
Closing for now based on discussion during ods-ci spring cleaning.
@risusing please keep this branch in your local repo / fork because we might need this in the future.

PR #253, opened 2022-03-15
Fix HCC Splash page on login and increased timeout to check card existence on Enabled page

bdattoma:
This PR wants to fix two failures faced running Sanity/Smoke on RHODS v1.8.0.

### DIFF
--- a/tests/Resources/Page/HybridCloudConsole/HCCLogin.robot
+++ b/tests/Resources/Page/HybridCloudConsole/HCCLogin.robot
@@ -24,7 +24,7 @@ Login to HCC
     Input Text  id=password  ${password}
     Click Button    Log in
   END
-  Run Keyword And Continue On Failure    Wait For HCC Splash Page
+  Run Keyword And Ignore Error     Wait For HCC Splash Page
   Maybe Handle Something Went Wrong Page
 
 Maybe Skip RHOSAK Tour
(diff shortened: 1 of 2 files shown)


tests/Resources/Page/HybridCloudConsole/HCCLogin.robot
```diff
@@ -24,7 +24,7 @@ Login to HCC
     Input Text  id=password  ${password}
     Click Button    Log in
   END
-  Run Keyword And Continue On Failure    Wait For HCC Splash Page
+  Run Keyword And Ignore Error     Wait For HCC Splash Page
```
tarukumar, line 27: Just a suggestion, Do you think it will be good idea to use "**Run Keyword And Warn On Failure**" . It will not fail the overall TC but will give warning so that we can get to know ?

tarukumar (approved):
LGTM

lugi0 (approved):

milindw96 (approved):

tests/Resources/Page/HybridCloudConsole/HCCLogin.robot
bdattoma: @tarukumar I think it's not worth because we are testing the splash page "feature", so we are not really interested in if it appears or not, the important thing is that the next keywords run successfully

PR #266, opened 2022-03-22
Add Test Case To Verify GPU Tutorials In Resources Section

risusing:


tests/Tests/400_ods_dashboard/400_ods_dashboard.robot
```diff
@@ -124,3 +130,9 @@ Verify JupyterHub Card CSS Style
     CSS Property Value Should Be    locator=${SIDEBAR_TEXT_CONTAINER_XP}/h1
     ...    property=font-family    exp_value=RedHatDisplay
     ...    operation=contains
+
+Check GPU Tutorials
+    ${version-check} =  Is RHODS Version Greater Or Equal Than  1.9.0
+    IF  ${version-check}==True
+        Page Should Contain     //a[@href="https://github.com/ContinuumIO/gtc2018-numba"]
+    END
```
bdattoma, line 138: to a more complete check, I would also write the ELSE statement. If RHODS < 1.9.0 the tutorial should not be displayed

bdattoma: could you also try by getting the cards and checking that there is one card with that title and that link? so, if any new item has unexpectly been added, the test fail and we get acknoledged

risusing: okay so now I am checking for the article and searching using it's id and also checking for the link also.

bdattoma: that's fine, but it doesn't cover this part `so, if any new item has unexpectly been added, the test fail and we get acknoledged`.

PR #268, opened 2022-03-22
Add rhods-groups configuration task

pablofelix:
The task allows to change dynamically rhods groups by modifying the test variables in the command:
         ./run_robot_test.sh --test-case ./tasks/Tasks/rhods-groups-config.robot -i Custom-Groups --test-variable ADMIN_GROUPS:<admin_groups> --test-variable ALLOWED_GROUPS:<allowed_groups>

### DIFF
--- a/run_robot_test.sh
+++ b/run_robot_test.sh
@@ -88,7 +88,7 @@ case "$(uname -s)" in
          ;;
     Linux)
        case "$(lsb_release --id --short)" in
-       "Fedora"|"CentOS")
+       "Fedora"|"CentOS"|"RedHatEnterprise")
              ## Bootstrap script to setup drivers ##
              echo "setting driver  to $currentpath/Drivers/fedora"
              PATH=$PATH:$currentpath/drivers/fedora
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,7 @@ setup(
         "robotframework-requests",
         "robotframework-seleniumlibrary",
         "robotframework-jupyterlibrary>=0.3.1",
+        "robotframework-openshift",
         "robotframework-OpenShiftCLI==1.0.1",
         "ipython",
         "pytest",
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
         "robotframework-requests",
         "robotframework-seleniumlibrary",
         "robotframework-jupyterlibrary>=0.3.1",
-        "robotframework-openshift",
+        "robotframework-openshift==1.0.0",
         "robotframework-OpenShiftCLI==1.0.1",
         "ipython",
         "pytest",
--- a/tasks/Resources/Configurations/groups/rhods-groups-config.resource
+++ b/tasks/Resources/Configurations/groups/rhods-groups-config.resource
@@ -9,10 +9,15 @@ ${RHODS_GROUPS_CONFIG} =        tasks/Resources/Configurations/groups/rhods_grou
 
 
 *** Keywords ***
-Modify RHODS Groups
+Modify ODS Groups
     [Documentation]    (string, string) -> list(dict)
+    ...    Args:
+    ...        admin_groups(str): One or more comma separated admin groups
+    ...        allowed_groups(str): One or more comma separated allowed groups
+    ...    Returns:
+    ...        group_config(list(dict)): Groups data from configmap 
     [Arguments]    ${admin_groups}    ${allowed_groups}
-    ${modified_label} =    Set Rhods Groups Modified Label    ${admin_groups}    ${allowed_groups}
+    ${modified_label} =    Set ODS Groups Modified Label    ${admin_groups}    ${allowed_groups}
     &{template_data} =    Create Dictionary
     ...    admin_groups=${admin_groups}
     ...    allowed_groups=${allowed_groups}
@@ -22,8 +27,13 @@ Modify RHODS Groups
     Restart JupyterHub Pods
     [Return]    ${groups_config[0]['data']}
 
-Set Rhods Groups Modified Label
+Set ODS Groups Modified Label
     [Documentation]    (str, str) -> str
+    ...    Args:
+    ...        admin_groups(str): One or more comma separated admin groups
+    ...        allowed_groups(str): One or more comma separated allowed groups
+    ...    Returns:
+    ...        modified_label(str): Modify label to true or false depending on default groups  
     [Arguments]    ${admin_groups}    ${allowed_groups}
 
     IF    "${admin_groups}" == "${DEFAULT_ADMIN_GROUPS}" and "${allowed_groups}" == "${DEFAULT_ALLOWED_GROUPS}"
--- a/tasks/Resources/Configurations/groups/rhods-groups-config.resource
+++ b/tasks/Resources/Configurations/groups/rhods-groups-config.resource
@@ -1,0 +1,39 @@
+*** Settings ***
+Documentation    Keywords for task RHODS groups configuration
+
+
+*** Variables ***
+${DEFAULT_ADMIN_GROUPS} =       dedicated-admins
+${DEFAULT_ALLOWED_GROUPS} =     system:authenticated
+${RHODS_GROUPS_CONFIG} =        tasks/Resources/Configurations/groups/rhods_groups_config.yaml
+
+
+*** Keywords ***
+Modify RHODS Groups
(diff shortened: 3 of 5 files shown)


milindw96 (approved):

setup.py
```diff
@@ -20,6 +20,7 @@
         "robotframework-requests",
         "robotframework-seleniumlibrary",
         "robotframework-jupyterlibrary>=0.3.1",
+        "robotframework-openshift",
```
jgarciao, line 23: It should include exact library version so we control which version ods-ci is using

tasks/Resources/Configurations/groups/rhods-groups-config.resource
```diff
@@ -0,0 +1,39 @@
...
+
+
+*** Variables ***
+${DEFAULT_ADMIN_GROUPS} =       dedicated-admins
+${DEFAULT_ALLOWED_GROUPS} =     system:authenticated
+${RHODS_GROUPS_CONFIG} =        tasks/Resources/Configurations/groups/rhods_groups_config.yaml
+
+
+*** Keywords ***
+Modify RHODS Groups
```
jgarciao, line 12: - This keyword should be in tests/Resources/ODS.robot
- I think it's better to rename it to `Modify ODS Groups`
- [Documentation] tag should also explain better what this keyword is doing (e.g. customize ODS admin and allowed groups)
- You can also add a Restore Default ODS Groups, restoring the default configuration

tasks/Resources/Configurations/groups/rhods-groups-config.resource
```diff
@@ -0,0 +1,39 @@
...
+    ...    admin_groups=${admin_groups}
+    ...    allowed_groups=${allowed_groups}
+    ...    modified_label=${modified_label}
+    ${groups_config} =    Oc Apply    kind=ConfigMap    src=${RHODS_GROUPS_CONFIG}
+    ...    api_version=v1    template_data=${template_data}
+    Restart JupyterHub Pods
+    [Return]    ${groups_config[0]['data']}
+
+Set Rhods Groups Modified Label
+    [Documentation]    (str, str) -> str
```
jgarciao, line 26: better documentation needed

tasks/Resources/Configurations/groups/rhods-groups-config.resource
```diff
@@ -0,0 +1,39 @@
...
+    [Arguments]    ${admin_groups}    ${allowed_groups}
+
+    IF    "${admin_groups}" == "${DEFAULT_ADMIN_GROUPS}" and "${allowed_groups}" == "${DEFAULT_ALLOWED_GROUPS}"
+        Set Local Variable    ${modified_label}    "false"
+    ELSE
+        Set Local Variable    ${modified_label}    "true"
+    END
+    [Return]    ${modified_label}
+
+Restart JupyterHub Pods
```
jgarciao, line 36: I think we have already a keyword for this

tasks/Tasks/rhods-groups-config.robot
```diff
@@ -0,0 +1,28 @@
...
+Resource            ../Resources/Configurations/groups/rhods-groups-config.resource
+
+
+*** Variables ***
+${ADMIN_GROUPS} =       rhods-admins
+${ALLOWED_GROUPS} =     rhods-users
+
+
+*** Test Cases ***
+Configure Custom RHODS Groups Configuration
```
jgarciao, line 19: This should be a regular test (in a test suite) suite not a task in my opinion

bdattoma: could you check PR #208 ? maybe they are testing the same scenario

pablofelix: I used Test Cases, but it was by mistake, I wanted to use Tasks instead, because my purpose is to create a task and not a test, sorry for being confusing, I'm changing it. This task is meant to be used to modify groups as part of the cluster configuration after installing ODS, and all of this should be done before executing the tests. Also, this fix was needed to avoid the problems that we were having with new groups, and solving it with this task, allow us to modify the groups without changing the tests, by just modifying the Configmap when configuring the cluster, and therefore, it can be used directly by Jenkins. And under this kind of thinking, we can add more tasks for configurations  and this will give us the power to fully customize our clusters depending on our needs.
@jgarciao, I think this also answers your above comment about the folder that should contain it. 
@bdattoma this might clarify that your tests and the task can perfectly coexist, test the behaviour is also needed.

bdattoma: oh okay Pablo, if it is a Task is ok :)

lugi0: Hey @pablofelix , if I understand correctly a Task cannot be used inside a TC like a keyword can, if that's the case I think it'd be more useful to have this as a keyword rather than a task.
I say that because if this were a keyword, we would be able to dinamically change the group configuration before/after/during a specific TC execution.
As an example, I would personally turn this into a keyword and use it as test setup and test teardown for this suite: https://github.com/red-hat-data-services/ods-ci/blob/master/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot

bdattoma: a keyword to do that is present in PR #208 if you want (it may seem like I'm advertising my PR ahaha it's not my intent). You'll just need to set the Suite variable with the desired new groups

jgarciao:
@pablofelix  as @bdattoma mentioned in a previous comment in #208 two keywords were introduced to in ODS.robot that have a similar purpose:
- Apply Access Groups Settings
- Set Default Access Groups Settings

You can see how they are used in tests/Tests/500__jupyterhub/jupyterhub-user-access.robot

tasks/Resources/Configurations/groups/rhods-groups-config.resource
pablofelix: Point one is discussed in your last comment.
Points two and three are done. 
Point four, a restore keyword is not necessary because this is a task that will be used to change the groups dynamically, and there is no need to restore the default values after, as how it would happen in a test

tasks/Resources/Configurations/groups/rhods-groups-config.resource
pablofelix: done

tasks/Resources/Configurations/groups/rhods-groups-config.resource
pablofelix: I haven't found any keyword that implements the same as this one

pablofelix:
@jgarciao @lugi0  I'm very much in favor of reusing code, but for me, this is a special case, and keeping in mind all that has been already said about it, I still truly believe that tasks and tests should be independent identities, and therefore, don't share code, and even could be taken into consideration to place them in different repos.
Think a minute about this scenario, let's say that we decide to share keywords between tests and tasks, and for some reason, one of this keywords is modified introducing a failure. This failure will cause some task to fail as well. We shouldn't compromise the tasks, because they will be the first step that will allow everything else to be executed, and hence its importance.

pablofelix:
@jgarciao about using Apply Access Groups Settings keyword, I see that the keyword is using Run oc, and I think that we should start to replace those Run oc gradually. The reason is that the Run keyword, as it is stated in the documentation, just runs the keyword and returns the output, and this output, is what contains the standard output produced by the command, including the errors that Robot Framework redirect to the standard output as well. So even if the command fails and this output contains an error, the test passes, and we could be getting false negatives in our tests.
About using Set Default Access Groups Settings, I don't see it necessary to restore values in a task, as I explained in a previous comment

tasks/Resources/Configurations/groups/rhods-groups-config.resource
bdattoma: you can try `Rollout JupyterHub` stored in `tests/Resources/Page/ODH/JupyterHub/HighAvailability.robot`

run_robot_test.sh
```diff
@@ -88,7 +88,7 @@ case "$(uname -s)" in
          ;;
     Linux)
        case "$(lsb_release --id --short)" in
-       "Fedora"|"CentOS")
+       "Fedora"|"CentOS"|"RedHatEnterprise")
```
bdattoma, line 91: why do we need to add RH Enterprise here? Was there an issue running robot?

bdattoma:
lgtm, is it possible to test it on jenkins?

PR #284, opened 2022-03-25
Add test case to verify jupyterlab and notebook version in Anaconda i…

risusing:
…mage

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -6,6 +6,7 @@ Resource  ../../ODH/ODHDashboard/ODHDashboard.robot
 Resource  LoginJupyterHub.robot
 Resource  JupyterLabSidebar.robot
 Resource  ../../OCPDashboard/InstalledOperators/InstalledOperators.robot
+Library   ../../../../libs/Helpers.py
 Library   String
 Library   Collections
 Library   JupyterLibrary
@@ -365,3 +366,15 @@ Verify Image Can Be Spawned
     Spawn Notebook With Arguments    ${retries}    ${image}    ${size}
     ...    ${spawner_timeout}    ${gpus}    ${refresh}    &{envs}
     End Web Test
+
+Verify Jupyter Lab Version
+    [Arguments]  ${target}
+    ${ver} =  Run Cell And Get Output   !pip show JupyterLab | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s.%s", b[1], b[2], b[3]}'
+    ${comparison} =  GTE  ${ver}  ${target}
+    [Return]  ${comparison}
+
+Verify Notebook Lab Version
+    [Arguments]  ${target}
+    ${ver} =  Run Cell And Get Output   !pip show notebook | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s.%s", b[1], b[2], b[3]}'
+    ${comparison} =  GTE  ${ver}  ${target}
+    [Return]  ${comparison}
--- a/tests/Tests/600__ai_apps/600__anaconda_commercial_edition/600__anaconda_commercial_edition.robot
+++ b/tests/Tests/600__ai_apps/600__anaconda_commercial_edition/600__anaconda_commercial_edition.robot
@@ -50,7 +50,7 @@ Verify Anaconda Commercial Edition Fails Activation When Key Is Invalid
 
 Verify User Is Able to Activate Anaconda Commercial Edition
   [Tags]  Tier2
-  ...     ODS-272  ODS-344  ODS-501  ODS-588
+  ...     ODS-272  ODS-344  ODS-501  ODS-588  ODS-1082
   ...     KnownIssues
   [Documentation]  Performs the Anaconda CE activation, spawns a JL using the Anaconda image,
   ...              validate the token, install a library and try to import it.
@@ -87,6 +87,8 @@ Verify User Is Able to Activate Anaconda Commercial Edition
   Maybe Open JupyterLab Sidebar   File Browser
   Fix Spawner Status  # used to close the server and go back to Spawner
   Wait Until Page Contains Element  xpath://input[@name='Anaconda Commercial Edition']  timeout=15
+  Verify Jupyter Lab Version    3.1.4
+  Verify Notebook Lab Version   6.4.1
   [Teardown]    Remove Anaconda Commercial Edition Component
 
 
--- a/tests/Resources/Page/ODH/AiApps/Anaconda.resource
+++ b/tests/Resources/Page/ODH/AiApps/Anaconda.resource
@@ -42,7 +42,7 @@ Validate Anaconda License Key
     [Arguments]    ${license_validity}=${TRUE}
     Click Button    Connect
     IF    ${license_validity} == ${TRUE}
-        Wait Until Keyword Succeeds    50    1    Page Should Not Contain Element
+        Wait Until Keyword Succeeds    60    1    Page Should Not Contain Element
         ...    xpath://*/div[contains(@class, "bullseye")]
     ELSE IF    ${license_validity} == ${FALSE}
         Wait Until Keyword Succeeds    30    1    Check Connect Button Status    false


tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -357,3 +358,15 @@ Verify Image Can Be Spawned
...
+    [Arguments]  ${target}
+    ${ver} =  Run Cell And Get Output   !pip show JupyterLab | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s.%s", b[1], b[2], b[3]}'
+    ${comparison} =  GTE  ${ver}  ${target}
+    [Return]  ${comparison}
+
+Verify Notebook Lab Version
+    [Arguments]  ${target}
+    ${ver} =  Run Cell And Get Output   !pip show notebook | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s.%s", b[1], b[2], b[3]}'
+    ${comparison} =  GTE  ${ver}  ${target}
+    [Return]  ${comparison}
```
bdattoma, line 372: ods-ci already contains keywords to check the version of lbraries. Pls have a look at `Verify Installed Library Version` and `Check Versions In JupyterLab`

bdattoma:
@risusing could you add the Polarion IDs?
And pls review the test case on polarion because I extended it a bit today to make it more generic. Thank  you!

risusing:
Added the polarian ID. And the major version for jupyter lab i.e 3.1 is already implemented which satisfies the need of test case.

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
bdattoma: could you create a kw (e.g., `Verify Library Version Is Greater Than`) which takes as args the version and the lib name? so we at least avoid the code duplication between `Verify Jupyter Lab Version` and `Verify Notebook Lab Version`

risusing:
Jenkins Job Run : https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/577/
Result : Passed

tests/Tests/600__ai_apps/600__anaconda_commercial_edition/600__anaconda_commercial_edition.robot
```diff
@@ -50,7 +50,7 @@ Verify Anaconda Commercial Edition Fails Activation When Key Is Invalid
 
 Verify User Is Able to Activate Anaconda Commercial Edition
   [Tags]  Tier2
-  ...     ODS-272  ODS-344  ODS-501  ODS-588
+  ...     ODS-272  ODS-344  ODS-501  ODS-588  ODS-1082
```
milindw96, line 53: please  reformat in a new style :-)

milindw96 (changes requested):

milindw96 (approved):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -365,3 +366,9 @@ Verify Image Can Be Spawned
     Spawn Notebook With Arguments    ${retries}    ${image}    ${size}
     ...    ${spawner_timeout}    ${gpus}    ${refresh}    &{envs}
     End Web Test
+
+Verify Library Version Is Greater Than
+    [Arguments]     ${library}      ${target}
+    ${ver} =  Run Cell And Get Output   !pip show ${library} | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s.%s", b[1], b[2], b[3]}'
+    ${comparison} =  GTE  ${ver}  ${target}
+    [Return]  ${comparison}
```
tarukumar, line 374: Why are you returning the Status ? I don't see you are using it anywhere

risusing: If you are talking about the comparison status then the task is all about that only. I have to compare with the expected version.

tarukumar: I know it about comparison why are you not returning the comparison if you are not using it in your test case? What will happens if version is not greater than, it should FAIL. Could you verify with wrong version and sen me the output via chat for negative test case?

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -365,3 +366,11 @@ Verify Image Can Be Spawned
     Spawn Notebook With Arguments    ${retries}    ${image}    ${size}
     ...    ${spawner_timeout}    ${gpus}    ${refresh}    &{envs}
     End Web Test
+
+Verify Library Version Is Greater Than
+    [Arguments]     ${library}      ${target}
+    ${ver} =  Run Cell And Get Output   !pip show ${library} | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s.%s", b[1], b[2], b[3]}'
+    ${comparison} =  GTE  ${ver}  ${target}
+    IF  ${comparison}==False
+        FAIL    Library Version Is Smaller Than Expected
```
tarukumar, line 375: I think the better option will be **Run Keyword And Continue On Failure** otherwise you whole test case will fail. I think you should avoid that

risusing: I have used **Run Keyword And Continue On Failure** while using this keyword. Even if this keyword fails the whole test case should complete that's why I have used it.

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -365,3 +366,11 @@ Verify Image Can Be Spawned
...
     ...    ${spawner_timeout}    ${gpus}    ${refresh}    &{envs}
     End Web Test
+
+Verify Library Version Is Greater Than
+    [Arguments]     ${library}      ${target}
+    ${ver} =  Run Cell And Get Output   !pip show ${library} | grep Version: | awk '{split($0,a); print a[2]}' | awk '{split($0,b,"."); printf "%s.%s.%s", b[1], b[2], b[3]}'
+    ${comparison} =  GTE  ${ver}  ${target}
+    IF  ${comparison}==False
+        Run Keyword And Continue On Failure     FAIL    Library Version Is Smaller Than Expected
+    END
```
lugi0, line 376: I don't think you should do this in the keyword, IMHO it would be better to do this check in the TC if you need to do it at all

tests/Tests/600__ai_apps/600__anaconda_commercial_edition/600__anaconda_commercial_edition.robot
```diff
@@ -84,12 +84,13 @@ Verify User Is Able to Activate Anaconda Commercial Edition
   Wait Until JupyterLab Code Cell Is Not Active
   Run Cell And Check For Errors  import numpy as np
   Capture Page Screenshot  conda_lib_install_result.png
+  Verify Library Version Is Greater Than  jupyterlab  3.1.4
+  Verify Library Version Is Greater Than  notebook    6.4.1
```
lugi0, line 88: You should only check for `x.y` as that is what we do in all other TCs and what we report in the documentation. The `z` version is bound to change frequently

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
risusing: I have did here just because whenever I'm using this keyword I don't have to write there also it will make the code redundant.

tests/Tests/600__ai_apps/600__anaconda_commercial_edition/600__anaconda_commercial_edition.robot
bdattoma: this TC is about checking the security issue on jupyter images whose solution was to have specific library versions. And the security fix has been implemented into .z version. https://blog.jupyter.org/cve-2021-32797-and-cve-2021-32798-remote-code-execution-in-jupyterlab-and-jupyter-notebook-a70fae0d3239

risusing:
Tested locally on a lower jupyterlab version and it failed as expected.
Screenshot from 2022-03-31 15-37-22

bdattoma (approved):

lugi0 (approved):
lgtm

PR #289, opened 2022-03-29
increaes timeout for PVC

tarukumar:
Fixing issue for PVC timeout after setting value to default 


tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -400,6 +400,7 @@ Restore PVC Value To Default Size
     Click Button    Restore Default
     Wait Until Keyword Succeeds    30    1
     ...    Wait Until Page Contains    Cluster settings updated successfully.
+    Sleep    120s
```
jgarciao, line 403: Could you add a reason for the Sleep? 

e.g.  ` Sleep    120s    reason=Wait until configuration is applied and traefik-proxy is restored`

jgarciao (commented):
@tarukumar I think until ODS-1220 and ODS-1221 are more tested they should be only in Sanity and Tier1, but not in Smoke

tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
tarukumar: Done

tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -400,6 +400,7 @@ Restore PVC Value To Default Size
     Click Button    Restore Default
     Wait Until Keyword Succeeds    30    1
     ...    Wait Until Page Contains    Cluster settings updated successfully.
+    Sleep    120s   reason=Wait until JupyterHub leader pod is elected
```
bdattoma, line 403: have a look at pr #208 it defines a keyword to handle Bad Gateway page after rolling out. `Handle Bad Gateway Page`.

Since #208  has not been merged yet, you could just copy the keyword here

jgarciao: Another option is to run   `Verify JupyterHub Deployment` inside `Wait Until Keyword Succeeds `  similar as it's done [here](https://github.com/red-hat-data-services/ods-ci/blob/59a3ec8bdc75fcc84247bf2d147f7a0a0176679d/tests/Resources/Page/ODH/Prometheus/Prometheus.robot#L169)

PR #292, opened 2022-03-31
Add test to verify jupyterlab and notebook library versions for Openvino operator

risusing:


tests/Resources/Page/ODH/AiApps/Openvino.robot
```diff
@@ -24,4 +24,6 @@ Verify JupyterHub Can Spawn Openvino Notebook
     Wait Until Element Is Enabled    xpath://input[@name="OpenVINO™ Toolkit"]    timeout=10
```
bdattoma, line 24: could you add the Polarion ID to this test case?

risusing: yes I have added it n the test where this keyword is used

milindw96 (approved):

PR #294, opened 2022-03-31
Fix bug in test to verify GPU learning resources in RHODS Resource Section

risusing:


bdattoma (approved):
lgtm. Please test it on the live build before merging

risusing:
I've tested on the live build and that's the reason it's unable to check for the version and that's why I forced version-check to TRUE. And results are as expected.
Screenshot from 2022-04-01 13-32-53

jgarciao (approved):
lgtm

PR #307, opened 2022-04-13
Add email report feature to ods-ci plus RF task compatibility

bdattoma:
with this PR ods-ci supports the sharing of RF artifacts via e-mails. There are some parameters to tune:
1) Enable reporting:  --email-report true/false. Defaulted to "true" 
2) Sender email: --email-from <youremail@mail.yxz>. e.g., ods-ci-results@redhat.com 
3) Receiver email: --email-to <youremail@mail.yxz>, e.g., ods-ci@redhat.com
4) SMTP server:  --server <smtp.server>.
5) Enable SSL: --email-server-ssl true/false (defaulted to "false")
6) use unsecure connection (no encryption): --email-server-unsecure true/false (defaulted to "false")
if 5) and 6) are set as "false", TLS encryption is used automatically by using `starttls` command

**attachments**
The script compress and attaches to the email all the content of output directory. If the compressed content is greater than 20MB, then the script will send only html and xml reports.

**SMTP Server**
There are some possibilities here:
1. using a local smtp server. This PR includes `Dockerfile_smtpserver` which runs a `postfix` mail server. If you go with this option, you can omit parameter 4) because it is defaulted to `localhost`. You could use `ods-ci.pod_with_postfix.yaml` to deploy the local smtp server in a dedicated container in the same pod as ods-ci
2. Use RH corporate smtp server. This options only works if the cluster is under RH network (i.e., PSI or your machine)
3. Use GMAIL or any other third party servers. (gmail has been tested)

Options 1. and 2. don't work on OSD. You should use third party smtp servers like gmail.

More instructions will be reported in the README


aloganat (approved):
Looks good to me

mwaykole (approved):
lgtm

PR #325, opened 2022-04-26
added more deley time to add idp oauth

tarukumar:
it came to our notice that sometime after adding the idp successfully we are not able to see the idp auth login. we suspect that it is the problem with the waiting time. Have increded the waiting time from 5 to 10 min. This is not the ideal solution. one of the Ideal solution could be wait for the osd version to get display then apply idp.


mwaykole (approved):

bdattoma:
if it could help, I faced the same problem in the kuttl scorecard for MSP. I found 3 minutes as the time to wait until the IdP is available in the login page.

bdattoma:
closing since this is already handled in CI/CD

PR #329, opened 2022-04-27
remove explicit secret creation from RHOAM Interop test

bdattoma:
This PR wants to cover [RHODS-2738](https://issues.redhat.com/browse/RHODS-2738). Now the test just checks if the required secrets are present and fails if they are not.
 

bdattoma:
jenkins job: [779](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/779/console). 
ABORTED. 
It is not possible to full testing this right now because of a bug on RHOAM installation process. However, the changes applied by this PR have been executed correctly since the installation has started correctly. If there were any errors on the changes here applied, then the installation wouldn't have started because of missing secrets.

So, to me, the changes have been tested but the TC cannot be successful due to other bugs (not related to ODS-CI or RHODS)

jgarciao (dismissed):
lgtm

libs/Helpers.py
```diff
@@ -47,7 +47,9 @@ def gte(self, version, target):
     def install_rhoam_addon(self, cluster_name):
         ocm_client = OpenshiftClusterManager()
         ocm_client.cluster_name = cluster_name
-        ocm_client.install_rhoam_addon(exit_on_failure=False)
+        result = ocm_client.install_rhoam_addon(exit_on_failure=False)
+        if not result:
+            raise FatalError("Something got wrong while installing RHOAM. Check the logs")
```
jgarciao, line 52: Please, replace FatalError by built-in fail

jgarciao (changes requested):

libs/Helpers.py
bdattoma: done

tarukumar (approved):
LGTM

jgarciao (approved):
lgtm

PR #330, opened 2022-04-28
change notbook size

tarukumar:

We have are covering enough negative scenario removing one . it seems unnecessary to have in the input varibale. removing the value


mwaykole (approved):

aloganat (approved):
LGTM

PR #334, opened 2022-04-29
Replacing load witrh safe_load for yaml  lib 

mwaykole:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,8 @@ setup(
         "robotframework-requests",
         "escapism",
         "semver>=2,<3",
-        "yq"
+        "yq",
+        "jinja2"
     ],
     zip_safe=True,
     include_package_data=True,
--- a/setup.py
+++ b/setup.py
@@ -31,8 +31,7 @@ setup(
         "robotframework-requests",
         "escapism",
         "semver>=2,<3",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -31,7 +31,8 @@
         "robotframework-requests",
         "escapism",
         "semver>=2,<3",
-        "yq"
+        "yq",
+        "Jinja2"
```
aloganat, line 35: Can u add pyyaml ? and wherever yaml.load() is used change it to yaml.safe_load(), this is to avoid this error message
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.

mwaykole: ack will do the changes

tarukumar:
@mwaykole Do we need to install Jinja2 explicitly since we install robotframework-openshift library(present in setup.py) which already has Jinja2 included in it.

mwaykole:
hi closing this pr , as looks like we don't need it

tarukumar (approved):

tarukumar:
LGTM

aloganat (approved):
LGTM

PR #341, opened 2022-05-04
Modify update_notification_email_address keyword so that it doesn't stop test execution on errors

tarukumar:
When we are using Update Notification Email Address keyword if keyword execution fails then it will stop whole robot execution du to sys.exit(1)  process chnaging it so that it return status

### DIFF
--- a/utils/scripts/ocm/ocm.py
+++ b/utils/scripts/ocm/ocm.py
@@ -852,7 +852,7 @@ class OpenshiftClusterManager():
         if ret is None:
             log.info("Failed to update email address to {} addon on cluster "
                   "{}".format(addon_name, self.cluster_name))
-            sys.exit(1)
+            return False
 
 
 
--- a/utils/scripts/ocm/ocm.py
+++ b/utils/scripts/ocm/ocm.py
@@ -855,7 +855,7 @@ class OpenshiftClusterManager():
             if exit_on_failure:
                 sys.exit(1)
             else:
-                return True
+                return False
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


tarukumar:
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/796/console

utils/scripts/ocm/ocm.py
```diff
@@ -852,7 +852,7 @@ def update_notification_email_address(self, addon_name, email_address):
         if ret is None:
             log.info("Failed to update email address to {} addon on cluster "
                   "{}".format(addon_name, self.cluster_name))
-            sys.exit(1)
+            return False
```
bdattoma, line 855: sys.exit I think is necessary when these API calls are done by Jenkins. So, I would suggest to give the option to use sys.exit or return, based on who is calling the function.
There are examples of what I'm suggesting in the same ocm.py

tarukumar: We are not using jenkins API to directly executeg this method so why i didn't implement it. If you would like i can do that as well

tarukumar: i have added it

jgarciao (approved):
lgtm

utils/scripts/ocm/ocm.py
bdattoma: @tarukumar yeah I know, since it's an easy change I thought it is preferable to do that to maintain consistency with the resto of ocm.py. If in the future we are going to use the function in jenkins, we won't need any change

bdattoma:
could you test it again? 
Is the failure expected?

PR #349, opened 2022-05-09
Add support to provide notification email during rhods installation

aloganat:

### DIFF
--- a/utils/scripts/ocm/ocm.py
+++ b/utils/scripts/ocm/ocm.py
@@ -50,6 +50,7 @@ class OpenshiftClusterManager():
         self.taints = args.get("taints")
         self.pool_name = args.get("pool_name")
         self.reuse_machine_pool = args.get("reuse_machine_pool")
+        self.notification_email = args.get("notification_email")
 
         ocm_env = glob.glob(dir_path+"/../../../ocm.json.*")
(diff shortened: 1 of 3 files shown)


mwaykole (dismissed):
LGTM

utils/scripts/ocm/ocm.py
```diff
@@ -450,7 +451,8 @@ def install_addon(self, addon_name="managed-odh",
         """Installs addon"""
         replace_vars = {
                        "CLUSTER_ID": self.cluster_name,
-                       "ADDON_NAME": addon_name
+                       "ADDON_NAME": addon_name,
+                       "NOTIFICATION_EMAIL": self.notification_email
```
bdattoma, line 455: could you use the `add_replace_vars` parameter to set the notification email? Because this function should be generic enough to work with any addon. As of now, I don't know if RHOAM would work setting here the notification mail.

You can have a look at `install_rhoam_addon` to see how I used the `add_replace_var` param

lugi0 (commented):
LGTM but I would suggest investigating Berto's comment

bdattoma (approved):
LGTM

lugi0 (approved):
lgtm

PR #350, opened 2022-05-10
Fix CUDA suite setup failure

lugi0:

This change allows the suite to continue execution even if the suite setup fails the check on the number of available GPUs, otherwise all other tests won't be run at all.


mwaykole (approved):
LGTM

jgarciao (approved):
lgtm

PR #372, opened 2022-05-18
update ods-ci container to install python3.8 and use RH ubi image

bdattoma:
This PR upgrades python version in ODS-CI container and replace the centos quay image with a RH UBI image

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,6 +1,5 @@
 FROM registry.access.redhat.com/ubi8/ubi
-# FROM quay.io/centos/centos:stream8
-# FROM registry.access.redhat.com/ubi8/ubi:8.1
+
 # Use this build arg to set any default test script arguments
 ENV RUN_SCRIPT_ARGS=${RUN_SCRIPT_ARGS}
 ENV ROBOT_EXTRA_ARGS=${ROBOT_EXTRA_ARGS}
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -9,9 +9,8 @@ ARG OC_CHANNEL=stable
 
 
 RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm &&\
-    dnf install -y jq git unzip chromium chromedriver redhat-lsb-core &&\
-    dnf clean all && \
-    yum install -y python38
+    dnf install -y python38 jq git unzip chromium chromedriver redhat-lsb-core &&\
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -9,8 +9,9 @@ ARG OC_CHANNEL=stable
 
 
 RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm &&\
-    dnf install -y jq git python3 unzip chromium chromedriver redhat-lsb-core &&\
-    dnf clean all
+    dnf install -y jq git unzip chromium chromedriver redhat-lsb-core &&\
+    dnf clean all && \
+    yum install -y python38
```
lugi0, line 14: why not do all installs in one line?

build/Dockerfile
```diff
@@ -39,6 +40,7 @@ COPY utils/scripts/logger.py  utils/scripts/logger.py
 COPY utils/scripts/util.py  utils/scripts/util.py
 RUN  chmod +x run.sh
 COPY requirements.txt setup.py .
+RUN  python3 --version
```
lugi0, line 43: Do you need to do this if you don't check the version? Is this just to have it logged somewhere?

build/Dockerfile
bdattoma: I find this way more readable. Would there be any benefits in doing in one line?

build/Dockerfile
bdattoma: it's not needed, but I wanted to have this logged. Do you prefer to remove it?

build/Dockerfile
lugi0: eh, not really, other than the fact that I would put the `clean all` at the end of all installs.
I would also either always use `dnf` or always use `yum`

build/Dockerfile
lugi0: No, having it logged might be useful if we ever run into weird compatibility issues

build/Dockerfile
lugi0: Actually, do you even need the clean all here?

bdattoma: so, I replaced yum with dnf. 

About the `clean all`, it helps in reducing the size of the container (even though this docker file will be still changed to optimize the final size of the container)

lugi0 (approved):
lgtm

mwaykole (approved):

PR #375, opened 2022-05-18
Update tensorflow version check in minimal image

lugi0:


pablofelix (approved):
lgtm

jgarciao (approved):
Code looked cleaner before, but I guess with this change you don't need to update the newest tensorflow version every time there is a new release

PR #378, opened 2022-05-23
Update ODS-CI container with new features

bdattoma:
This PR contains new developments for the ODS-CI container:
1. **SA Login**. A new (user)option to allow the robot to do "oc login" using the Service account instead of using user/password. It is necessary to make our tests running in MSP when they need to explicitly log into the cluster (e.g., use of `Run oc` or `openshiftlibrary for RF`). The script `run_robot_test.sh` automatically sets the following global variables to be used in the tests:
`SERVICE_ACCOUNT.NAME` (e.g., `rhods-test-runner`) and `SERVICE_ACCOUNT.FULL_NAME` (e.g., `system:serviceaccount:<namespace>:<service_account_name>`)
Plus, it changes `Clean Up User Notebook` and `Delete Folder In User Notebook` to run when the login is performed using SA instead of admin credentials

2. **Auto-set Prometheus variables**. `run_robot_test.sh` automatically fetches the RHODS Prometheus URL and its token, and set the global variables to be used in the tests: `RHODS_PROMETHEUS_URL` and `RHODS_PROMETHEUS_TOKEN`

3. **Update container README**

Signed-off-by: bdattoma [bdattoma@redhat.com](bdattoma@redhat.com)

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -29,6 +29,8 @@ Additional arguments for container run
 # env variables to control test execution
 RUN_SCRIPT_ARGS:
   --skip-oclogin (default: false): script does not perform login using OC CLI
+    --service-account (default: ""): if assigned, ODS-CI will try to log into the cluster using the given service account. ODS-CI automatically creates SERVICE_ACCOUNT.NAME and SERVICE_ACCOUNT.FULL_NAME global variables to be used in tests.
+    --sa-namespace (default: "default"): the namespace where the service account is created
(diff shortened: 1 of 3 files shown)


bdattoma:
Tested by running Smoke suite in the container on stage cluster: 
image
All PASS including Alerts/Metrics test cases which use the prometheus URL (automatically set)
There are Warnings raised by a test for PVC size changes, but these shouldn't be related to the changes in this PR (pinged @tarukumar about this)

lugi0 (approved):
lgtm

pablofelix (approved):
lgtm

PR #388, opened 2022-05-27
fix ODS-293 by adding wait for rollout to start

bdattoma:
PR to fix the flaky ODS-293. It adds a keyword to wait until the rollout is actually started

### DIFF
--- a/tests/Resources/Page/OCPDashboard/DeploymentConfigs/DeploymentConfigs.robot
+++ b/tests/Resources/Page/OCPDashboard/DeploymentConfigs/DeploymentConfigs.robot
@@ -1,4 +1,5 @@
 *** Settings ***
+Documentation       Collections of keyword to handle Deployment config objects in OCP
 Library     OpenShiftLibrary
 
 
@@ -15,8 +16,8 @@ Start Rollout
 
 Wait Until Rollout Is Started
     [Documentation]     Wait until the old pods are replace with the new ones after a rollout is triggered
-    [Arguments]     ${namespace}  ${label_selector}   ${comparison_fields}=['metadata.name']  ${retries}=5   ${retries_interval}=5s
-    ${previous_pods}=    Oc Get    kind=Pod  namespace=${namespace}   label_selector=${label_selector}   fields=${comparison_fields}
+    [Arguments]     ${previous_pods}    ${namespace}    ${label_selector}
+    ...             ${comparison_fields}=['metadata.name']    ${retries}=5    ${retries_interval}=5s
     Log     ${previous_pods}
     FOR  ${retry_idx}  IN RANGE  0  1+${retries}
         ${current_pods}=    Oc Get    kind=Pod  namespace=${namespace}   label_selector=${label_selector}   fields=${comparison_fields}
--- a/tests/Resources/Page/OCPDashboard/DeploymentConfigs/DeploymentConfigs.robot
+++ b/tests/Resources/Page/OCPDashboard/DeploymentConfigs/DeploymentConfigs.robot
@@ -17,10 +17,10 @@ Wait Until Rollout Is Started
     [Documentation]     Wait until the old pods are replace with the new ones after a rollout is triggered
     [Arguments]     ${namespace}  ${label_selector}   ${comparison_fields}=['metadata.name']  ${retries}=5   ${retries_interval}=5s
     ${previous_pods}=    Oc Get    kind=Pod  namespace=${namespace}   label_selector=${label_selector}   fields=${comparison_fields}
-    Log To Console      ${previous_pods}
+    Log     ${previous_pods}
     FOR  ${retry_idx}  IN RANGE  0  1+${retries}
         ${current_pods}=    Oc Get    kind=Pod  namespace=${namespace}   label_selector=${label_selector}   fields=${comparison_fields}
-        Log To Console      ${current_pods}
+        Log      ${current_pods}
         ${equal_flag}=     Run Keyword And Return Status    Should Not Be Equal As Strings    ${previous_pods}    ${current_pods}
         Exit For Loop If    $equal_flag == True
         Sleep    ${retries_interval}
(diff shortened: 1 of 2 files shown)


mwaykole (approved):

tests/Resources/Page/OCPDashboard/DeploymentConfigs/DeploymentConfigs.robot
```diff
@@ -8,3 +12,19 @@ Start Rollout
     [Documentation]     Rollout a deployment in Openshift fetching the latest version
     [Arguments]  ${dc_name}  ${namespace}
     Run    oc rollout latest dc/${dc_name} -n ${namespace}
+
+Wait Until Rollout Is Started
```
jgarciao, line 16: If it works we can merge it now to have it fixed today, but I think it could be improved in a future PR:
- It could be renamed to `Wait Until Rollout Is Finished` and wait until the pods are different and also until they are in  Running state (you can use the keyword `Get Pod Status`)
- Instead of looping you can use `Wait Until Keyword Succeeds` like [in this example](https://github.com/red-hat-data-services/ods-ci/blob/9e507824fc6a0268b8c4ca2f79f7e4813fa33e19/tests/Resources/Page/ODH/Prometheus/Prometheus.robot#L191)

jgarciao (approved):
Please, test it before merging the PR

tests/Resources/Page/OCPDashboard/DeploymentConfigs/DeploymentConfigs.robot
bdattoma: 1. it is not waiting until rollout is finished. It just waits until it actually starts. By analyzing the failures, it came out that the rollout didn't start immediatedly and the `Wait Until JH Deployment Is Ready` (which do the checks on the "running" status etc) passed without actually waiting for it.
2. I don't think I can use it because I need two run two keyword: lines 22-24. I could use Run Keywords in combination with `Wait Until Keyword Succeeds` but I feel it is not very readable, but if you prefer I can do that

bdattoma:
Jenkins Job: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/914/ _SUCCESS_

I want to run it another time plus running a Sanity before merging the PR

tests/Resources/Page/ODH/JupyterHub/HighAvailability.robot
```diff
@@ -100,6 +100,7 @@ Wait Until JH Deployment Is Ready
 Rollout JupyterHub
     [Documentation]     Rollouts JupyterHub deployment and wait until it is finished
     Start Rollout   dc_name=jupyterhub  namespace=redhat-ods-applications
+    Wait Until Rollout Is Started       namespace=redhat-ods-applications  label_selector=app=jupyterhub
```
lugi0, line 103: I think you should pass the `${previous_pods}` as an argument to `Wait Until Rollout Is Started`, and you fetch them before doing `Start Rollout`. In the cases in which the rollout is started by saving changes to the settings the same strategy could be applied, and the previous pods are fetched before saving the changes.
It feels like there is a risk of `Wait until rollout is started` failing because it fetches the new pods when doing the first `OC get` otherwise.

bdattoma: hm, yes you're right. I will change it, thanks!

bdattoma: done

tests/Resources/Page/OCPDashboard/DeploymentConfigs/DeploymentConfigs.robot
jgarciao: I think it could be useful for other tests to verify that the rollout has been _finished_  (not just started), in order to not be necessary to implement an extra keyword for each deployment to check if it actually finished

I was thinking in something like this:

```
Verify Deployment Pods Have Been Updated
    [Arguments]     ${namespace}  ${label_selector}   ${comparison_fields}=['metadata.name']   ${previous_pods}
     ${current_pods}=    Oc Get    kind=Pod  namespace=${namespace}   label_selector=${label_selector}   fields=${comparison_fields}
     Should Not Be Equal As Strings    ${previous_pods}    ${current_pods}

Wait Until Rollout Is Finished
    [Documentation]     Wait until the old pods are replace with the new ones after a rollout is triggered
    [Arguments]     ${namespace}  ${label_selector}   ${comparison_fields}=['metadata.name']  ${retry}=5   ${retry_interval}=5s
    
    ${previous_pods}=    Oc Get    kind=Pod  namespace=${namespace}   label_selector=${label_selector}   fields=${comparison_fields}
    
    Wait Until Keyword Succeeds    ${retry}    ${retry_interval}
    ...    Verify Deployment Pods Have Been Updated    ${namespace}  ${label_selector}   ${comparison_fields} ${previous_pods}
    
    Wait Until Pod Status Is  ${namespace}  ${label_selector}   ${comparison_fields}   status=Running


```

bdattoma: Agree, but in this particular case we wouldn't use this because Jupyterhub needs its own keyword to check if the deployment is ready (which automatically says if the rollout is finished).
So, I would avoid to implement in this PR a keyword which is not needed for this test case. Could we implement in the future when needed?

jgarciao: Sure!

lugi0 (approved):
lgtm

tests/Resources/Page/OCPDashboard/DeploymentConfigs/DeploymentConfigs.robot
bdattoma: I took a note about this. Thanks

jgarciao (approved):
lgtm

PR #407, opened 2022-06-10
venv: only create when absent

fcami:
run_robot_test.sh would create the venv at each run.
Only create it if absent.

Signed-off-by: François Cami <fcami@redhat.com>

### DIFF
--- a/run_robot_test.sh
+++ b/run_robot_test.sh
@@ -278,10 +278,10 @@ fi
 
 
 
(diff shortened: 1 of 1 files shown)


fcami:
hi @bdattoma is this satisfactory?

bdattoma (approved):

mwaykole (commented):
@fcami  i would suggest to run smoke with  the changes
@bdattoma  your thoughts?

PR #419, opened 2022-06-16
verify dashboard pod is not getting restarted

rkamble5:
Signed-off-by: rkamble <rkamble@redhat.com>

### DIFF
--- a/tests/Tests/400__ods_dashboard/400__ods_dashboard.robot
+++ b/tests/Tests/400__ods_dashboard/400__ods_dashboard.robot
@@ -32,7 +32,7 @@ ${RHOSAK_DISPLAYED_APPNAME}             OpenShift Streams for Apache Kafka
 ...                                     Securing a deployed model using Red Hat OpenShift API Management
 @{EXPECTED_ITEMS_FOR_COMBINATIONS}      Create List                                                         JupyterHub    OpenShift API Management    OpenShift Streams for Apache Kafka
 ...                                     PerceptiLabs
-
+${DASHBOARD_LABEL_SELECTOR}       app=rhods-dashboard
 
 *** Test Cases ***
 Verify That Login Page Is Shown When Reaching The RHODS Page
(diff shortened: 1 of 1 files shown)


tests/Tests/400__ods_dashboard/400__ods_dashboard.robot
```diff
@@ -188,6 +188,13 @@ Verify Favorite Resource Cards
     Favorite Items Should Be Listed First When Sorted By    ${favorite_ids}    duration
     [Teardown]    Remove Items From favorites    @{favorite_ids}
 
+Verify Dashboard Pod Is Not Getting Restarted	
+    [Documentation]    Verify Dashboard Pod container restart
+    [Tags]    Sanity
```
jgarciao, line 193: Can you add the Tier1 tag too

tests/Tests/400__ods_dashboard/400__ods_dashboard.robot
```diff
@@ -188,6 +188,13 @@ Verify Favorite Resource Cards
     Favorite Items Should Be Listed First When Sorted By    ${favorite_ids}    duration
     [Teardown]    Remove Items From favorites    @{favorite_ids}
 
+Verify Dashboard Pod Is Not Getting Restarted
```
jgarciao, line 191: The test in Polarion have these steps:
1. Look odh Dashboard Pod
2. Check the odh dashboard is not restarted by the internal pilot by the memory limit 

Right now, how the test is implemented, I see it more as a post-install check (it could be renamed to `Verify Dashboard Pod Is Not Getting Restarted During Installation`)

@pablofelix @bdattoma , I think the test in polarion is not very clear. What do you mean with the internal pilot? Is this meant to be a Sandbox test?

tests/Tests/400__ods_dashboard/400__ods_dashboard.robot
```diff
@@ -188,6 +188,13 @@ Verify Favorite Resource Cards
     Favorite Items Should Be Listed First When Sorted By    ${favorite_ids}    duration
     [Teardown]    Remove Items From favorites    @{favorite_ids}
 
+Verify Dashboard Pod Is Not Getting Restarted	
+    [Documentation]    Verify Dashboard Pod container restart
+    [Tags]    Sanity
+    ...       ODS-374
+    ${pod_names}    Get POD Names    redhat-ods-applications    ${LABEL_SELECTOR}
```
jgarciao, line 195: `${LABEL_SELECTOR}` in my opinion is not necessary. If you want to keep it, it could be renamed to` ${DASHBOARD_LABEL_SELECTOR}`

jgarciao (changes requested):

tests/Tests/400__ods_dashboard/400__ods_dashboard.robot
bdattoma: the test should check that the dashboard pods don't have a Out-of-Memory error, because this is why the TC was created. 
Already talked yday with @rkamble5 and he will try to reproduce the OOM error on a cluster to refine this implementation

PR #421, opened 2022-06-17
Small changes in Sanity

bdattoma:
this PR wants to bring 3 changes in sanity:
1. `Verify License Of Disabled Cards Can Be Re-validated` is a bit flaky, sometimes it fails while enabling anaconda (timing issue). I've increase a timeout
2. `Verify License Of Disabled Cards Can Be Re-validated` I've added the "FlakyTest" tag because of point 1)
3. `Verify Culler Kills Inactive Server` . I've added the "ProductBug" tag, because otherwise it wasn't possible to filter out from runs. The test used to add the tag as outcome of "Fail" keyword, but it doesn't allow to filter out the test case before running it


lugi0 (approved):

tarukumar (approved):
LGTM

PR #425, opened 2022-06-21
Add test Verify That The Operator Pod Does Not Get Stuck After Upgrade

pablofelix:

### DIFF
--- a/tests/Tests/100__deploy/100__installation/109__operator.robot
+++ b/tests/Tests/100__deploy/100__installation/109__operator.robot
@@ -24,6 +24,19 @@ Verify Odh-deployer Checks Cluster Platform Type
     ${odhdeployer_logs}=    Fetch Odh-deployer Pod Logs
     Should Contain    ${odhdeployer_logs}    ${odhdeployer_logs_content}
 
+Verify That The Operator Pod Does Not Get Stuck After Upgrade
+    [Documentation]    Verifies that the operator pod doesn't get stuck 
+    [Tags]    Sanity
+    ...       Tier1
+    ...       ODS-818
+    ${operator_pod_info}=    Fetch operator Pod Info
+    ${length}=    Get length    ${operator_pod_info}
+    IF    ${length} == 2
+            ${err_image_pull}=    Verify Operator Pods Has ErrImagePull After upgrade    ${operator_pod_info}
+            IF   ${err_image_pull}
+               Fail   operator stuck
+            END
+    END
 
 *** Keywords ***
 Fetch Odh-deployer Pod Info
@@ -48,3 +61,27 @@ Fetch Odh-deployer Pod Logs
     ...                         namespace=redhat-ods-operator
     ...                         container=rhods-deployer
     [Return]    ${odhdeployer_pod_Logs}
+
+Fetch Operator Pod Info
+    [Documentation]  Fetches information about operator pod
+    ...    Args:
+    ...        None
+    ...    Returns:
+    ...        operator_pod_info(dict): Dictionary containing the information of the operator pod
+    @{operator_pod_info}=    Oc Get    kind=Pod    api_version=v1    label_selector=name=rhods-operator
+    [Return]    @{operator_pod_info}
+
+Verify Operator Pods Has ErrImagePull After upgrade
+        [Documentation]  Fetches information about operator pod
+        ...    Args:
+        ...        operator_pod_info(dict): Dictionary containing the information of the operator pod 
+        ...    Returns:
+        ...        err_image_pull(bool): True when the error ErrImagePull is present
+        [Arguments]    ${operator_pod_info}
+        ${err_image_pull}=    Run Keyword And Return Status   
+        ...    wait until keyword succeeds  60 seconds  1 seconds
+        ...    OpenShift Resource Field Value Should Be Equal As Strings    
+        ...    status.containerStatuses[0].state.waiting.reason
+        ...    ErrImagePull
+        ...    @{operator_pod_info}
+        [Return]    ${err_image_pull}
--- a/tests/Tests/100__deploy/100__installation/109__operator.robot
+++ b/tests/Tests/100__deploy/100__installation/109__operator.robot
@@ -25,17 +25,17 @@ Verify Odh-deployer Checks Cluster Platform Type
     Should Contain    ${odhdeployer_logs}    ${odhdeployer_logs_content}
 
 Verify That The Operator Pod Does Not Get Stuck After Upgrade
-    [Documentation]    Verifies that the operator pod doesn't get stuck 
+    [Documentation]    Verifies that the operator pod doesn't get stuck after an upgrade
     [Tags]    Sanity
     ...       Tier1
     ...       ODS-818
     ${operator_pod_info}=    Fetch operator Pod Info
     ${length}=    Get length    ${operator_pod_info}
-    IF    ${length} == 2
-            ${err_image_pull}=    Verify Operator Pods Has ErrImagePull After upgrade    ${operator_pod_info}
-            IF   ${err_image_pull}
-               Fail   operator stuck
-            END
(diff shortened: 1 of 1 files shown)


pablofelix:
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1049/

tests/Tests/100__deploy/100__installation/109__operator.robot
```diff
@@ -24,6 +24,19 @@ Verify Odh-deployer Checks Cluster Platform Type
...
 
+Verify That The Operator Pod Does Not Get Stuck After Upgrade
+    [Documentation]    Verifies that the operator pod doesn't get stuck 
+    [Tags]    Sanity
+    ...       Tier1
+    ...       ODS-818
+    ${operator_pod_info}=    Fetch operator Pod Info
+    ${length}=    Get length    ${operator_pod_info}
+    IF    ${length} == 2
+            ${err_image_pull}=    Verify Operator Pods Has ErrImagePull After upgrade    ${operator_pod_info}
```
bdattoma, line 35: if length is 2, then it means there are two instances of the operator pod running at the same time, right? If yes, then should line 35 be run for both the pods?

tests/Tests/100__deploy/100__installation/109__operator.robot
```diff
@@ -24,6 +24,19 @@ Verify Odh-deployer Checks Cluster Platform Type
...
+            IF   ${err_image_pull}
+               Fail   operator stuck
+            END
```
bdattoma, line 38: hm very specific error. What if we just chek the status if "error" (or whatever is the appropriate value)?

tests/Tests/100__deploy/100__installation/109__operator.robot
pablofelix: The keyword `Verify Operator Pods Has ErrImagePull After upgrade `  contains `OpenShift Resource Field Value Should Be Equal As Strings` that run in both pods and look for the error in each pod.

tests/Tests/100__deploy/100__installation/109__operator.robot
lugi0: @bdattoma what do you mean? I didn't really get your point here

tests/Tests/100__deploy/100__installation/109__operator.robot
```diff
@@ -24,6 +24,19 @@ Verify Odh-deployer Checks Cluster Platform Type
     ${odhdeployer_logs}=    Fetch Odh-deployer Pod Logs
     Should Contain    ${odhdeployer_logs}    ${odhdeployer_logs_content}
 
+Verify That The Operator Pod Does Not Get Stuck After Upgrade
+    [Documentation]    Verifies that the operator pod doesn't get stuck
```
lugi0, line 28: I'm just wondering about the timing of this test. Would this run during or after the upgrade?

tests/Tests/100__deploy/100__installation/109__operator.robot
bdattoma: right, sorry stupid question, i missed something xD thx

tests/Tests/100__deploy/100__installation/109__operator.robot
pablofelix: The scope of this test is to verify that the operator pod is not stuck after the upgrade. In that case there was an specific error the pod was stuck and when the old pod was scaled to 0 the new pod with the new image run correctly.

bdattoma: @lugi0 the "fail" kw will be triggered only if the pod get "ErrImagePull" error, but I think the container may fail for different reasons

tests/Tests/100__deploy/100__installation/109__operator.robot
pablofelix: During the upgrade, the test will work, but we need to know if a pod is stuck after the upgrade. So it's better after the upgrade.

tests/Tests/100__deploy/100__installation/109__operator.robot
lugi0: So are you suggesting this should fail for any error? Or that the error message should be more specific? @bdattoma 
I think Pablo is testing a specific failure here, but I probably agree it would make more sense to have this fail for any error in the operator pod

bdattoma: @lugi0 yes I'm suggesting to fail for any error, because if there's an error in the pod during the deployment the operator would still be stuck, I guess. 

And I would explicitly log the error msg in case of failure.

lugi0: @pablofelix do you think this can be generalized to work on any error?

pablofelix: For me, make it fail for any errors would be a different test. We could create that test, but it’s not trivial at all, there are a lot of events to take care of, so many, that monitoring tools exist for that purpose, and it couldn’t be done just by checking for error and logging the message, as happens in other applications. Check for all errors would take a lot of effort and time, and I’m not even sure if it would be feasible for us to do it. I think that we could be able to check some errors, but not all of them, and on the other hand, this test will fail for a lot for things like network failures that resolve by themselves and are not real bugs. I’m not sure if creating that kind of test makes sense.


In my opinion, this test case was created to automate RHODS-2078, and it should test the particular error that made the pod get stuck.

In the issue, the CrashLoopBackOff error is reported, but this error just means that the pod is in a crash loop, but an underlying reason must exist. I assumed, by the context, that it was ErrImagePull. Thanks to your comments, I've realized that I should reproduce the error to check if that was the real reason for sure, but doing it now is more difficult. So if we find a CrashLoopBackOff error, I think that it might be good if we could investigate a bit further to get the underlying reason and add that info to the jira issue when we file it, so we can know for sure the error to create the automation for.

bdattoma: > but an underlying reason must exist. I assumed, by the context, that it was ErrImagePull


I don't think it was ErrImagePull because the events report "Successfuly pulled image ...". However, I think you could check the container status, and if it is not in a "ready" state, then the TC can fail printing out the error. 


> In my opinion, this test case was created to automate RHODS-2078, and it should test the particular error that made the pod get stuck.


Anyway, this makes sense too, but I think the "ErrImagePull" is not necesserely the error to check for. Maybe, we could just rely on the first check (i.e., the number of operator pods returned) to check this particular case. And then the TC could print out the error message whatever it is

pablofelix: Yes, ErrImagePull might not be the error to check then, I need to reproduce the specific error that left the pod stuck. Yesterday I didn’t succeed to reproduce it. The issue was open some time ago and the image was failing to pull, I guess because of secrets. Today I will try other ways to see if I’m lucky.

If I’m not able to reproduce it, then I can do the first check as you say, and what I could also do is to wait for the CrashLoopBackOff error to happen and log the error why the container was exited. Once we hit the issue again, if we do, we can debug the CrashLoopBackOff error to see if we can find the underlying reason that led to the crash loop and modify the test.

Verify that containers status is ready (also failing pods) is a verification that, in my opinion, should be done for all pods after the installation or upgrade. In the implementation I did  for Robot Framework this verification exists, and perhaps it would be good that we could migrate from the script to use them. However, just because a container is not in ready status doesn’t mean necessarily that it is failing, and the test will be failing for a lot for reasons that are not real failures.

pablofelix: Done

bdattoma (approved):

tarukumar (approved):
lgtm

PR #428, opened 2022-06-22
Add script to split Suites

bdattoma:
This PR wants to implement a script to split long running suites in multiple portion. It helps with MSP onboarding of our Sanity gate.

Moreover, it moves the suite `culler` out from `500__jupyterhub` to `800__culler` to balance the suite split, since the tests in `culler` can last for much time (50 minutes approx, 30 minutes if we remove tests marked as `ProductBug`)

As last thing, it mark `Verify Spwaned Notebook Size` as `FlakyTest` and   `Verify "Enabled" Keeps Being Available After One Of The ISV Operators If Uninstalled` as `ProductBug`


utils/scripts/SplitSuite.py
```diff
@@ -0,0 +1,86 @@
...
+def fetch_subsuites(parent):
+    children = [child for child in parent.suites]
+    return children
+
+
+def fetch_subsuites_from_list(parents_list):
+    children = []
+    for parent in parents_list:
+        children += fetch_subsuites(parent)
+    return children
```
lugi0, line 26: this doesn't seem to be used

bdattoma: that's true, removing it

lugi0 (approved):
lgtm

jgarciao (approved):

PR #439, opened 2022-06-30
Update example file for test-variables.yml

bdattoma:


jgarciao (changes requested):
Could you also update this line?
```
# RHODS_PROMETHEUS_TOKEN can be obtained running: oc serviceaccounts new-token prometheus -n redhat-ods-monitoring #
```

PR #443, opened 2022-07-01
Add test to check if Intel images are pinned to a numeric version

bdattoma:
This PR covers ODS-1085 and ODS-1247.
The check on image tag cannot be 100% reliable because every ISV can use different versioning formats. So, the code here is first checking if the image tag contains generic strings like "latest" (fail if true). If the first check is false, then it looks for numbers in the image tag. If numbers are there the test is passed. However, it checks if the numeric tag follow the usual versioning formats like x.y.z-n or x.y-n or x.y.

Signed-off-by: bdattoma [[bdattoma@redhat.com](mailto:bdattoma@redhat.com)]

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
@@ -490,3 +490,28 @@ Get Current User
    ${url} =  Get Location
    ${current_user} =  Evaluate  '${url}'.split("/")[4]
    [Return]  ${current_user}
+
+Image Should Be Pinned To A Numeric Version
+    [Documentation]     Verifies if the Image Tag is (probably) pinned to a specific image version (e.g., 2.5.0-8).
+    ...                 Since each image provider could use different versioning format (e.g., x.y.z, 10May2021, etc),
+    ...                 the check may not be always reliable. The logic is that if numbers are present in the tag, the image
+    ...                 is likely pinned to a specific version rather than a generic tag (e.g., latest). After that it tries to
+    ...                 see if the numbers follow usual versioning pattern (raising a Warning in the negative case).
+    ${image_spec}=   Run Cell And Get Output  import os; os.environ['JUPYTER_IMAGE_SPEC']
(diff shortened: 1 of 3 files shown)


bdattoma:
[Jenkins Job](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1070/TestReport/)
Test on Aikit failed as expected (the image tag doesn't refer to a numeric version)
Test on Openvino passed (the image tag is something like <year>_<month>_<day> where <year> and <day> are numeric values)

tarukumar (approved):
LGTM

lugi0 (approved):
the PR looks good but I don't really understand the need for this check.
If a specific versioning is enforced for all images then it makes sense to have the check on that specific format. Having this check that may or may not fail, and even when it does or doesn't we don't know if it's correct or not, seems useless to me.

bdattoma:
> if a specific versioning is enforced for all images then it makes sense to have the check on that specific format.


there is not an enforced versioning format for the ISVs, that's why the check is kept generic. It raises the flag when an ISV uses a versioning system which looks to be labeled/generic (e.g., "latest", "main", non-numeric). However, we can see it in the next auto meeting to decide if we want to keep it or not

lugi0:
Yes, I understand why you did it the way you did, but unless a specific format is enforced IMHO we should not be doing this check.
If an ISV follows a different format and this checks flags it as incorrect, will they fix it or will it keep being flagged by the test? My understanding is that they are free to use whatever format they prefer, meaning we would have tests (e.g. the AiKit one) constantly failing.

bdattoma:
> Yes, I understand why you did it the way you did, but unless a specific format is enforced IMHO we should not be doing this check. If an ISV follows a different format and this checks flags it as incorrect, will they fix it or will it keep being flagged by the test? My understanding is that they are free to use whatever format they prefer, meaning we would have tests (e.g. the AiKit one) constantly failing.

the only request done to ISVs was to not use static label for the versioning. there are a bunch of bugs open for fixing this, and at least one of them is fixed

tarukumar (approved):

lugi0 (approved):

PR #452, opened 2022-07-11
Wait until cuda builds are Completed in deployment test

jgarciao:
Signed-off-by: Jorge Garcia Oncins <jgarciao@redhat.com>

### DIFF
--- a/tests/Resources/ODS.robot
+++ b/tests/Resources/ODS.robot
@@ -400,3 +400,30 @@ Delete RHODS Config Map
     [Arguments]     ${name}  ${namespace}=redhat-ods-applications
     OpenShiftLibrary.Oc Delete    kind=ConfigMap  name=${name}  namespace=${namespace}
 
+Wait Until Cuda Builds Are Completed
+    [Documentation]    Waits until pytorch and tensorflow builds have Complete Status.
+    ...    The keyword waits 20 minutes for each build to be in that status, failing otherwise.
+
+    ${pytorch_build_name}=    Search Last Build    namespace=redhat-ods-applications
+    ...    build_name_includes=pytorch
+    Wait Until Build Status Is    namespace=redhat-ods-applications
+    ...    build_name=${pytorch_build_name}    expected_status=Complete
+
+    ${tensorflow_build_name}=    Search Last Build    namespace=redhat-ods-applications
+        ...    build_name_includes=tensorflow
+    Wait Until Build Status Is    namespace=redhat-ods-applications
+    ...    build_name=${tensorflow_build_name}    expected_status=Complete
+
+Verify Cuda Builds Are Completed
+    [Documentation]    Verify All Cuda Builds have status as Complete
+    ${Pods} =    Run    oc get build -n redhat-ods-applications
+    @{builds} =    Split String    ${Pods}    \n
+    ${len} =    Get Length    ${builds}
+    FOR    ${ind}    IN RANGE    1    ${len}
+        @{pre} =    Split String    ${builds}[${ind}]
+        ${is_cuda_build} =   Run Keyword And Return Status   Should Contain    ${pre}[0]    cuda
+        IF    ${is_cuda_build} == True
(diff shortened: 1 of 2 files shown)


tests/Resources/ODS.robot
```diff
@@ -400,3 +400,30 @@ Delete RHODS Config Map
...
+    ...    build_name=${pytorch_build_name}    expected_status=Complete
+
+    ${tensorflow_build_name}=    Search Last Build    namespace=redhat-ods-applications
+        ...    build_name_includes=tensorflow
+    Wait Until Build Status Is    namespace=redhat-ods-applications
+    ...    build_name=${tensorflow_build_name}    expected_status=Complete
+
+Verify Cuda Builds Are Completed
+    [Documentation]    Verify All Cuda Builds have status as Complete
+    ${Pods} =    Run    oc get build -n redhat-ods-applications
```
pablofelix, line 419: Could you change the run oc for the openshift-library  `Oc get  kind=Build   namespace=redhat-ods-applications`

pablofelix (dismissed):

tests/Resources/ODS.robot
```diff
@@ -400,3 +400,30 @@ Delete RHODS Config Map
     [Arguments]     ${name}  ${namespace}=redhat-ods-applications
     OpenShiftLibrary.Oc Delete    kind=ConfigMap  name=${name}  namespace=${namespace}
 
+Wait Until Cuda Builds Are Completed
+    [Documentation]    Waits until pytorch and tensorflow builds have Complete Status.
+    ...    The keyword waits 20 minutes for each build to be in that status, failing otherwise.
```
lugi0, line 405: Is 20 minutes enough?
What about all the cuda builds before pytorch/tensorflow? I think this should wait until either build at least starts, and then eventually wait for 20 minutes until the build is completed.
Otherwise if we run this TC right after install it will fail.

bdattoma: +1 it should wait for all the cuda builds to be finished, including the base ones before notebook images.

if 20 minutes is enough? I guess it is enough because the robot wait 20 minutes for each build, which take max 10 minutes, if I'm not wrong

tests/Resources/ODS.robot
jgarciao: I've deleted this keyword and instead I've added "Verify All Builds Are Complete" in Builds.robot. This one is using the openshift-library

tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -123,10 +124,12 @@ Verify Oath-Proxy Image Is fetched From CPaaS
 Verify Pytorch And Tensorflow Can Be Spawned
     [Documentation]    Check Cuda builds are complete and  Verify Pytorch and Tensorflow can be spawned
     [Tags]    Sanity
+    ...       Tier1
     ...       ODS-480  ODS-481
-    Verify Cuda Builds Are Completed
-    Verify Image Can Be Spawned  image=pytorch  size=Default
-    Verify Image Can Be Spawned  image=tensorflow  size=Default
+    Wait Until All Builds Are Complete    namespace=redhat-ods-applications
+    Verify All Builds Are Complete    namespace=redhat-ods-applications
```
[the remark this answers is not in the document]
jgarciao: Yes, it's really not needed. I'll remove it

lugi0 (approved):
lgtm

tests/Resources/ODS.robot
jgarciao: In the end I've removed the call to "Verify All Builds Are Complete" because it was not needed after "Wait Until All Builds Are Complete"

tests/Resources/ODS.robot
jgarciao: In the end the test is using "Wait Until All Builds Are Complete", which will wait 20 minutes for each build. As @bdattoma  said, they all take aprox 10 minutes, so it should be enough

PR #454, opened 2022-07-13
POC: Remove the dependency on `oc` being connected to the cluster

kpouget:
The `OpenShiftCLI` library requires a valid `KUBECONFIG` to load
properly. Without it, the framework initialization fails because it
cannot access the cluster.

As part of our scale testing, we cannot allow any unnecessary call to
OpenShift, be it `oc login` or resource queries.

This commit is a quick-and-dirty POC showing that we can run our OSD-CI-based scale
test without relying on `OpenShiftCLI`.


lugi0:
@kpouget I'm implementing this in #588

kpouget:
nice, thanks for the update

PR #455, opened 2022-07-14
moving PVC size verification TCs out of smoke

tarukumar:
It is suitable to move back end pvc size verification to Tier2 and UI to sanity. Removing both the TC(ODS-1220,1221) out of smoke.

### DIFF
--- a/tests/Tests/400__ods_dashboard/411__ods_dashboard_settings_pvc_size.robot
+++ b/tests/Tests/400__ods_dashboard/411__ods_dashboard_settings_pvc_size.robot
@@ -30,9 +30,7 @@ ${SIZE_CODE}    import subprocess;
 Verify User Can Spawn Notebook After Changing PVC Size Using Backend
     [Documentation]   Verify if user can spawn notebook
     ...    for supported PVC size got changed
-    [Tags]    Smoke
-    ...       Sanity
-    ...       Tier1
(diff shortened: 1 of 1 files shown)


tests/Tests/400__ods_dashboard/411__ods_dashboard_settings_pvc_size.robot
```diff
@@ -30,9 +30,7 @@ ${SIZE_CODE}    import subprocess;
 Verify User Can Spawn Notebook After Changing PVC Size Using Backend
     [Documentation]   Verify if user can spawn notebook
```
jgarciao, line 31: Could you add the id of the product bug in [Documentation] ?
`...    ProductBug: RHODS-XXXX`

jgarciao (changes requested):

jgarciao (approved):

tests/Tests/400__ods_dashboard/411__ods_dashboard_settings_pvc_size.robot
tarukumar: @jgarciao There is no product bug for this TC. I think you wanted me to add product bug to the ODS-1229. I have added documentation product bug in the documentation for that.

mwaykole (approved):

PR #457, opened 2022-07-15
Spawn Notebook With Arguments: fail if the retry count is exhausted

kpouget:
Without this patch, the keyword would try `${retries}` times and
return happily, even if none of the tries succeeded.

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -399,12 +399,12 @@ Handle Bad Gateway Page
 
 Verify Image Can Be Spawned
     [Documentation]    Verifies that an image with given arguments can be spawned
-    [Arguments]    ${retries}=1    ${image}=s2i-generic-data-science-notebook    ${size}=Small    ${spawner_timeout}=600 seconds
-    ...    ${gpus}=0    ${refresh}=${False}    &{envs}
+    [Arguments]    ${retries}=1    ${retries_delay}=0 seconds    ${image}=s2i-generic-data-science-notebook    ${size}=Small
+    ...    ${spawner_timeout}=600 seconds    ${gpus}=0    ${refresh}=${False}    &{envs}
     Begin Web Test
     Launch JupyterHub Spawner From Dashboard
-    Spawn Notebook With Arguments    ${retries}    ${image}    ${size}
-    ...    ${spawner_timeout}    ${gpus}    ${refresh}    &{envs}
+    Spawn Notebook With Arguments    retries=${retries}   retries_delay=${retries_delay}    image=${image}    size=${size}
+    ...    spawner_timeout=${spawner_timeout}    gpus=${gpus}    refresh=${refresh}    envs=&{envs}
     End Web Test
 
 Verify Library Version Is Greater Than
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -134,10 +134,10 @@ Has Spawn Failed
 
 Spawn Notebook With Arguments  # robocop: disable
    [Documentation]  Selects required settings and spawns a notebook pod. If it fails due to timeout or other issue
-   ...              It will try again ${retries} times (Default: 1). Environment variables can be passed in as kwargs
-   ...              By creating a dictionary beforehand
+   ...              It will try again ${retries} times (Default: 1) after ${retries_delay} delay (Default: 0 seconds).
+   ...              Environment variables can be passed in as kwargs by creating a dictionary beforehand
    ...              e.g. &{test-dict}  Create Dictionary  name=robot  password=secret
-   [Arguments]  ${retries}=1  ${image}=s2i-generic-data-science-notebook  ${size}=Small
+   [Arguments]  ${retries}=1  ${retries_delay}=0 seconds  ${image}=s2i-generic-data-science-notebook  ${size}=Small
    ...    ${spawner_timeout}=600 seconds  ${gpus}=0  ${refresh}=${False}  &{envs}
    FOR  ${index}  IN RANGE  0  1+${retries}
       ${spawner_ready} =    Run Keyword And Return Status    Wait Until JupyterHub Spawner Is Ready
@@ -173,8 +173,12 @@ Spawn Notebook With Arguments  # robocop: disable
          Exit For Loop If  ${spawn_fail} == False
          Click Element  xpath://span[@id='jupyterhub-logo']
       ELSE
+         Sleep  ${retries_delay}
          Click Element  xpath://span[@id='jupyterhub-logo']
       END
+   END
+   IF  ${spawn_fail} == True
+      Fail  msg= Spawner failed loading after ${retries} retries
    END
 
 Spawned Image Check


lugi0 (changes requested):
Thanks @tarukumar, I was in the process of sending a similar PR. I would change this up slightly, let me know if you agree.

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -173,7 +173,9 @@ Spawn Notebook With Arguments  # robocop: disable
          Exit For Loop If  ${spawn_fail} == False
          Click Element  xpath://span[@id='jupyterhub-logo']
       ELSE
+         Run Keyword If  ${index} == ${retries}  Fail
          Click Element  xpath://span[@id='jupyterhub-logo']
+         Sleep  ${retries_delay}
```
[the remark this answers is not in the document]
tarukumar: Sure, Your proposal looks good to me. Thanks @lugi0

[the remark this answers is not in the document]
lugi0: I just realized the PR was sent by @kpouget and @tarukumar only added the reviewers. Sorry and thanks for the contribution @kpouget :D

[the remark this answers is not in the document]
kpouget: >  I think the expectation is that the spawner would load everything rather quickly regardless of user count, and adding an additional artificial wait time would hide the issue that these FAILs would show.

the reason why we want this delay is to have a chance that the test succeeds, even if it takes time.
Currently we're at 80% of failures (out of 100 users) because of this bottleneck, we need to get passed this quickly.
The delay can be 0 by default to keep the existing behavior.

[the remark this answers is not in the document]
kpouget: they will be executed if `${index} != ${retries}`, no?
only when `${index} == ${retries}` this will be skipped (no need to reload, we tried too many times)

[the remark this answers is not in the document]
tarukumar: yes it will execute. I was thinking if it is fail also reload and wait. However i think above proposal of luca outside the loop is good.

kpouget:
updated with the `IF` after the test, and `retries_delay` of `0 seconds` by default

lugi0 (approved):
I think this is good enough, I like the idea of keeping the delay at 0 by default but configurable if needed.
@kpouget Do you have the ability to run a PR test on our Jenkins? If so please test these changes (I would do a full Sanity run).
Otherwise please let me know and I'll run the test for you

kpouget:
@lugi0 no I don't think I can trigger the test on Jenkins, if you can please kick it (and let me know how to follow the progress)

lugi0:
@kpouget you should be able to follow the job here: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1105/console
Let me know if you can access that page, we might need to give you some additional permissions

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -174,6 +174,10 @@ Spawn Notebook With Arguments  # robocop: disable
          Click Element  xpath://span[@id='jupyterhub-logo']
       ELSE
          Click Element  xpath://span[@id='jupyterhub-logo']
+         Sleep  ${retries_delay}
+      END
+      IF ${spawn_fail} == True
+         Fail  msg= Spawner failed loading after ${retries} retries
       END
```
lugi0, line 181: This whole block should go after the last END (right below, line 182)

lugi0 (changes requested):
@kpouget Sorry, had to stop the test, I just saw that the IF block is placed incorrectly. It should go outside the first FOR loop, so after the last END currently at line 182

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
kpouget: good catch, fixed

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -174,8 +174,12 @@ Spawn Notebook With Arguments  # robocop: disable
          Click Element  xpath://span[@id='jupyterhub-logo']
       ELSE
          Click Element  xpath://span[@id='jupyterhub-logo']
+         Sleep  ${retries_delay}
       END
    END
+   IF ${spawn_fail} == True
```
lugi0, line 180: There's only one whitespace between the IF and ${spawn_fail}, so the test are currently failing:
```
1) No keyword with name 'IF ${spawn_fail} == True' found.
```

@kpouget please update it to be two whitespaces or more, `IF  ${spawn_fail} == True`

lugi0 (changes requested):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
kpouget: fixed (the `robot` language is hard 🙃)

lugi0: I would say annoying more than hard :p

lugi0:
@kpouget Good catch, sorry for not noticing before :( I'll send another run today

lugi0:
test job: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1113/console

lugi0:
Run finished with 8 failures, all seem expected failures with the exception of one, which however seems an automation bug on our end. 
`Verify Pytorch And Tensorflow Can Be Spawned` failed because it tried to select an image with name `Default` (instead of `pytorch`). @jgarciao if I remember correctly you added this test, can you take a look?

Other than this, I think this is verified and good to go.

kpouget:
thanks, I tested it on my end and I'm happy with the PR as well.

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -134,10 +134,10 @@ Has Spawn Failed
 
 Spawn Notebook With Arguments  # robocop: disable
    [Documentation]  Selects required settings and spawns a notebook pod. If it fails due to timeout or other issue
-   ...              It will try again ${retries} times (Default: 1). Environment variables can be passed in as kwargs
-   ...              By creating a dictionary beforehand
+   ...              It will try again ${retries} times (Default: 1) after ${retries_delay} delay (Default: 0 seconds).
+   ...              Environment variables can be passed in as kwargs by creating a dictionary beforehand
    ...              e.g. &{test-dict}  Create Dictionary  name=robot  password=secret
-   [Arguments]  ${retries}=1  ${image}=s2i-generic-data-science-notebook  ${size}=Small
+   [Arguments]  ${retries}=1  ${retries_delay}=0 seconds  ${image}=s2i-generic-data-science-notebook  ${size}=Small
```
jgarciao, line 140: _Verify Pytorch And Tensorflow Can Be Spawned_ failed because it is calling the keyword `Verify Image Can Be Spawned` (defined in tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot), which is calling  `Spawn Notebook With Arguments` using positional arguments instead of named arguments

I see two possible solutions:
- Define the new ${retries_delay} argument at the end, so it doesn't break existing code
- Review and fix all calls to `Spawn Notebook With Arguments` not to have this problem

To me it's fine if you do the fist one

jgarciao (changes requested):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
kpouget: I had a quick at the locations where `Spawn Notebook With Arguments` is called, and I could only find one place (`Verify Image Can Be Spawned`) where the keyword is called with positional arguments, so I'll change it there

kpouget: should `Verify Image Can Be Spawned` accept the `retries_delay` ? 🤔

jgarciao: >I had a quick at the locations where Spawn Notebook With Arguments is called, and I could only find one place (Verify Image Can Be Spawned) where the keyword is called with positional arguments, so I'll change it there
I've searched for it too and also found it just in one place

>should Verify Image Can Be Spawned accept the retries_delay ?
ok for me. I think that keyword is only used in `Verify Pytorch And Tensorflow Can Be Spawned`

kpouget: ok for you ... to accept it? or not to accept it (current code in https://github.com/red-hat-data-services/ods-ci/pull/457/commits/f6b0528b4b547fd9d7e05dab7eed3252ec59861b)

lugi0: IMHO it should accept it, even though it doesn't currently use it it could come in handy down the line

kpouget: updated as suggested

jgarciao (approved):

lugi0:
7 failures, same as this morning and all expected. Merging the PR.

lugi0 (approved):

PR #469, opened 2022-07-22
Spawn Notebook With Arguments: initialize ${spawn_fail} variable

kpouget:
Without this initialization, if the spawner never gets ready, the
`${spawn_fail}` variable is never initialized and the final
`IF  ${spawn_fail} == True` crashes.


tarukumar (approved):

lugi0 (commented):
I was sure we already had made this change, weird

lugi0 (approved):

kpouget:
hum, I said I would do it ... but I didn't. Shame on my 🤦🏻‍♂️ 

what is the correct syntax:
```
${spawn_fail} =   Set Variable  True
```
or (this commit)
```
${spawn_fail} =   True
```
?

tarukumar:
It should be ${spawn_fail} =   Set Variable  True

PR #470, opened 2022-07-22
Add tensorboard test for tensorflow/pytorch images

lugi0:

### DIFF
--- a/tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
+++ b/tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
@@ -34,6 +34,19 @@ PyTorch Image Workload Test
     Capture Page Screenshot
     JupyterLab Code Cell Error Output Should Not Be Visible
 
+Tensorboard Workload Test
+    [Documentation]  Verifies that tensorboard is accessible
+    [Tags]  Sanity
+    ...     PLACEHOLDER
+    ...     ODS-1414
+    Close Previous Server
+    # For KFNBC the environment variable should TENSORBOARD_PROXY_URL=/notebook/<nb-namespace>/<nb-name>/proxy/6006/
+    &{tensorboard_proxy} =  Create Dictionary  TENSORBOARD_PROXY_URL=/user/${TEST_USER.USERNAME}/proxy/6006/
+    Spawn Notebook With Arguments  image=${NOTEBOOK_IMAGE}  size=Default  envs=&{tensorboard_proxy}
+    Run Keyword And Ignore Error  Clone Git Repository And Run  https://github.com/redhat-rhods-qe/ods-ci-notebooks-main  
+    ...    ods-ci-notebooks-main/notebooks/500__jupyterhub/tensorboard/pytorch/tensorboard_profiling_pytorch.ipynb  10m
+    Page Should Contain Element    xpath://iframe[contains(@id, "tensorboard-frame")]
+
 Verify PyTorch Image Can Be Spawned With GPU
     [Documentation]    Spawns PyTorch image with 1 GPU
     [Tags]  Sanity
@@ -74,3 +87,10 @@ Verify PyTorch Image Suite Setup
     Begin Web Test
     Launch JupyterHub Spawner From Dashboard
     Spawn Notebook With Arguments  image=${NOTEBOOK_IMAGE}  size=Default
+
+Close Previous Server
+    [Documentation]  Closes previous server and goes back to Spawner
+    Clean Up Server
+    Stop JupyterLab Notebook Server
+    Fix Spawner Status
+    Wait Until JupyterHub Spawner Is Ready
--- a/tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
+++ b/tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
@@ -45,7 +45,9 @@ Verify Tensorboard Is Accessible
     Spawn Notebook With Arguments  image=${NOTEBOOK_IMAGE}  size=Default  envs=&{tensorboard_proxy}
     Run Keyword And Ignore Error  Clone Git Repository And Run  https://github.com/redhat-rhods-qe/ods-ci-notebooks-main  
     ...    ods-ci-notebooks-main/notebooks/500__jupyterhub/tensorboard/pytorch/tensorboard_profiling_pytorch.ipynb  10m
+    Select Frame    xpath://iframe[contains(@id, "tensorboard-frame")]
     Page Should Contain Element    xpath://html//mat-toolbar/span[.="TensorBoard"]
+    Unselect Frame
 
 Verify PyTorch Image Can Be Spawned With GPU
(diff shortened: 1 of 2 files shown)


lugi0:
Test job: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1143/

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
```diff
@@ -34,6 +34,19 @@ PyTorch Image Workload Test
     Capture Page Screenshot
     JupyterLab Code Cell Error Output Should Not Be Visible
 
+Tensorboard Workload Test
```
bdattoma, line 37: why `Workload`? sounds like a load testing

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
```diff
@@ -34,6 +34,19 @@ PyTorch Image Workload Test
...
+    [Tags]  Sanity
+    ...     PLACEHOLDER
+    ...     ODS-1414
+    Close Previous Server
+    # For KFNBC the environment variable should TENSORBOARD_PROXY_URL=/notebook/<nb-namespace>/<nb-name>/proxy/6006/
+    &{tensorboard_proxy} =  Create Dictionary  TENSORBOARD_PROXY_URL=/user/${TEST_USER.USERNAME}/proxy/6006/
+    Spawn Notebook With Arguments  image=${NOTEBOOK_IMAGE}  size=Default  envs=&{tensorboard_proxy}
+    Run Keyword And Ignore Error  Clone Git Repository And Run  https://github.com/redhat-rhods-qe/ods-ci-notebooks-main  
+    ...    ods-ci-notebooks-main/notebooks/500__jupyterhub/tensorboard/pytorch/tensorboard_profiling_pytorch.ipynb  10m
+    Page Should Contain Element    xpath://iframe[contains(@id, "tensorboard-frame")]
```
bdattoma, line 48: should we check that the iframe contains something? a logo, a particular menu for example. Something which can tell us the iframe is not empty

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
```diff
@@ -34,6 +34,19 @@ PyTorch Image Workload Test
     Capture Page Screenshot
     JupyterLab Code Cell Error Output Should Not Be Visible
 
+Tensorboard Workload Test
+    [Documentation]  Verifies that tensorboard is accessible
+    [Tags]  Sanity
+    ...     PLACEHOLDER
```
bdattoma, line 40: can "placeholder" be removed or is it here for a specific reason?

jgarciao (dismissed):
In the past, this test suite spawned a notebook in the Suite Setup and then had 2 or 3 tests doing things in the already spawned notebook. This was a solution so the tests didn't depend on the execution of the previous one and the notebook didn't have to be spawned each time

After that, other tests have been added that need different notebooks, so the first step they do is to close the notebook that was spawned in the Suite Setup. Also, the GPU tests depend on the execution of the previous one.

So, If this PR needs to be merged quickly I'm fine with it (I've approved it), but in my opinion the hole test suite should be refactored to be more reliable and maintainable. 

Possible solutions:
- Split the tests in different test suites. Spawn the notebook in the Suite Setup when it makes sense
- Keep just one test suite and: 
  - Do not spawn a server in the Suite Setup
  - Merge tests that are currently chained (e.g. ODS-1149 & ODS-1150)
  - Make each test responsible of closing its own notebook server (so no need of "Close Previous Server" in the next one)

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
lugi0: In order to show something in Tensorboard you need to run a workload, albeit small :) I don't see a problem with the name

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
lugi0: I'll double check the behaviour when tensorboard doesn't load

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
lugi0: Placeholder is there for the `Category` tags, which I think we've never implemented

lugi0:
@jgarciao 
>This was a solution so the tests didn't depend on the execution of the previous one and the notebook didn't have to be spawned each time
This is actually still the case. The way in which the suite is set up right now, you can run any test or any combination of tests and it won't fail, because all tests already expecting a notebook will find one due to the suite setup, while all others which need to respawn the image first tear down the server that is created by the suite setup.
We probably lose a couple of minutes in the case in which we only want to run a single test which needs to respawn the server, but honestly I don't think it's a big deal, as that is quite unlikely to happen.

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
bdattoma: Maybe a more self-explanatory name would have been sth like `Verify Tensorboard Is Accessible` or similar.

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
lugi0: You're right, even when it doesn't load it still has an iframe with id `tensorboard-frame-...`. I'll update this.

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
lugi0: Done

tests/Tests/500__jupyterhub/minimal-pytorch-test.robot
lugi0: Done

lugi0:
Test job: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1145/console

jgarciao:
@lugi0 Maybe I'm loosing something but I think these tests will have different results depending on the previous execution of  `PyTorch Image Can Be Spawned With GPU`: 
- Verify PyTorch Image Includes Expected CUDA Version
- Verify PyTorch Library Can See GPUs In PyTorch Image
- Verify PyTorch Image GPU Workload

Anyway, it's good to me to merge now, but revisit in the future when we have time

lugi0:
New test job because of failures: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1146/console

lugi0:
Both passed

jgarciao (approved):
lgtm for now, but still think this test suite should be refactored in the future not to have tests dependent of each other. We can discuss it in a dedicated meeting

bdattoma (approved):

PR #478, opened 2022-07-28
Fix special user testing

lugi0:

### DIFF
--- a/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+++ b/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
@@ -69,7 +69,7 @@ Logout From RHODS Dashboard
     # Another option for the logout button
     #${user} =  Get Text  xpath:/html/body/div/div/header/div[2]/div/div[3]/div/button/span[1]
     #Click Element  xpath://span[.="${user}"]/..
-    Click Button  xpath://div[@class="pf-c-page__header-tools"]/div/div[@class="pf-c-page__header-tools-item"][3]//button[@id="toggle-id"]
+    Click Button  xpath:(//button[@id="toggle-id"])[2]
     Wait Until Page Contains Element  xpath://a[.="Log out"]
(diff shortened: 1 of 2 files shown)


mwaykole (approved):

tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -62,6 +62,18 @@ Login To RHODS Dashboard
...
    Run Keyword If  ${authorize_service_account}  Authorize rhods-dashboard service account
 
+Logout From RHODS Dashboard
+    [Documentation]  Logs out from the current user in the RHODS dashboard
+    ...    This will reload the page and show the `Log in with OpenShift` page
+    ...    so you want to use `Login to RHODS Dashboard` after this
+    # Another option for the logout button
+    #${user} =  Get Text  xpath:/html/body/div/div/header/div[2]/div/div[3]/div/button/span[1]
+    #Click Element  xpath://span[.="${user}"]/..
+    Click Button  xpath://div[@class="pf-c-page__header-tools"]/div/div[@class="pf-c-page__header-tools-item"][3]//button[@id="toggle-id"]
```
tarukumar, line 72: Xpath looks way too big. Could we try simplifying it something like below:

`(//button[@id="toggle-id"])[1]`

lugi0: That xpath unfortunately doesn't work, it's the first one I tried

bdattoma (approved):

tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
tarukumar: oh, i see it would might have not worked because of indexing, below one should work

`(//button[@id="toggle-id"])[2]`

tarukumar (approved):

PR #479, opened 2022-07-29
change container size to smal for sanity run

tarukumar:

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
@@ -112,7 +112,8 @@ Stop JupyterLab Notebook Server
   #   Capture Page Screenshot
   # END
   SeleniumLibrary.Switch Window  Red Hat OpenShift Data Science
-  Click Element  xpath://button[@aria-label="Close"]
+  #Click Element  xpath://button[@aria-label="Close"]
+  SeleniumLibrary.Wait Until Page Contains     Stop notebook server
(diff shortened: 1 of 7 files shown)


tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
```diff
@@ -112,7 +112,8 @@ Stop JupyterLab Notebook Server
   #   Capture Page Screenshot
   # END
   SeleniumLibrary.Switch Window  Red Hat OpenShift Data Science
-  Click Element  xpath://button[@aria-label="Close"]
+  #Click Element  xpath://button[@aria-label="Close"]
+  SeleniumLibrary.Wait Until Page Contains     Stop notebook server
```
lugi0, line 116: Clicking on the button is sometimes needed (maybe not always). iirc the execution does not fail completely when the button is not there, it just moves on to the next button (`Stop notebook server`).
I think it would be better to have an if/else block, or try/catch, or a `Run keyword and ignore error` to still click the button in the cases in which it is needed

lugi0 (changes requested):

tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
tarukumar: Sure i have made the changes. I was verifying in the morning i didn't find anywhere the close locator that the reason i commented out

lugi0 (approved):

PR #499, opened 2022-08-16
[FIX] JupyterLab-git_Version change from  v0.30 to  v0.32

mwaykole:

### DIFF
--- a/tests/Tests/500__jupyterhub/test-versions.robot
+++ b/tests/Tests/500__jupyterhub/test-versions.robot
@@ -18,7 +18,7 @@ Force Tags          JupyterHub
 &{package_versions}      # robocop: disable
 ${JupyterLab_Version}         v3.2
 ${Notebook_Version}           v6.4
-${JupyterLab-git_Version}     v0.30
+${JupyterLab-git_Version}     v0.32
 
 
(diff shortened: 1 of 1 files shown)


aloganat (approved):
lgtm

tests/Tests/500__jupyterhub/test-versions.robot
```diff
@@ -18,7 +18,7 @@ Force Tags          JupyterHub
 &{package_versions}      # robocop: disable
 ${JupyterLab_Version}         v3.2
 ${Notebook_Version}           v6.4
-${JupyterLab-git_Version}     v0.30
+${JupyterLab-git_Version}     v0.32
```
jgarciao, line 21: In order to support RHODS 1.14 and RHODS 1.15 (used during live testing) you can define a variable with the old value (e.g. `${JupyterLab-git_Version-legacy})` and use it with` Is RHODS Version Greater Or Equal Than`

jgarciao (changes requested):

lugi0:
@mwaykole please check if in 1.14 the img getting pulled also includes the updated plugin. If it does I think this PR can be merged as is, otherwise I agree with Jorge

PR #501, opened 2022-08-18
Verify cuda buildchain in smoke

jgarciao:
- Fix keyword `Wait Until All Builds Are Complete` to use buildconfigs instead of buils
- Add "Verify That CUDA Build Chain Succeeds" to smoke. 
  - Previously this test was called "Verify Pytorch And Tensorflow Can Be Spawned" but that test doesn't exist in Polarion anymore(same for polarion id's ODS-480 and ODS-481)


jgarciao:
Closed (I'll send it again after rebasing)

PR #508, opened 2022-08-24
Fix test-versions suite

lugi0:

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -257,11 +257,13 @@ Handle Server Is Stopping
 
 Control Panel Is Visible
    [Documentation]  Checks if Control Panel page is open
+   Sleep  2s
    ${control_panel_visible} =  Run Keyword And Return Status  Page Should Contain  Notebook server control panel
    [Return]  ${control_panel_visible}
 
 Handle Control Panel
    [Documentation]  Handles control panel page
+   Wait Until Page Contains     Stop notebook server    timeout=10s
    Click Button  Stop notebook server
    Wait Until Page Contains Element  xpath://button[.="Stop server"]
    Click Button  xpath://button[.="Stop server"]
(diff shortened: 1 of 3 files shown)


lugi0:
PR test: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1177/console

Expecting a few failures due to lib versions

lugi0:
Tests finished, only expected failures

mwaykole (approved):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -257,11 +257,13 @@ Handle Server Is Stopping
 
 Control Panel Is Visible
    [Documentation]  Checks if Control Panel page is open
+   Sleep  2s
```
jgarciao, line 260: Instead of using this Sleep, how about using Wait Until Page Contains in the next statement?

`${control_panel_visible} =  Run Keyword And Return Status     Wait Until Page Contains  text=Notebook server control panel  timeout=5s`

lugi0: `Wait until page contains` will add overhead every time `Fix Spawner Status` is called to check if the Control Panel is visible. I don't think it's worth it to add running time to almost all test cases just to replace this Sleep, which is just there to make sure the page has fully loaded before doing its check

jgarciao: I don't see why it should add an appreciable overhead over Sleep (on the contrary, it could finish faster if the text is available before 2s) but I it's not a big deal (I've approved the PR)

jgarciao (approved):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
lugi0: In the worst case (i.e. when the Control Panel page does not load, or when the test is on another page) using `Wait Until` would wait the full timeout. In the best case, when the page loads instantly, it would not wait at all. In a normal case, where the text takes some time to appear, it would potentially wait ~1s or something like that?
The sleep instead just waits for a couple of seconds in all cases.
We would need to understand how likely it is for this code to be reached when we are in the worst case to understand which option would add less overhead on average. I am merging this since I want to run the full sanity suite, but we can also drop the sleep down to 1s (or even 0.5s if it works), that way it should be on par with the best case scenario of the wait until without the risk of the added overhead.
Let me know if it works for you @jgarciao

PR #509, opened 2022-08-24
Minor Fixes

bdattoma:
1) Verify User Is Able to Activate Anaconda Professional. 
The kw which is checking the KfDef takes only the first KfDef instead of looping over all of them. Since in v1.15 Anaconda has its own KfDef, the kw is not able to capture it.

2) Verify Intel AIKIT Operator Can Be Installed Using OpenShift Console. 
Increased the time for checking if the service is available in the dashboard

3) Verify User Is Able to Produce and Consume Events (RHOSAK)
RHODS Dashboard changed Window title in the last release, so this PR changes it in the TC

4) Keyword "Remove Disabled Application From Enabled Page"
screenshot is replaced everytime by the last TC run because it uses static name. PR changes it by appending the ISV app id in the screenshot title

5) ods-ci.pod.yaml
Syntax error fixed

6) test case "Verify That Idle JupyterLab Servers Are Culled In Sandbox Environment After 24h" 
RHODS Dashboard changed Window title in the last release, so this PR changes it in the TC

7) test case "Verify RHODS Dashboard Explore And Enabled Page Has No Message With No Component Found". A config map has changed title breaking the TC. The title will be changed back but for the moment I will patch the TC to work with the new name (bug has been raised)

### DIFF
--- a/tests/Tests/100__deploy/100__installation/102__post_install.robot
+++ b/tests/Tests/100__deploy/100__installation/102__post_install.robot
@@ -245,7 +245,7 @@ Verify RHODS Dashboard Explore And Enabled Page Has No Message With No Component
     ...     configmap in openshift
     ...     ProductBug:RHODS-4308
     [Setup]   Test Setup For Rhods Dashboard
-    Oc Patch    kind=ConfigMap      namespace=redhat-ods-applications    name=rhods-enabled-applications-config    src={"data":null}   #robocop: disable
+    Oc Patch    kind=ConfigMap      namespace=redhat-ods-applications    name=odh-enabled-applications-config    src={"data":null}   #robocop: disable
     Delete Dashboard Pods And Wait Them To Be Back
     Reload Page
     Menu.Navigate To Page    Applications   Enabled
--- a/tests/Tests/100__deploy/100__installation/102__post_install.robot
+++ b/tests/Tests/100__deploy/100__installation/102__post_install.robot
@@ -271,7 +271,7 @@ Test Setup For Rhods Dashboard
 
 Test Teardown For Configmap Changed On RHODS Dashboard
     [Documentation]    Test Teardown for Configmap changes on Rhods Dashboard
-    Oc Patch    kind=ConfigMap      namespace=redhat-ods-applications    name=odh-enabled-applications-config    src={"data": {"jupyterhub": "true"}}   #robocop: disable
+    Oc Patch    kind=ConfigMap      namespace=redhat-ods-applications    name=odh-enabled-applications-config    src={"data": {"jupyter": "true"}}   #robocop: disable
     Delete Dashboard Pods And Wait Them To Be Back
     Close All Browsers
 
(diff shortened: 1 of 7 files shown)


lugi0 (dismissed):
lgtm

bdattoma:
fixes 1),2),4) still in progress due to some failures obtained in previous Jenkins runs

fix 3) verified with [Job#1184](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1184/TestReport/) _SUCCESS_

fix 5) tested offline (not possible to test with Jenkins at the moment)

fix 6) not feasible the testing for this PR, however it very simple change in the window title (same as 3 which is verified)

fix 7) is "passed" in the [Job#1184](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1184/TestReport/) but it is expected to fail, unless the bug has been resolved. Need investigation

jgarciao (dismissed):
lgtm

tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -271,7 +271,7 @@ Test Setup For Rhods Dashboard
 
 Test Teardown For Configmap Changed On RHODS Dashboard
     [Documentation]    Test Teardown for Configmap changes on Rhods Dashboard
-    Oc Patch    kind=ConfigMap      namespace=redhat-ods-applications    name=rhods-enabled-applications-config    src={"data": {"jupyterhub": "true"}}   #robocop: disable
+    Oc Patch    kind=ConfigMap      namespace=redhat-ods-applications    name=odh-enabled-applications-config    src={"data": {"jupyterhub": "true"}}   #robocop: disable
```
tarukumar, line 274: shouldn't you change  from  src={"data": {"jupyterhub": "true"}} to src={"data": {"jupyterhub": "true", "jupyter": "true"}} for KFNBC

bdattoma: I think it should contain only  "jupyter": "true" now that we are going to merge kfnbc branch

Thanks for catching it

jgarciao (approved):

tests/Tests/100__deploy/100__installation/102__post_install.robot
tarukumar: What i have observed during migration is that it contains both 'jupyterhub` and `jupyter` in the configmap. It shouldn't contain both it should only contain jupyter

bdattoma: it's a bug/question to raise to developers about migration script

tarukumar: even in fresh cluster you can observe this, I think we can open a bug in downstream. Regarding migration i have asked developer in git

bdattoma: I will open the bug

tarukumar: nope

tarukumar (approved):
LGTM

PR #515, opened 2022-08-26
remove JH

tarukumar:

Add coded to skip TC after JH removal in rhods 1.16

### DIFF
--- a/tests/Tests/100__deploy/100__installation/102__post_install.robot
+++ b/tests/Tests/100__deploy/100__installation/102__post_install.robot
@@ -48,16 +48,22 @@ Verify Traefik Deployment
     ${containerNames} =  Create List  traefik-proxy  configmap-puller
     Verify Deployment  ${traefik}  3  2  ${containerNames}
 
-Verify JH Deployment
+Verify JH or Notebook Controller Deployment
     [Documentation]  Verifies RHODS JH deployment
     [Tags]    Sanity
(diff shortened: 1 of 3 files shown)


tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -41,17 +41,29 @@ Verify Traefik Deployment
     ...       Tier1
     ...       ODS-546
     ...       ODS-552
+     ${version-check} =  Is RHODS Version Greater Or Equal Than  1.16.0
+    Skip if   ${version-check}==True
+    ...   Traefik proxy is removed after KFNBC migration
```
lugi0, line 46: Since this is going in the KFNBC branch, the version check is not needed. You can skip directly.
Same for all other version checks

lugi0 (changes requested):

pablofelix (approved):
lgtm

tests/Tests/100__deploy/100__installation/102__post_install.robot
tarukumar: I think it is good idea to keep it once it become master

lugi0: I don't think it will be needed since we have a 1.15 branch that was created for the release, and master will become the branch for 1.16

tarukumar: Done

PR #523, opened 2022-08-29
[KFNBC] Adapt different TCs to work with KFNBC and without old config maps

bdattoma:
ODS-293 "Verify User Can Set Custom RHODS Groups" is still using the old config map way to set RHODS Access settings, instead of the new Dashboard CRD. This PR wants to fix the TC to work with the CRD.

Besides it fixes the compatibility issues between the TC and KFNBC

**UPDATE**:
In addition, the PR adapt ODS-1408,1494,1500 which used to read the configmap.
The PR leaves outside ODS-1495 which will need to be updated, but I'll do in another PR

### DIFF
--- a/tests/Resources/ODS.robot
+++ b/tests/Resources/ODS.robot
@@ -106,24 +106,21 @@ Set Standard RHODS Groups Variables
 Apply Access Groups Settings
     [Documentation]    Changes the rhods-groups config map to set the new access configuration
     ...                and rolls out JH to make the changes effecting in Jupyter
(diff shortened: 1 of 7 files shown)


bdattoma:
[Jenkins Job 1211 ](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1211/TestReport/) _SUCCESS_

bdattoma:
closed in favor of #533

PR #531, opened 2022-08-30
Add TCs for pod tolerations

lugi0:

### DIFF
--- a/tests/Tests/500__jupyterhub/test-pod-affinity.robot
+++ b/tests/Tests/500__jupyterhub/test-pod-affinity.robot
@@ -1,0 +1,66 @@
+*** Settings ***
+Resource         ../../Resources/Page/ODH/ODHDashboard/ODHDashboard.resource
+Suite Setup      Begin Web Test
+Suite Teardown   Teardown
+
+*** Variables ***
+
+*** Test Cases ***
+
+Test Setting Pod Toleration Via UI
+    [Documentation]    Sets a Pod toleration via the admin UI
+    [Tags]  Sanity    Tier1
+    ...     ODS-1684
+    Menu.Navigate To Page    Settings    Cluster settings
+    Wait Until Page Contains    Notebook pod tolerations
+    Set Pod Toleration Via UI    TestToleration
+    Save Changes In Cluster Settings
+
+Verify Toleration Is Applied To Pod
+    [Documentation]    Verifies Pod spawns with toleration
+    [Tags]  Sanity    Tier1
+    ...     ODS-1685
+    Launch JupyterHub Spawner From Dashboard
+    Spawn Notebook With Arguments    image=s2i-minimal-notebook
+    Verify Pod Toleration    TestToleration
+
+
+*** Keywords ***
+
+Set Pod Toleration Via UI
+    [Documentation]    Sets toleration using admin UI
+    [Arguments]    ${toleration}
+    Wait Until Page Contains Element    xpath://input[@id="tolerations-enabled-checkbox"]
+    Sleep  2s
+    Click Element    xpath://input[@id="tolerations-enabled-checkbox"]
+    Wait Until Element Is Enabled    xpath://input[@id="toleration-key-input"]
+    Input Text    xpath://input[@id="toleration-key-input"]    ${toleration}
+
+Verify Pod toleration
+    [Documentation]    Verifies Pod contains toleration
+    [Arguments]    ${toleration}
+    ${expected} =    Set Variable    Tolerations    ${toleration}:NoSchedule op=Exists
+    ${current_user} =    Get Current User In JupyterLab
+    ${notebook_pod_name} =   Get User Notebook Pod Name  ${current_user}
+    OpenShiftLibrary.Search Pods    ${notebook_pod_name}  namespace=rhods-notebooks
+    ${output} =    Run   oc describe pod ${notebook_pod_name} -n rhods-notebooks | grep -i tolerations:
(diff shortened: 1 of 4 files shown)


lugi0:
Tested locally, need to test through Jenkins

tests/Tests/500__jupyterhub/test-pod-affinity.robot
```diff
@@ -0,0 +1,66 @@
...
+    Set Pod Toleration Via UI    TestToleration
+    Save Changes In Cluster Settings
+
+Verify Toleration Is Applied To Pod
+    [Documentation]    Verifies Pod spawns with toleration
+    [Tags]  Sanity    Tier1
+    ...     ODS-1685
+    Launch JupyterHub Spawner From Dashboard
+    Spawn Notebook With Arguments    image=s2i-minimal-notebook
+    Verify Pod Toleration    TestToleration
```
bdattoma, line 25: maybe these 2 lines could be put inside `Test Setting Pod Toleration Via UI`

tests/Tests/500__jupyterhub/test-pod-affinity.robot
```diff
@@ -0,0 +1,66 @@
...
+    @{received} =    Split String    ${output}    separator=:${SPACE}
+    FOR    ${idx}    IN RANGE    2
+        ${el} =  Remove From List  ${received}  0
+        ${el} =  Strip String  ${el}
+        Append To List  ${received}  ${el}
+    END
+    Lists Should Be Equal    ${received}    ${expected}
+    Log    ${received}
+    Log    ${expected}
+
```
bdattoma, line 56: would make sense to check the node where the pod has been scheduled on? To check if the toleration settings are correct

lugi0:
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1216/console
Pass

tests/Tests/500__jupyterhub/test-pod-affinity.robot
lugi0: IMHO this wouldn't really show anything useful. The pod can still be scheduled anywhere, but if there is a node with a matching taint then only pods with the toleration set should get scheduled on it.
Testing that behaviour completely (i.e. making sure that only pods with a toleration set - and default openshift pods - get scheduled on a tainted node) seems really out of scope for us. I think we can assume that OCP tests and validates this functionality for us.

tests/Tests/500__jupyterhub/test-pod-affinity.robot
lugi0: That would mean having only one TC with both Polarion IDs, which I wanted to avoid.
Or are you suggesting something else?

bdattoma: hmm I think ODS-1685 would complete the test in ODS-1684. So, basically, I'm proposing to have one sigle polarion ID gathering the checks from ODS-1685 and ODS-1684.

tests/Tests/500__jupyterhub/test-pod-affinity.robot
lugi0: Well, the test checks that the toleration is set correctly in the pod, and if that's the case we can rely on OCP to make sure that it would work with the corresponding taint. 
The only way I can think of to verify what you're suggesting is to fill all the other nodes with pods and make sure that at least one pod is scheduled on the tainted node (rather than fail being scheduled), but that would make the test run quite longer for no real benefit IMHO

tests/Tests/500__jupyterhub/test-pod-affinity.robot
lugi0: Mh, I don't really agree. ODS-1684 is testing the admin UI and verifying that the toleration can be set (and the setting saved). 1865 then takes it a step further to verify that the toleration is actually applied.
I agree that it could be one bigger test case, but I personally prefer it this way so that if something break we know which part it is right away.

bdattoma: you would know in any case which part is broken and which is right. Well anyway, we can leave as it is, don't have a strong preference here.

bdattoma (approved):

tests/Tests/500__jupyterhub/test-pod-affinity.robot
```diff
@@ -0,0 +1,66 @@
...
+    ${notebook_pod_name} =   Get User Notebook Pod Name  ${current_user}
+    OpenShiftLibrary.Search Pods    ${notebook_pod_name}  namespace=rhods-notebooks
+    ${output} =    Run   oc describe pod ${notebook_pod_name} -n rhods-notebooks | grep -i tolerations:
+    @{received} =    Split String    ${output}    separator=:${SPACE}
+    FOR    ${idx}    IN RANGE    2
+        ${el} =  Remove From List  ${received}  0
+        ${el} =  Strip String  ${el}
+        Append To List  ${received}  ${el}
+    END
+    Lists Should Be Equal    ${received}    ${expected}
```
jgarciao, line 53: For people not knowing this about this test, a `msg=Unexpected Pod Toleration` could help when looking at the logs

tests/Tests/500__jupyterhub/test-pod-affinity.robot
```diff
@@ -0,0 +1,66 @@
+*** Settings ***
```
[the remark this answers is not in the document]
bdattoma: +1

PR #548, opened 2022-09-06
Add support to create osd cluster on gcp

aloganat:

### DIFF
--- a/utils/scripts/ocm/ocm.py
+++ b/utils/scripts/ocm/ocm.py
@@ -8,6 +8,8 @@ import sys
 import jinja2
 import time
 import glob
+import io
+from contextlib import redirect_stdout, redirect_stderr
 
 dir_path = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(dir_path + "/../")
@@ -26,7 +28,7 @@ Class for Openshift Cluster Manager
 """
 
 
-class OpenshiftClusterManager():
+class OpenshiftClusterManager:
     def __init__(self, args={}):
 
         # Initialize instance variables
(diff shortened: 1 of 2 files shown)


utils/scripts/ocm/ocm.py
```diff
@@ -172,28 +182,46 @@ def osd_cluster_create(self):
...
+            self._create_service_account_file()
+            cmd = (
+                "ocm create cluster --provider {} --service-account-file {} "
+                "--ccs --region {} --compute-nodes {} "
+                "--compute-machine-type {} {} {}"
+                "{}".format(
+                    self.cloud_provider,
+                    self.service_account_file,
+                    self.region,
+                    self.compute_nodes,
```
bdattoma, line 214: is it `compute_nodes` the number of nodes? if yes, could we use `num_compute_nodes` like done with AWS? so we don't have 2 variables for the same value

utils/scripts/ocm/ocm.py
```diff
@@ -1145,12 +1178,13 @@ def compare_with_old_version_file(self):
         "releases/download/v0.1.55/ocm-linux-amd64"
     )
     parser = argparse.ArgumentParser(
+        usage=argparse.SUPPRESS,
```
bdattoma, line 1181: what does it change?

utils/scripts/ocm/ocm.py
```diff
@@ -1270,48 +1317,228 @@ def compare_with_old_version_file(self):
...
+        f = io.StringIO()
+        with redirect_stdout(f), redirect_stderr(f):
+            known_args = parser.parse_known_args()
+    except SystemExit:
+        pass
+    if known_args and "cloud_provider" in known_args[0]:
+        provider = known_args[0].cloud_provider
+        if provider == "aws":
+            required_create_cluster_parser.add_argument(
+                "--aws-account-id",
```
bdattoma, line 1423: wasn't it already defined at line 1324?

utils/scripts/ocm/ocm.py
aloganat: It is same berto. But as per argparse logic written below, it needs to be 2 different variable. I will take this as enhancement and send a new PR for this.

utils/scripts/ocm/ocm.py
aloganat: Yes, one is exclusively for help content and another one is for getting input from user. This is kind of workaround in argparse that we have to do, if we are using if else stmt in between the user inputs.

utils/scripts/ocm/ocm.py
aloganat: argparse.SUPPRESS is used to handle hidden arguments. This is required for showing the expected help content since we used if else stmt in between user inputs.

bdattoma (approved):

tarukumar (approved):
lgtm

tarukumar:
@aloganat could you also update document and add how to create cluster on GCP. https://github.com/red-hat-data-services/ods-ci/blob/master/utils/scripts/ocm/README.md

PR #549, opened 2022-09-06
Workaround for modal not appearing during spawn

lugi0:

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -19,6 +19,7 @@ ${JUPYTERHUB_DROPDOWN_XPATH} =    //button[@aria-label="Options menu"]
 ${JUPYTERHUB_CONTAINER_SIZE_TITLE} =    //div[.="Deployment size"]/..//span[.="Container Size"]
 ${KFNBC_MODAL_HEADER_XPATH} =    //div[@aria-label="Starting server modal"]
 ${KFNBC_MODAL_CANCEL_XPATH} =    ${KFNBC_MODAL_HEADER_XPATH}//button[.="Cancel"]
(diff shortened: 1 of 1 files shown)


bdattoma (approved):

jgarciao (approved):
lgtm

jgarciao (approved):

bdattoma (approved):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -130,10 +131,60 @@ Spawn Notebook
     [Documentation]  Start the notebook pod spawn and wait ${spawner_timeout} seconds (DEFAULT: 600s)
     [Arguments]  ${spawner_timeout}=600 seconds
     Click Button  Start server
-    Wait Until Page Contains  Starting server  60s
+    # Waiting for 60 seconds, since a long wait seems to redirect the user to the control panel
+    # if the spawn was successful
+    ${modal} =    Run Keyword And Return Status    Wait Until Page Contains
+    ...    Starting server    60s
+    IF  ${modal}==False
+        Log    message=Starting server modal didn't appear after 60s    level=ERROR
```
kpouget, line 139: @lugi0 @jgarciao @bdattoma, the error message shouldn't be after 60s, it should be after 5s, 10s max.

It's fine for the test to wait longer and do what's necessary not to crash when the popup doesn't appear this fast, but you can't have only green lights when this happens...

bdattoma: this line would raise a red light since the level is set to ERROR. @lugi0 right?

kpouget: it will raise the red flag ... but only after `60s`. The popup should appear "immediately".

PR #552, opened 2022-09-07
Added rhods version to be available globally

tarukumar:
Set rhods version to be available globally

### DIFF
--- a/tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
+++ b/tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
@@ -92,10 +92,8 @@ Check IF URL On The Page Is Commercial
      END
 
 Get RHODS version
-    #@{list} =  OpenShiftCLI.Get  kind=ClusterServiceVersion  label_selector=olm.copiedFrom=redhat-ods-operator
-    #&{dict} =  Set Variable  ${list}[0]
-    #Log  ${dict.spec.version}
-    IF  "${RHODS_VERSION}" == "${None}"
+    [Arguments]    ${force_fetch}=False
+    IF  "${RHODS_VERSION}" == "${None}" or "${force_fetch}"=="True"
          ${RHODS_VERSION} =  Run  oc get csv -n redhat-ods-operator | grep "rhods-operator" | awk '{print $1}' | sed 's/rhods-operator.//'
     END
     Log  ${RHODS_VERSION}
--- a/tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
+++ b/tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
@@ -7,6 +7,7 @@ Library  ../../../../libs/Helpers.py
 
 *** Variables ***
 @{verification_list}           beta   preview   stage
+${RHODS_VERSION}              None
 
 *** Keywords ***
 Uninstall Operator
@@ -94,9 +95,11 @@ Get RHODS version
     #@{list} =  OpenShiftCLI.Get  kind=ClusterServiceVersion  label_selector=olm.copiedFrom=redhat-ods-operator
     #&{dict} =  Set Variable  ${list}[0]
     #Log  ${dict.spec.version}
-    ${ver} =  Run  oc get csv -n redhat-ods-operator | grep "rhods-operator" | awk '{print $1}' | sed 's/rhods-operator.//'
-    Log  ${ver}
-    [Return]  ${ver}
+    IF  "${RHODS_VERSION}" == "${None}"
+         ${RHODS_VERSION} =  Run  oc get csv -n redhat-ods-operator | grep "rhods-operator" | awk '{print $1}' | sed 's/rhods-operator.//'
+    END
+    Log  ${RHODS_VERSION}
+    [Return]  ${RHODS_VERSION}
 
 Is RHODS Version Greater Or Equal Than
     [Arguments]  ${target}
--- a/tests/Resources/RHOSi.resource
+++ b/tests/Resources/RHOSi.resource
@@ -3,7 +3,7 @@ Documentation       Applies RHOSi settings to run the test suites
 
 Library             RPA.RobotLogListener
 Resource            Page/ODH/Monitoring/Monitoring.resource
-
+Resource            Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
 
 *** Variables ***
 @{PROTECTED_KEYWORDS}=      Login To RHODS Dashboard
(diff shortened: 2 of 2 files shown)


tests/Resources/RHOSi.resource
```diff
@@ -27,6 +27,7 @@ RHOSi Setup
     ...                Do Not extend this keyword for high-level setup, e.g., don't open browser
     Protect Sensitive Variables In Keywords
     Suite Availability Setup    ${RHODS_PROMETHEUS_URL}    ${RHODS_PROMETHEUS_TOKEN}
+    Initialize Global Variables
```
jgarciao, line 30: I think this keyword should be after `Protect Sensitive Variables In Keywords`

tests/Resources/RHOSi.resource
```diff
@@ -40,3 +41,8 @@ Protect Sensitive Variables In Keywords
     [Documentation]    Register keywords which use sensitive data as "Protected"
     ...                to turn their log level to NONE using RobotLogListener by Robocorp
     Register Protected Keywords    names=@{PROTECTED_KEYWORDS}
+
+Initialize Global Variables
+    [Documentation]   Set varibled globally
+    ${RHODS_VERSION}    Get RHODS Version
```
jgarciao, line 47: From the style guide: _We’ll use the equal sign for variable assignments without having a space after the variable name (e.g. ${dashboard_url}=  Get Dashboard URL)_

tests/Resources/RHOSi.resource
```diff
@@ -40,3 +41,8 @@ Protect Sensitive Variables In Keywords
     [Documentation]    Register keywords which use sensitive data as "Protected"
     ...                to turn their log level to NONE using RobotLogListener by Robocorp
     Register Protected Keywords    names=@{PROTECTED_KEYWORDS}
+
+Initialize Global Variables
+    [Documentation]   Set varibled globally
```
jgarciao, line 46: Sets global variables

jgarciao (changes requested):

tests/Resources/RHOSi.resource
tarukumar: done

tests/Resources/RHOSi.resource
tarukumar: done

tests/Resources/RHOSi.resource
tarukumar: done

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
```diff
@@ -94,9 +95,11 @@ Get RHODS version
     #@{list} =  OpenShiftCLI.Get  kind=ClusterServiceVersion  label_selector=olm.copiedFrom=redhat-ods-operator
```
jgarciao, line 95: Could you add a force-reload argument to force getting the info from oc even if the global variable is already set?

tests/Resources/RHOSi.resource
```diff
@@ -40,3 +41,8 @@ Protect Sensitive Variables In Keywords
     [Documentation]    Register keywords which use sensitive data as "Protected"
     ...                to turn their log level to NONE using RobotLogListener by Robocorp
     Register Protected Keywords    names=@{PROTECTED_KEYWORDS}
+
+Initialize Global Variables
+    [Documentation]   Sets global variables
+    ${RHODS_VERSION}=    Get RHODS Version
```
jgarciao, line 47: force-reload=True   (see above comment)

jgarciao (dismissed):

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
```diff
@@ -94,9 +95,11 @@ Get RHODS version
...
     #&{dict} =  Set Variable  ${list}[0]
     #Log  ${dict.spec.version}
-    ${ver} =  Run  oc get csv -n redhat-ods-operator | grep "rhods-operator" | awk '{print $1}' | sed 's/rhods-operator.//'
-    Log  ${ver}
-    [Return]  ${ver}
+    IF  "${RHODS_VERSION}" == "${None}"
+         ${RHODS_VERSION} =  Run  oc get csv -n redhat-ods-operator | grep "rhods-operator" | awk '{print $1}' | sed 's/rhods-operator.//'
```
bdattoma, line 99: does it replace the global variable or create a local copy?

tarukumar: it creates local copy

jgarciao: Maybe an argument  set-global-variable=True (default True) could be added (instead of argument force-reload, proposed in another comment)

bdattoma: hm since you this kw will run in the RHOSi Setup, maybe here we don't need to return any value but just replace the Global variable value. What do you think?

tarukumar: I have used force fetch as an optional parameter

tests/Resources/RHOSi.resource
tarukumar: I have used force fetch and by default it will be false

tests/Resources/Page/OCPDashboard/InstalledOperators/InstalledOperators.robot
tarukumar: @bdattoma I think if i remove return if anyone is using the keyword directly for some TC they need modify the suite I don't want TC to failed if they are using it

bdattoma: yes the TCs need to ne updated. The refactoring is becoming always bigger xD but okay

bdattoma (approved):

jgarciao (approved):

kpouget:
How is `${RHODS_VERSION}` supposed to be populated when `Initialize Global Variables` is called with `${force_fetch}=False`?

in my code I call `Get Environment Variable  RHODS_VERSION  ${None}` to fetch it from the environment variable, or in other places I use `%{NOTEBOOK_URL}` to access a global variable when it must be defined, but I don't see any of it here ... ?

jgarciao:
@kpouget, you can add the variable with the value in test-variables.yml or [set it in command line](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#setting-variables-in-command-line) when running robot

kpouget:
I see, thanks. I'll update my code to use this

@tarukumar I thought you wanted to reduce the use of `test-variables.yml`, did I get that wrong?

PR #559, opened 2022-09-12
Attempt at optimizing openshift login step

lugi0:

### DIFF
--- a/tests/Resources/Page/LoginPage.robot
+++ b/tests/Resources/Page/LoginPage.robot
@@ -34,7 +34,9 @@ Select Login Authentication Type
 Login To Openshift
    [Arguments]  ${ocp_user_name}  ${ocp_user_pw}  ${ocp_user_auth_type}
     # Give the login prompt time to render after browser opens
-    Wait Until Element is Visible  xpath://div[@class="pf-c-login"]  timeout=15seconds
+    # FIXME: this wastes $timeout seconds every time it's called if 
+    # no login screen appears
+    Wait Until Element is Visible  xpath://div[@class="pf-c-login"]  timeout=10s
     ${select_auth_type} =  Does Login Require Authentication Type
     Run Keyword If  ${select_auth_type}  Select Login Authentication Type  ${ocp_user_auth_type}
     Wait Until Page Contains  Log in to your account
@@ -43,7 +45,7 @@ Login To Openshift
     Click Element  xpath=/html/body/div/div/main/div/form/div[4]/button
     # FIXME: replace this sleep for something more efficient, considering that this method is used for
     # authentication in OpenShift Console, but also RHODS dashboard and other places
-    Sleep  10
+    Sleep  2s
     # FIXME: this only applies to login to OpenShift Console, but this method is used
     # for authentication in RHODS dashboard and other places
     Maybe Skip Tour
--- a/tests/Tests/500__jupyterhub/test-jupyterlab-time.robot
+++ b/tests/Tests/500__jupyterhub/test-jupyterlab-time.robot
@@ -66,13 +66,15 @@ Spawn and Stop Server
     Run Keyword And Return Status    Wait Until JupyterHub Spawner Is Ready
     ${time1} =    Get Time    format=%H:%M:%S.%f
     Spawn Notebook
-    Run Keyword And Continue On Failure  Wait Until Page Does Not Contain Element
-         ...    id:progress-bar    timeout=600s
+    #Run Keyword And Continue On Failure  Wait Until Page Does Not Contain Element
+    #     ...    id:progress-bar    timeout=600s
     #Wait For JupyterLab Splash Screen    timeout=30
     Run Keyword And Warn On Failure   Login To Openshift  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
     ${authorization_required} =  Is Service Account Authorization Required
     Run Keyword If  ${authorization_required}  Authorize jupyterhub service account
-    Wait Until Page Contains Element  xpath://div[@id="jp-top-panel"]  timeout=60s
+    # If this fails we waited for 60s. Avg. time will be thrown off, might be acceptable
+    # given that we weren't able to spawn?
+    Run Keyword And Continue On Failure  Wait Until Page Contains Element  xpath://div[@id="jp-top-panel"]  timeout=60s
     ${time2} =    Get Time
     ${time} =    Subtract Date From Date    ${time2}    ${time1}
     Stop JupyterLab Notebook Server


tests/Resources/Page/LoginPage.robot
```diff
@@ -34,7 +34,9 @@ Select Login Authentication Type
 Login To Openshift
    [Arguments]  ${ocp_user_name}  ${ocp_user_pw}  ${ocp_user_auth_type}
     # Give the login prompt time to render after browser opens
-    Wait Until Element is Visible  xpath://div[@class="pf-c-login"]  timeout=15seconds
+    # FIXME: this wastes $timeout seconds every time it's called if 
+    # no login screen appears
+    Wait Until Element is Visible  xpath://div[@class="pf-c-login"]  timeout=10s
```
jgarciao, line 39: I've investigated a little bit about this and I think there is another approach that could work to optimize this keyword:
- if the url starts with https://oauth-openshift  login is required and we need to wait for the elements to be visible
- if the url is something else (e.g. https://jupyter-nb-ldap-2dadmin13-rhods-notebooks.apps.qeaisrhods-mon.XXXX.s1.devshift.org/notebook/rhods-notebooks/jupyter-nb-ldap-2dadmin13/lab) it means that there was a session already opened and we can just return from this keyword without waiting

lugi0: This should work. I've taken a first stab at it but we probably need to test is extensively before merging

jgarciao: Thanks!

tests/Resources/Page/LoginPage.robot
```diff
@@ -43,7 +51,7 @@ Login To Openshift
     Click Element  xpath=/html/body/div/div/main/div/form/div[4]/button
     # FIXME: replace this sleep for something more efficient, considering that this method is used for
     # authentication in OpenShift Console, but also RHODS dashboard and other places
-    Sleep  10
+    Sleep  2s
```
kpouget, line 54: what is the goal of this `Sleep 2s`, what it is supposed to wait for?
couldn''t it be replaced with a `Wait Until Page Contains  <something>` ?

lugi0: Unfortunately this keyword is called to log into Openshift, but after logging in you can be taken to any number of different web pages (e.g. the cluster web console, the rhods dashboard or the JL UI).
I don't think there's a single element we can use to understand when the login is complete, having a list of elements for every possible page we end up in sounds impossible, and waiting until something on the login page disappears doesn't guarantee that the next page is loaded.
As of right now I don't think we can do anything about this, the smallest sleep we can get away with is probably the best option

kpouget: but what's the point of the `Sleep  2s`? 
I mean, I get it clear from your comment that you cannot know what's coming next, but not knowing for not knowing, why waiting?

kpouget: > we try to hand off from one keyword to the next when the keyword is done doing its thing, 

the way I see it (in the few pages I coded) is that a keyword should do what's necessary to move to the next step ... but don't do anything related to the next step,

eg 1, the notebook spawner should let the page "loading" towards the next step (the authentication page, or jupyterlab)
eg 2, the authentication keyword should let the page "loading" towards the next step (the console, the dashboard, jupyterlab)

so `Maybe Skip Tour` doesn't belong to the Authentication keyword, but to the Console page that called it

tests/Resources/Common.robot
```diff
@@ -151,5 +151,17 @@ Skip If RHODS Version Greater Or Equal Than
        Skip If    condition=${version-check}==True    msg=This test is skipped for RHODS ${version} or greater
     END
 
-
-
+Get Domain From Current URL
+    [Documentation]    Gets the lowest level domain from the current URL
```
jgarciao, line 155: Can you add some examples in the [Documentation]?

tests/Resources/Common.robot
```diff
@@ -151,5 +151,17 @@ Skip If RHODS Version Greater Or Equal Than
...
+    ${current_url} =    Get Location
+    ${domain} =    Fetch From Left    string=${current_url}    marker=.
+    [Return]    ${domain}
+
+Is Current Domain Equal To
+    [Documentation]    Compare the lowest level domain to a given string
+    ...   and returns True/False
+    [Arguments]    ${url}
```
jgarciao, line 163: This keyword seems similar to https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Location%20Should%20Contain

jgarciao (changes requested):

tests/Resources/Common.robot
lugi0: Mh, I disagree. The Selenium keyword seems to be ok with the text being in any portion of the URL, instead this keyword explicitly checks the lowest level domain.
It might give us the same result for now, but we don't know how these URLs are going to evolve over time. I guess using the Selenium keyword with the `https` prefix anchors it to the beginning or the URL, but I'd rather use a keyword we have full control over

tests/Resources/Common.robot
lugi0: Done

tests/Resources/Common.robot
jgarciao: ok!

jgarciao (dismissed):
lgtm

tests/Resources/Page/LoginPage.robot
tarukumar: @lugi0 What about if we remove hard coded  sleep and add below code 
`Run Keyword And Ignore Error    Wait Until Keyword Succeeds    retry=2    retry_interval=1sec    Maybe Skip Tour`

lugi0:
Tested here: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1279/TestReport/
Takes down the login time from more than 10s to ~2 seconds. I've had to add an half second sleep in the test to account for the faster execution of the keywords. Before merging I'd try to run this with a bigger number of test cases to confirm that nothing breaks because of this change.

tests/Resources/Page/LoginPage.robot
lugi0: @tarukumar I think we can actually remove the Sleep now that `Maybe Skip Tour` has an additional check. I'll look into it.

lugi0: Removed the sleep, now `Maybe Skip Tour` uses `Wait Until Page Contains Element` rather than `Page Should Contain` since we can wait out the timeout when we know for sure we are in the ocp web console page.

lugi0:
Tested this locally using both the average spawn time test case and the full smoke suite. I haven't seen any failures due to the change in behaviour in the login keyword, but I would hold off on merging until we've tested this through Jenkins on a full Sanity run.

lugi0:
Test runs:
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1286/console
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1285/console

kpouget:
FYI, I applied this PR to the branch we're using for testing at scale, and it ran all fine.

[Timing](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/509/pull-ci-openshift-psap-ci-artifacts-master-ods-nb-ux-on-ocp/1572140635459162112/artifacts/nb-ux-on-ocp/test/artifacts/test_run_1/plotting/report_03_report:_launch_time_and_execution_time_distribution_files/fig_3.html) _with_ the PR applied:
image

[Timing](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/507/pull-ci-openshift-psap-ci-artifacts-master-ods-nb-ux-on-ocp/1571409987962408960/artifacts/nb-ux-on-ocp/test/artifacts/test_run_1/plotting/report_03_report:_launch_time_and_execution_time_distribution_files/fig_3.html) without it
image

==> median went from 22s to 11s \o/

lugi0:
@kpouget thanks for the data, awesome to know that it has a substantial positive effect!
Unfortunately from the test runs it does look like some of our tests are breaking due to these changes. I'll have to review the failures and see what can be done about it before we're able to merge this into ods-ci.

lugi0:
Latest run: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1291/TestReport/
I don't see any failure caused by these changes now. 

The only suspect thing I see is the `Maybe Close Popup` keyword misbehaving a couple of times (right after loading the JL UI) and keywords right after it failing because of an open popup. 
During the execution of the `maybe close popup` kw itself I see no popup open, so the speedup in logging into the server could somehow be part of it. 

It's a not a perfect solution but for now I'm adding .5s of sleep inside the keyword to try and catch all popups. 

Other than that I think this is now verified and could be merged.

bdattoma (approved):

pablofelix (approved):
lgtm

jgarciao (approved):
lgtm

PR #567, opened 2022-09-19
Make sure Pod Toleration setting is applied before running TC

lugi0:

### DIFF
--- a/tests/Tests/400__ods_dashboard/413__ods_dashboard_settings_pod_tolerations.robot
+++ b/tests/Tests/400__ods_dashboard/413__ods_dashboard_settings_pod_tolerations.robot
@@ -15,6 +15,9 @@ Test Setting Pod Toleration Via UI
     Wait Until Page Contains    Notebook pod tolerations
     Set Pod Toleration Via UI    TestToleration
     Save Changes In Cluster Settings
+    # New setting applies after ~5-10s. Sleep here to make
+    # sure it is applied for the next TC.
+    Sleep  10s
 
 Verify Toleration Is Applied To Pod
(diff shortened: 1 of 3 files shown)


tests/Tests/400__ods_dashboard/413__ods_dashboard_settings_pod_tolerations.robot
```diff
@@ -15,6 +15,9 @@ Test Setting Pod Toleration Via UI
     Wait Until Page Contains    Notebook pod tolerations
     Set Pod Toleration Via UI    TestToleration
     Save Changes In Cluster Settings
```
jgarciao, line 17: If you look for `Save Changes In Cluster Settings` you'll find at least 2 other places where a Sleep is added after the keyword. Could you add the Sleep inside `Save Changes In Cluster Settings`  and remove the other ones?

jgarciao (changes requested):

tests/Tests/400__ods_dashboard/413__ods_dashboard_settings_pod_tolerations.robot
lugi0: Are we sure we want the sleep in all cases? 
I'm adding it here simply because the test would fail otherwise as the new setting is not applied quickly enough

lugi0: Done

bdattoma (approved):

jgarciao (approved):
lgtm

PR #571, opened 2022-09-21
Spawn in same tab for rhods 1.17+

lugi0:

### DIFF
--- a/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -133,8 +133,11 @@ Get Spawner Environment Variable Value
 Spawn Notebook
     [Documentation]  Start the notebook pod spawn and wait ${spawner_timeout} seconds (DEFAULT: 600s)
     [Arguments]  ${spawner_timeout}=600 seconds
-    # Make sure server spawns in same tab
-    Click Element  xpath://input[@id="checkbox-notebook-browser-tab-preference"]
+    # Make sure server spawns in same tab in 1.17+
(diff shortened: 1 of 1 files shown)


jgarciao (approved):

bdattoma (approved):

tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -133,6 +133,12 @@ Get Spawner Environment Variable Value
 Spawn Notebook
     [Documentation]  Start the notebook pod spawn and wait ${spawner_timeout} seconds (DEFAULT: 600s)
     [Arguments]  ${spawner_timeout}=600 seconds
+    # TODO: Make sure server spawns in same tab in 1.17+
+    # Currently no way to know if option already selected or not
```
kpouget, line 137: what about something like that ? 🤔 
```
  ${notebook_browser_tab_preference}=  Set Variable  //input[@id="checkbox-notebook-browser-tab-preference"]
  ${new_tab_checked}  Get Element Attribute  ${notebook_browser_tab_preference}  checked
  Run Keyword If  "${new_tab_checked}" == "${None}"  Click Element  xpath:${notebook_browser_tab_preference}
```

lugi0: Last I looked at this the `checked` attribute was no different between the checkbox actually being checked or not.
I went down a rabbit hole chasing `Checkbox Should Be Selected` and `Select Checkbox` keywords inherited from Selenium to try and understand what they were doing, but I came up empty handed since it looks like those rely on the WebDriver / Browser / WebElement combination to behave according to the Living HTML standard and self-report their own _selectedness_ or _checkedness_

PR #587, opened 2022-10-03
Add parameter to spawn notebook in new tab

lugi0:

Tested both variations on a single TC each and both are working.
Did a full dry run to confirm no tests were broken by this change.
Might need a full sanity run to confirm no unexpected failures derive from this, although by default all TCs will use the same behaviour they've used until now.


jgarciao (approved):
lgtm

bdattoma (approved):

PR #590, opened 2022-10-03
Check for modal error message when there's not enough resources to spawn notebook

lugi0:

Automates TC for https://issues.redhat.com/browse/RHODS-5146

### DIFF
--- a/tests/Tests/500__jupyterhub/test.robot
+++ b/tests/Tests/500__jupyterhub/test.robot
@@ -38,6 +38,7 @@ Can Login to Jupyterhub
 
 Can Spawn Notebook
    [Tags]  Sanity  Smoke
+   ...     ODS-1789
    Fix Spawner Status
    Select Notebook Image  s2i-generic-data-science-notebook
    Select Notebook Image  s2i-minimal-notebook
--- a/tests/Tests/500__jupyterhub/test.robot
+++ b/tests/Tests/500__jupyterhub/test.robot
@@ -55,7 +55,7 @@ Can Spawn Notebook
    Remove Spawner Environment Variable  env_four
    Remove Spawner Environment Variable  env_five
    Remove Spawner Environment Variable  env_six
-   Verify Modal Does Not Get Stuck
+   Verify Notebook Spawner Modal Does Not Get Stuck When Requesting Too Many Resources To Spawn Server
    Spawn Notebook
    Login To Openshift  ${TEST_USER.USERNAME}  ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
    ${authorization_required} =  Is Service Account Authorization Required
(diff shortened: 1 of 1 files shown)


tests/Tests/500__jupyterhub/test.robot
```diff
@@ -107,3 +98,13 @@ Start New Pytorch Build
     ${new_buildname}=  Start New Build    namespace=redhat-ods-applications    buildconfig=s2i-pytorch-gpu-cuda-11.4.2-notebook
     Wait Until Build Status Is    namespace=redhat-ods-applications    build_name=${new_buildname}   expected_status=Running
     [Return]    ${new_buildname}
+
+Verify Modal Does Not Get Stuck
```
jgarciao, line 102: Could this keyword be renamed to be more descriptive, like for example using the polarion test tittle? `Verify Notebook Spawner Modal Does Not Get Stuck When Requesting Too Many Resources To Spawn Server`

jgarciao (changes requested):

tests/Tests/500__jupyterhub/test.robot
lugi0: Done

jgarciao (dismissed):

tests/Tests/500__jupyterhub/test.robot
```diff
@@ -106,6 +107,17 @@ Start New Pytorch Build
...
+Verify Notebook Spawner Modal Does Not Get Stuck When Requesting Too Many Resources To Spawn Server
+   [Documentation]    Try spawning a server size for which there's not enough resources
+   ...    spawner modal should show an error instead of being stuck waiting for resources
+   Select Container Size    X Large
+   Click Button    Start server
+   # This could fail because of https://bugzilla.redhat.com/show_bug.cgi?id=2132043
+   Wait Until Page Contains    Insufficient resources to start    timeout=1min
+   ...    error=Modal did not fail within 1 minute
+   Click Button    Cancel
+   Select Container Size    Small
```
bdattoma, line 119: why selecting Small? was the goal of the TC to explicitly use a large container?
Plus, if the spawning doesn't fail, this kw would fail and block the overall TC I guess

lugi0: It selects Small because the main test was already using that, so I just reset it back to the original value

This test I don't think will ever fail considering it's selecting XLarge and we never used nodes that are that big

bdattoma: hm so why it gets X Large selected?

lugi0: Because I need to make the spawn fail because of "not enough resources", and the easiest way to do that is to try spawning with the biggest size available

bdattoma: gotcha

bdattoma (approved):

jgarciao (approved):

PR #612, opened 2022-11-01
Modify aws cluster config to expose variables to user

aloganat:

### DIFF
--- a/tasks/Resources/Provisioning/Hive/AWS/aws-cluster.yaml
+++ b/tasks/Resources/Provisioning/Hive/AWS/aws-cluster.yaml
@@ -8,7 +8,7 @@ items:
   spec:
     baseDomain: rhods.ccitredhat.com
     imageSetRef:
-      name: rhods-openshift
+      name: {{ image_set|default("rhods-openshift") }}
     platform:
       aws:
         region: {{ aws_region|default("us-east-1") }}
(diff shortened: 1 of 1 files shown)


tarukumar (approved):
LGTM. Lets wait for the @pablofelix approval before we merge it

tasks/Resources/Provisioning/Hive/AWS/aws-cluster.yaml
```diff
@@ -77,7 +77,7 @@ items:
 - apiVersion: hive.openshift.io/v1
   kind: ClusterImageSet
   metadata:
-    name: rhods-openshift
+    name: {{ cluster_image_set|default("rhods-openshift") }}
```
jgarciao, line 80: @pablofelix Is there a way to populate the same variable in two places?

pablofelix (changes requested):
Variable cluster_image_set should have the same value as the base_image_set. It could be better to use only one variable image_set for example instead

aloganat:
> 

Hi @pablofelix , As I have mentioned in the call, if I use the same variable for example image_set, its not been substituted in both the places - https://github.com/aloganat/ods-ci/blob/aws_config_hive/tasks/Resources/Provisioning/Hive/AWS/oc-provision.robot#L8 . Could you let me know if there is a way to do it?

pablofelix (approved):
lgtm

jgarciao (approved):

PR #615, opened 2022-11-05
[WIP] add sys exit when cmd fails Minor changes

mwaykole:

### DIFF
--- a/utils/scripts/ocm/ocm.py
+++ b/utils/scripts/ocm/ocm.py
@@ -1179,6 +1179,7 @@ class OpenshiftClusterManager:
         ret = execute_command(run_change_channel_cmd)
         if ret is None:
             log.info("Failed to update the channel to {}".format(self.cluster_name))
+            sys.exit(1)
             return ret
 
     def update_ocm_policy(self):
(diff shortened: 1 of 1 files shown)


utils/scripts/ocm/ocm.py
```diff
@@ -1179,6 +1179,7 @@ def change_cluster_channel_group(self):
         ret = execute_command(run_change_channel_cmd)
         if ret is None:
             log.info("Failed to update the channel to {}".format(self.cluster_name))
+            sys.exit(1)
```
tarukumar, line 1182: I think it better if you check for the exiting channel on the cluster before making and modification. If channel is "Candidat" the do not modify it .

mwaykole: sure  that can be done in different PR this pr is just for sys.exit(1)

bdattoma, line 1182: is this funciton going to be used in Robot Framework? maybe tasks. If yes, then we should aboid sys.exit. You could check the solution applied in functions like install_rhoam_addon

bdattoma:
@mwaykole is this PR still necessary? if not, please close it

PR #621, opened 2022-11-15
Tweak model serving TCs for openvino backend

lugi0:

### DIFF
--- a/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
+++ b/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
@@ -72,7 +72,6 @@ Temporary Label MM Namespace
     ${label} =    Run    oc label namespace ${MODEL_MESH_NAMESPACE} opendatahub.io/generated-namespace=true --overwrite=true
     Log    ${label}
     Run Keyword And Continue On Failure  Should Be Equal As Strings    ${label}    namespace/${MODEL_MESH_NAMESPACE} not labeled
-    #Run Keyword And Continue On Failure  Should Be Equal As Strings    ${label}    "'opendatahub.io/generated-namespace' already has a value (true), and --overwrite is false"
 
 Verify Openvino Deployment
     Run Keyword And Continue On Failure  Temporary Label MM Namespace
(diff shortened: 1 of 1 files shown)


tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -8,7 +8,7 @@ Suite Teardown    Teardown Model Serving
 ${MODEL_MESH_NAMESPACE}=    mesh-test
 ${ODH_NAMESPACE}=    redhat-ods-applications
 ${MS_REPO}=    https://github.com/opendatahub-io/modelmesh-serving
-${EXPECTED_INFERENCE_OUTPUT}=    {"model_name":"example-onnx-mnist__isvc-82e2bf7ea4","model_version":"1","outputs":[{"name":"Plus214_Output_0","datatype":"FP32","shape":[1,10],"data":[-8.233052,-7.749704,-3.4236808,12.363028,-12.079106,17.26659,-10.570972,0.7130786,3.3217115,1.3621225]}]}
+${EXPECTED_INFERENCE_OUTPUT}=    {"model_name":"example-onnx-mnist__isvc-b29c3d91f3","model_version":"1","outputs":[{"name":"Plus214_Output_0","datatype":"FP32","shape":[1,10],"data":[-8.233053,-7.7497034,-3.4236815,12.3630295,-12.079103,17.266596,-10.570976,0.7130762,3.321715,1.3621228]}]}
```
jgarciao, line 11: Is it possible that the output will be slightly different on each execution? If yes I think it would be better to compare that the values are equal for some decimals and not the whole number

jgarciao (approved):

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: No, it shouldn't change between executions on the same model; As you can see it is slightly different from an older output, but that's because of the switch from nvidia triton to openvino.
In theory this shouldn't change unless there's a model/backend refresh.

tarukumar (approved):

PR #622, opened 2022-11-15
Use different image set names to load the config as expected

aloganat:


aloganat:
Issue is fixed in another file. This PR is not needed anymore.

PR #631, opened 2022-11-25
Happy path test for Model Serving UI

lugi0:


tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -1,48 +1,82 @@
...
+    # Verify No Models Are Present
+    Click Button  Create server
+    # Verify moved to DSP page
+    Create Data Science Project    title=${PRJ_TITLE}    description=${PRJ_DESCRIPTION}
+    Create S3 Data Connection    project_title=${PRJ_TITLE}    dc_name=model-serving-connection
+    ...            aws_access_key=${S3.AWS_ACCESS_KEY_ID}    aws_secret_access=${S3.AWS_SECRET_ACCESS_KEY}
+    ...            aws_bucket_name=ods-ci-s3
+    Create Model Server
+    Open Model Serving Home Page
+    Serve Model    project_name=${PRJ_TITLE}    model_name=test-model    framework=onnx    existing_data_connection=${TRUE}
```
bdattoma, line 47: Would it be possible/good idea to add some checks here to ensure the previous operations have created the necessary items/resources? 
Maybe also the Verify Openvino Deployment and Verify Serving Service keywords could be moved here

lugi0: Yes, seems like a good idea

tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.robot
```diff
@@ -0,0 +1,78 @@
...
+${S3_KEY_DC_INPUT_XP}=          xpath=//input[@aria-label="AWS field AWS_ACCESS_KEY_ID"]
+${S3_SECRET_DC_INPUT_XP}=          xpath=//input[@aria-label="AWS field AWS_SECRET_ACCESS_KEY"]
+${S3_ENDPOINT_DC_INPUT_XP}=          xpath=//input[@aria-label="AWS field AWS_S3_ENDPOINT"]
+${S3_REGION_DC_INPUT_XP}=          xpath=//input[@aria-label="AWS field AWS_DEFAULT_REGION"]
+${DC_SECTION_XP}=             xpath=//div[div//h4[@id="data-connections"]]
+${DC_ADD_BTN_1_XP}=           ${DC_SECTION_XP}/div/button[text()="Add data connection"]
+${DC_ADD_BTN_2_XP}=           xpath=//footer/button[text()="Add data connection"]
+${S3_BUCKET_DC_INPUT_XP}=     xpath=//input[@aria-label="AWS field AWS_S3_BUCKET"]
+${S3_DEFAULT_ENDPOINT}=    https://s3.amazonaws.com/
+${S3_DEFAULT_REGION}=    us-east-1
```
bdattoma, line 19: would it make sense to move these variables in the test-variables.yaml along with other S3 data which are already in there?

tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.robot
```diff
@@ -0,0 +1,78 @@
...
+    ${current}=    Get Element Attribute    xpath://div[@class="pf-c-number-input"]/div/input    value
+    ${difference}=    Evaluate    int(${number})-int(${current})
+    ${op}=    Set Variable    plus
+    IF    ${difference}<${0}
+        ${difference}=    Evaluate    abs(${difference})
+        ${op}=    Set Variable    minus
+    END
+    FOR  ${idx}  IN RANGE  ${difference}
+        IF ${op}==plus
+            Click Plus Button
```
bdattoma, line 52: where is it defined?

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -0,0 +1,66 @@
...
+        Select Radio Button    group_name=model-location    value=new-data-connection-radio
+        # Set Up New Data Connection
+    END
+    Click Button    Configure
+
+Select Project
+    [Documentation]
+    [Arguments]    ${project_name}
+    Click Element    xpath://span[.="Project"]/../../..//button[@aria-label="Options menu"]
+    Click Element    xpath://li/button[.="${project_name}"]
```
bdattoma, line 41: you could create a parameterized keyword to do selection, since I see the same 2 lines are used for other selection tasks (e.g., Select Framework or Select Existing Data Connection)

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -1,48 +1,82 @@
...
+#     # TODO: Replace with http://robotframework.org/robotframework/latest/libraries/Process.html#Run%20Process
+#     Run    git clone ${MS_REPO}
+#     Run    cd modelmesh-serving/quickstart && ./quickstart.sh ${ODH_NAMESPACE} ${MODEL_MESH_NAMESPACE}
 
 Verify Model Serving Installation
     [Documentation]    Verifies Model Serving resources
     [Tags]    ModelMesh_Serving
-    Run Keyword And Continue On Failure  Wait Until Keyword Succeeds  5 min  10 sec  Verify Openvino Deployment
+    # Run Keyword And Continue On Failure  Wait Until Keyword Succeeds  5 min  10 sec  Verify Openvino Deployment
     Run Keyword And Continue On Failure  Wait Until Keyword Succeeds  5 min  10 sec  Verify odh-model-controller Deployment
```
bdattoma, line 30: where is it defined?

tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.robot
lugi0: Yes, will do that

tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.robot
lugi0: At the end of the file, but not implemented yet

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
lugi0: Yeah, I was waiting to see if an id or other would be added to these elements

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: Not sure to which one you are referring, but all "Verify xxx Deployment" should be among the keywords of the same file

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -68,29 +112,48 @@ Verify odh-model-controller Deployment
...
+#     Log    ${label}
+#     Run Keyword And Continue On Failure  Should Be Equal As Strings    ${label}    namespace/${MODEL_MESH_NAMESPACE} not labeled
 
 Verify Openvino Deployment
-    Run Keyword And Continue On Failure  Temporary Label MM Namespace
+    # Run Keyword And Continue On Failure  Temporary Label MM Namespace
     @{ovms} =  Oc Get    kind=Pod    namespace=${MODEL_MESH_NAMESPACE}    label_selector=name=modelmesh-serving-ovms-1.x
     ${containerNames} =  Create List  rest-proxy  oauth-proxy  ovms  ovms-adapter  mm
     Verify Deployment    ${ovms}  2  5  ${containerNames}
     ${all_ready} =    Run    oc get deployment -l name=modelmesh-serving-ovms-1.x | grep 2/2 -o
```
bdattoma, line 125: it'd be possible to achieve the same by using openshiftlibrary

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: The openvino deployment gets scaled to 1 pod only after the first model is set up, so I've put those checks after the `Serve Model` keyword

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -0,0 +1,165 @@
...
+    [Return]    ${url}
+
+Open ${section} Options Menu
+    [Documentation]    Opens the "Options menu" dropdown for different sections
+    ...    Valid sections are: "Name", "Model framework", "Project"
+    Click Element    xpath://span[.="${section}"]/../../..//button[@aria-label="Options menu"]
+
+    # //div[.="test-model-no-token "]/../../td[@data-label="Project"] (.= project name)
+    # //div[.="test-model-no-token "]/../../td[@data-label="Inference endpoint"]//div[@class="pf-c-clipboard-copy__group"]/input (value=url)
+    # //div[.="test-model-no-token "]/../../td[@data-label="Status"]//span[contains(@class,"pf-c-icon__content")] (class=status?) (pf-m-danger > failed)  (pf-m-danger) (pf-m-success)
```
bdattoma, line 161: do you need these lines under comment?

lugi0: good catch, I don't, it was just a bunch of xpaths that I could need

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -68,29 +89,29 @@ Verify odh-model-controller Deployment
...
-
-Teardown Model Serving
-    [Documentation]  delete modelmesh stuff, Temporary until included in RHODS
-    # Seems like it's taking 10 minutes to stop reconciling deployments
-    Wait Until Keyword Succeeds  15 min  15s  Delete Model Serving Resources
-    Run    rm -rf modelmesh-serving
\ No newline at end of file
+    Verify Deployment    ${ovms}  ${num_replicas}  5  ${containerNames}
+    ${all_ready} =    Run    oc get deployment -n ${project_name} -l name=modelmesh-serving-model-server-${project_name} | grep ${num_replicas}/${num_replicas} -o
+    Should Be Equal As Strings    ${all_ready}    ${num_replicas}/${num_replicas}
```
bdattoma, line 98: is it needed? if the previous command return something, than I guess this is automatically passed. Because in the previous "oc get.." you are looking explicitly for `${num_replicas}/${num_replicas}`

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -68,29 +89,29 @@ Verify odh-model-controller Deployment
...
+
+Get Access Token via UI
+    [Documentation]
+    [Arguments]    ${project_name}    ${service_account_name}=default-name
+    Open Data Science Projects Home Page
+    Project Should Be Listed    ${project_name}
+    Open Data Science Project Details Page    ${project_name}
+    ${token}=    Get Model Serving Access Token via UI    ${service_account_name}
+    Open Model Serving Home Page
```
bdattoma, line 107: you got the token in the previous line, the kw can end at that point no?

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -68,29 +89,29 @@ Verify odh-model-controller Deployment
...
+    [Return]    ${token}
+
+Model Serving Suite Teardown
+    [Documentation]    Suite teardown steps after testing DSG. It Deletes
+    ...                all the DS projects created by the tests and run RHOSi teardown
+    Delete Model Via UI    test-model
```
bdattoma, line 113: instead of deleting models via UI in the teardown, would it be safer to delete it via CLI? Maybe you could have a separated test case to test the deletion via UI

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: The grep could return an empty line (if the pods are not replicas/replicas), so I need to check that what is returned is the correct thing

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: Yes, I guess it could

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: I think our main test should use the UI, since that will be what end users will also use, no?
We can have a CLI clean up in other tests where we don't explicitly care about this

bdattoma: >I think our main test should use the UI


yeah, i was thinking of situations where this fails and the suite teardown fails to clean up. However, you are deleting the project via CLI so even if the first fails, the model should get deleted. Do you know if it is the case? Would the model be successfully deleted if the project gets deleted?

lugi0: Yes, it would

bdattoma: ok then!

bdattoma (dismissed):
lgtm

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -70,6 +69,8 @@ Select Project
         Open Project Options Menu
         Click Element    xpath://li/button[.="${project_name}"]
         Wait Until Page Contains Element    //span[.="Project"]/../../..//span[.="${project_name}"]
+        # modal refresh is unreliable
+        Sleep    0.5s
```
bdattoma, line 73: you could try with the kw `Wait Until Generic Modal Disappears` (Common.robot)

lugi0: I don't need it to disappear, but it refreshes once the project is selected. If execution continues too fast some elements might not be found because they are still (re)loading

bdattoma: oh okay right, forget about that then

mwaykole (approved):

PR #633, opened 2022-11-29
Add tasks for Managed Starburst and first tests

bdattoma:
Adding tasks for:
- installing/Uninstalling Starburst addon [MISV-79](https://polarion.engineering.redhat.com/polarion/redirect/project/ManagedISV/workitem?id=MISV-79) [MISV-84](https://polarion.engineering.redhat.com/polarion/redirect/project/ManagedISV/workitem?id=MISV-84) [MISV-82](https://polarion.engineering.redhat.com/polarion/redirect/project/ManagedISV/workitem?id=MISV-82)
- registering the cluster to marketplace

Adding tests for:
- [MISV-86](https://polarion.engineering.redhat.com/polarion/redirect/project/ManagedISV/workitem?id=MISV-86)
- [MISV-85](https://polarion.engineering.redhat.com/polarion/redirect/project/ManagedISV/workitem?id=MISV-85)


libs/Helpers.py
```diff
@@ -142,3 +142,21 @@ def parse_file_for_tolerations(self, filename):
...
+        if not result:
+            self.BuiltIn.fail(
+                "Something got wrong while installing Managed Starburst. Check the logs"
+            )
+    
+    @keyword
+    def uninstall_managed_starburst_using_addon_flow(self, cluster_name):
+        ocm_client = OpenshiftClusterManager()
+        ocm_client.cluster_name = cluster_name
+        ocm_client.uninstall_managed_starburst_addon(exit_on_failure=False)
```
lugi0, line 162: No output from this call in case uninstallation fails?

tasks/Resources/RedHatMarketplace.resource
```diff
@@ -0,0 +1,46 @@
...
+
+Install RedHat Marketplace Operator
+    [Documentation]    Installs the RHM operator
+    Run    oc apply -f "https://marketplace.redhat.com/provisioning/v1/rhm-operator/rhm-operator-subscription?approvalStrategy=Automatic"
+
+Create RedHat Marketplace Secret
+    [Documentation]    Creates pull secret for fetching operators from RHM
+    [Arguments]    ${token}
+    Run    oc create secret generic redhat-marketplace-pull-secret -n openshift-redhat-marketplace --from-literal=PULL_SECRET=${token}
+    Run And Return Rc And Output    curl -sL https://marketplace.redhat.com/provisioning/v1/scripts/update-global-pull-secret | bash -s ${token}
```
lugi0, line 46: Is this output saved/used for anything?

tasks/Resources/SERH_OLM/install.resource
```diff
@@ -0,0 +1,145 @@
...
+
+Is Managed Starburst Installed
+    [Documentation]    Checks Managed Staburst is installed on the cluster.
+    ...                Returns TRUE if the resources created by the install process
+    ...                are found in the cluster
+    ${result}=  Run Keyword And Return Status
+    ...  Run Keywords
+    ...  Oc Get  kind=Project  field_selector=metadata.name=${STARBURST_CR_DEFAULT_NAMESPACE}  AND
+    ...  Oc Get  kind=CatalogSource  namespace=${STARBURST_CR_DEFAULT_NAMESPACE}
+    ...    field_selector=metadata.name=addon-managed-starburst-catalog
```
lugi0, line 28: Isn't this check enough? If this CS exists then the NS/Project must exist as well

tasks/Resources/SERH_OLM/install.resource
```diff
@@ -0,0 +1,145 @@
...
+    [Return]  ${result}
+
+Get Managed Starburst CR Status
+    [Documentation]     Get the installation stage by reading the "RHOAM installation object" YAML
+    ${status}=     Run    oc get starburstenterprise -n ${STARBURST_CR_DEFAULT_NAMESPACE} -o json | jq '.items[].status.conditions[] | .reason,.type' | sed 's/"//g'   # robocop: disable
+    ${status_list}=    Text To List    ${status}
+    # Run Keyword And Continue On Failure    List Should Contain Value    ${status_list}    Initialized
+    # Run Keyword And Continue On Failure    List Should Contain Value    ${status_list}    InstallSuccessful
+    # Run Keyword And Continue On Failure    List Should Contain Value    ${status_list}    Deployed
```
lugi0, line 37: Can these be removed?

tasks/Resources/SERH_OLM/install.resource
```diff
@@ -0,0 +1,145 @@
...
+    ...                are found in the cluster
+    ${result}=  Run Keyword And Return Status
+    ...  Run Keywords
+    ...  Oc Get  kind=Project  field_selector=metadata.name=${STARBURST_CR_DEFAULT_NAMESPACE}  AND
+    ...  Oc Get  kind=CatalogSource  namespace=${STARBURST_CR_DEFAULT_NAMESPACE}
+    ...    field_selector=metadata.name=addon-managed-starburst-catalog
+    [Return]  ${result}
+
+Get Managed Starburst CR Status
+    [Documentation]     Get the installation stage by reading the "RHOAM installation object" YAML
```
lugi0, line 32: "RHOAM installation object" YAML > "Starburst installation object" ?

tasks/Resources/SERH_OLM/install.resource
```diff
@@ -0,0 +1,145 @@
...
+        ${current_status}=   Get Managed Starburst CR Status
+        Log    Last status of Managed Starburst CR installation: ${current_status}    console=yes
+        ${completed_flag}=     Run Keyword And Return Status    Should Be Equal    ${current_status}    Deployed
+        Exit For Loop If    $completed_flag == True
+        Sleep    ${retries_interval}
+    END
+    IF    $completed_flag == False
+        Fail    Managed Starburst CR didn't get deployed after ${retries} retries.
+        ...     (time between retries: ${retries_interval}). Check the cluster...
+    END
```
lugi0, line 55: Wouldn't `Wait Until Keyword Succeeds    ${retries}x    ${retries_interval}    Get Managed Starburst CR Status` work as well?
Perhaps by slightly changing `Get Managed Starburst CR Status` to return pass/fail based on the internal status

tasks/Resources/SERH_OLM/install.resource
```diff
@@ -0,0 +1,145 @@
...
+        Log    Checking Managed Starburst pods are successfully deployed: retry ${retry_idx}    console=yes
+        ${current_status}=   Run Keyword And Return Status    Managed Starburst Pods Should Be Deployed
+        Log    Last status of Managed Starburst CR installation: ${current_status}    console=yes
+        Exit For Loop If    $current_status == True
+        Sleep    ${retries_interval}
+    END
+    IF    $current_status == False
+        Fail    Managed Starburst Pods didn't get deployed after ${retries} retries.
+        ...     (time between retries: ${retries_interval}). Check the cluster...
+    END
```
lugi0, line 102: Can't this be replace by 
```
Wait Until Keyword Succeeds    ${retries}x    ${retries_interval}    Managed Starburst Pods Should Be Deployed
```

tasks/Resources/SERH_OLM/install.resource
```diff
@@ -0,0 +1,145 @@
...
+
+Wait Until Managed Starburst Installation Is Completed
+    [Documentation]    Keeps checking for Starburst resources until
+    ...                all of them get the expected status
+    [Arguments]    ${cr_chk_retries}=30   ${cr_chk_retries_interval}=30s
+    ...            ${pods_chk_retries}=30   ${pods_chk_retries_interval}=30s
+    Wait Until Managed Starburst CR Is Deployed      retries=${cr_chk_retries}      retries_interval=${cr_chk_retries_interval}
+    Wait Until Managed Starburst Pods Are Deployed    retries=${pods_chk_retries}    retries_interval=${pods_chk_retries_interval}
```
lugi0, line 110: Consider replacing calls with `Wait Until Keyword Succeeds`

tasks/Resources/SERH_OLM/install.resource
```diff
@@ -0,0 +1,145 @@
...
+
+Delete Managed Starburst CRs
```
lugi0, line 112: Why are we manually deleting resources during testing?

tests/Resources/Page/ODH/AiApps/ManagedStarburst.resource
```diff
@@ -0,0 +1,105 @@
...
+${USERNAME_XP}=    xpath=//input[@id="username"]
+${STARBURST_ROUTE}=    web-ui
+
+
+*** Keywords ***
+Create Starburst Route
+    [Documentation]    Exposes starburst service for external connections
+    [Arguments]    ${name}
+    ${rc}=    Run And Return Rc    oc expose svc starburst --name ${name} -n ${STARBURST_CR_DEFAULT_NAMESPACE}
+    # ${rc}=    Run And Return Rc    oc create route edge ${name} --service=starburst --insecure-policy="Redirect" -n ${STARBURST_CR_DEFAULT_NAMESPACE}    # robocop: disable
```
lugi0, line 19: can this be removed?

tests/Resources/Page/ODH/AiApps/ManagedStarburst.resource
```diff
@@ -0,0 +1,105 @@
...
+
+
+*** Variables ***
+${STARBURST_CR_DEFAULT_NAMESPACE}=    redhat-starburst-operator
+${USERNAME_XP}=    xpath=//input[@id="username"]
+${STARBURST_ROUTE}=    web-ui
+
+
+*** Keywords ***
+Create Starburst Route
```
lugi0, line 15: I don't think we should be creating routes while testing?

tasks/Resources/SERH_OLM/install.resource
bdattoma: because per design, the Addon won't delete the CR

bdattoma: so in order to test the uninstallation, we need to remove the CR in the test

lugi0 (approved):
a few comments but overall lgtm

libs/Helpers.py
bdattoma: this function is just a wrapper for the actual function defined in ocm.py to let us using the latter as RF keyword. The check on uninstall is done in `uninstall_addon` function:
```def uninstall_addon(self, addon_name="managed-odh", exit_on_failure=True):
        """Uninstalls addon"""

        addon_state = self.get_addon_state(addon_name)
        if addon_state != "not installed":
            cluster_id = self.get_osd_cluster_id()
            cmd = "ocm delete /api/clusters_mgmt/v1/clusters/{}/addons/" "{}".format(
                cluster_id, addon_name
            )
            log.info("CMD: {}".format(cmd))
            ret = execute_command(cmd)
            if ret is None:
                log.info(
                    "Failed to uninstall {} addon on cluster "
                    "{}".format(addon_name, self.cluster_name)
                )
                if exit_on_failure:
                    sys.exit(1)```

File: tasks/Resources/SERH_OLM/install.resource (hunk 9 above)
Comment (lugi0) [reply tasks/Resources/SERH_OLM/install.resource:112]: Why won't the addon delete its own resources during uninstallation?
If the CR being left in the cluster is an expected behaviour after addon uninstallation, then we shouldn't need to delete it ourselves, right? We should actually expect to find it in the cluster after the addon has been uninstalled?

File: tasks/Resources/RedHatMarketplace.resource (hunk 2 above)
Comment (bdattoma) [reply tasks/Resources/RedHatMarketplace.resource:46]: ah no, it's a mistake. Fixing it

Comment (bdattoma) [reply tasks/Resources/RedHatMarketplace.resource:46]: ah no, we need the output only for debug reasons, because the output of this curl will say something like "global pull secret patched" or similar. If I do only "Run" the logs won't show anything

File: libs/Helpers.py (hunk 1 above)
Comment (lugi0) [reply libs/Helpers.py:162]: I understand that it's a wrapper, but wouldn't it make more sense to also pass along the output of the function that it's wrapping?

File: tasks/Resources/SERH_OLM/install.resource (hunk 3 above)
Comment (bdattoma) [reply tasks/Resources/SERH_OLM/install.resource:28]: hm probably yes? I mean, I expect during installation that the addon first creates the NS and then the CS. But in general, a CS can exist without the NS if i'm not mistaken. You could try deleting a NS and checking that the CS remains

File: tasks/Resources/SERH_OLM/install.resource (hunk 4 above)
Comment (bdattoma) [reply tasks/Resources/SERH_OLM/install.resource:37]: yes!

File: tasks/Resources/SERH_OLM/install.resource (hunk 5 above)
Comment (bdattoma) [reply tasks/Resources/SERH_OLM/install.resource:32]: fixed

File: tasks/Resources/SERH_OLM/install.resource (hunk 6 above)
Comment (bdattoma) [reply tasks/Resources/SERH_OLM/install.resource:55]: okay so by changing `Get Managed Starburst CR Status` into `Managed Starburst CR Status should be` and use it in combination with `Wait Until Keyword Succeeds`. Can do that

File: tasks/Resources/SERH_OLM/install.resource (hunk 7 above)
Comment (bdattoma) [reply tasks/Resources/SERH_OLM/install.resource:102]: makes sense

File: tests/Resources/Page/ODH/AiApps/ManagedStarburst.resource (hunk 10 above)
Comment (bdattoma) [reply tests/Resources/Page/ODH/AiApps/ManagedStarburst.resource:19]: I'm keeping it because trying to understand the best way of creating the route. Some conversations with Staburst going on

File: tasks/Resources/RedHatMarketplace.resource (hunk 2 above)
Comment (lugi0) [reply tasks/Resources/RedHatMarketplace.resource:46]: But then wouldn't it make sense to have a comparison/check on the output to confirm that it's what you expect to see?

File: tasks/Resources/SERH_OLM/install.resource (hunk 9 above)
Comment (bdattoma) [reply tasks/Resources/SERH_OLM/install.resource:112]: we discussed about it in a meeting  about this, and engineers said it's not ideal to delete the Custom Resource which can be managed by the customer. However, it may change in future release, I will raise a bug/story to continue the discussion on this

File: tasks/Resources/RedHatMarketplace.resource (hunk 2 above)
Comment (bdattoma) [reply tasks/Resources/RedHatMarketplace.resource:46]: yes I can add it

Comment (bdattoma) [reply tasks/Resources/RedHatMarketplace.resource]: done

Comment (bdattoma) [reply tasks/Resources/RedHatMarketplace.resource]: oh I did and reverted the changes. Basically this kw is used in the Setup, and if it fails (cluster already registered for example), the installation wouldn't continue. That's why I didn't do it earlier. I would leave the output variable there only for debug

**Review by lugi0 — APPROVED** [review]

File: utils/scripts/testconfig/test-variables.yml [hunk 12]
```diff
@@ -52,3 +52,7 @@ GITHUB_USER:
   EMAIL: GITHUB_EMAIL
   USERNAME: GITHUB_USERNAME
   TOKEN: GITHUB_TOKEN
+STARBURST:
```
mwaykole, line 55: Missing 
  OBS_CLIENT_SECRET:  
  OBS_CLIENT_ID
  OBS_URL:  
  OBS_TOKEN_URL:

mwaykole (changes requested):
lgtm once added 
Missing
OBS_CLIENT_SECRET:
OBS_CLIENT_ID
OBS_URL:
OBS_TOKEN_URL:

in utils/scripts/testconfig/generateTestConfigFile.py &
utils/scripts/testconfig/test-variables.yml

utils/scripts/testconfig/test-variables.yml
bdattoma: these are in the other PR: https://github.com/red-hat-data-services/ods-ci/pull/635/files#diff-c5322351257b41d6672b2a6ca4dfc126281fe481369b2db4a170e223e6c8c00d

mwaykole (approved):

PR #634, opened 2022-11-29
Restore ODS-293 test case

bdattoma:
Since "Administration" tab will be back in RHODS   v1.20, we can re-enable ODS-293 in Sanity


jgarciao (approved):
lgtm

tarukumar (approved):
lgtm

PR #636, opened 2022-12-01
Add GCP provision

pablofelix:

./run_robot_test.sh --test-case tasks/Tasks/provision_self_managed_cluster.robot --extra-robot-args '--variablefile infrastructure_configuration.yaml' --include self_managed_provision --skip-oclogin


tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -36,7 +36,7 @@ Get Cluster Credentials
     Append to File  cluster_details.txt  password=${credentials_splited[1]}\n
 
 Login To Cluster
-    # Set Log Level    None
+    Set Log Level    None
```
bdattoma, line 39: instead of using Set Log Level, you can register the keywords as "protected" in the RHOSi Setup. it manages to hide the values in logs without losing track of the keywords which are executed

tasks/Resources/Provisioning/Hive/AWS/aws-cluster.yaml
```diff
@@ -80,6 +71,6 @@ items:
     name: {{ image_set|default("rhods-openshift") }}
     namespace: rhods
   spec:
-    releaseImage: {{ openshift_image|default("quay.io/openshift-release-dev/ocp-release:4.10.18-x86_64") }}
+    releaseImage: {{ openshift_image|default("quay.io/openshift-release-dev/ocp-release:4.10.42-x86_64") }}
```
bdattoma, line 74: isn't there a way to fetch this value automatically?

tasks/Resources/Provisioning/Hive/gpu-provision.robot
```diff
@@ -0,0 +1,22 @@
+*** Settings ***
+Library           OperatingSystem
+
+*** Keywords ***
+Create GPU Node In Self Managed AWS Cluster
+    Set Log Level    Info
```
bdattoma, line 6: could you register the keyword as "protected" in the RHOSi setup instead of using Set Log Level?

tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -0,0 +1,91 @@
...
+    Append to File  cluster_details.txt  username=${credentials_splited[3]}\n
+    Append to File  cluster_details.txt  password=${credentials_splited[1]}\n
+
+Login To Cluster
+    Set Log Level    None
+    ${ns} =    Oc Get    kind=Namespace    label_selector=hive.openshift.io/cluster-pool-name=${infrastructure_configurations['hive_cluster_name']}
+    ${ClusterDeployment} =    Oc Get    kind=ClusterDeployment    name=${ns[0]['metadata']['name']}    
+    ...    namespace=${ns[0]['metadata']['name']}    api_version=hive.openshift.io/v1
+    ${apiURL} =    Set Variable    "${ClusterDeployment[0]['status']['apiURL']}"
+    ${credentials} =    Run and Return Rc And Output    oc extract -n ${ns[0]['metadata']['name']} secret/$(oc -n ${ns[0]['metadata']['name']} get cd ${ns[0]['metadata']['name']} -o jsonpath='{.spec.clusterMetadata.adminPasswordSecretRef.name}') --to=-
```
bdattoma, line 44: which credententials is it extracting? kubeadmin token?

tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -0,0 +1,91 @@
...
+    ${credentials_splited} =    Split To Lines    ${credentials[1]}
+    Run And Return Rc    oc login --username=${credentials_splited[3]} --password=${credentials_splited[1]} ${apiURL} --insecure-skip-tls-verify
+    Log    Logged in to Hive cluster    console=True
```
bdattoma, line 47: is it loggin into the created gcp cluster or into the cluster which creates the gcp cluster?

tasks/Tasks/provision_self_managed_cluster.robot
```diff
@@ -0,0 +1,38 @@
...
+Resource         ../Resources/Provisioning/Hive/deprovision.robot
+Resource         ../Resources/Provisioning/Hive/gpu-provision.robot
+Library          OperatingSystem
+Library          OpenShiftLibrary
+Library          String
+Library          Process
+
+*** Tasks ***
+Provision Self-Managed Cluster
+    [Documentation]    Provision a self-managed cluster
```
bdattoma, line 14: what's a self-managed cluster?

tasks/Resources/Provisioning/Hive/GCP/gcp-cluster.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - name: worker
+        platform:
+          gcp:
+            type: {{ worker_node_instance_type|default("custom-8-32768") }}
+        replicas: {{ worker_node_replicas|default("2") }}
+      controlPlane:
+        name: master
+        platform:
+          gcp:
+            type: custom-8-32768
```
aloganat, line 37: Could we expose master nodes platform type to get inputs from user and defaults to this value?

tasks/Resources/Provisioning/Hive/gpu-provision.robot
```diff
@@ -0,0 +1,22 @@
...
+    ${gpu_node} =    Run Process    sh    tasks/Resources/Provisioning/Hive/AWS/provision-gpu.sh
+
+Delete GPU Node In Self Managed AWS Cluster
+    ${gpu_nodes} =    Oc Get    kind=Machine    label_selector=machine.openshift.io/instance-type=g4dn.xlarge
```
lugi0, line 10: This only considers `g4dn.xlarge` but there can be other node flavours. I think it should be a parameter, no?

tasks/Resources/Provisioning/Hive/AWS/aws-cluster.yaml
pablofelix: We are using the variable openshift_image to add the value manually or dynamically. To fetch this value automatically, we can use the API to get the last version in the stable channel and use that value, but since we are testing in different channels with different OpenShift versions, that variable is better managed in the infra file. I'm using a default value but maybe it would be better to delete the default value to avoid confusion. What do you think?

tasks/Resources/Provisioning/Hive/gpu-provision.robot
pablofelix: I'm planning to make this change for the next sprint also at the moment is only available on AWS

tasks/Resources/Provisioning/Hive/provision.robot
pablofelix: The problem is that I don't want to have a dependency on the tests. Do you think that we could move the protected keyword to a  task?  given that this has to be a temporary user I believe that there is no problem with deleting the log level at the moment

tasks/Resources/Provisioning/Hive/provision.robot
pablofelix: kubeadmin

tasks/Resources/Provisioning/Hive/provision.robot
pablofelix: into the created GCP cluster

tasks/Tasks/provision_self_managed_cluster.robot
pablofelix: a cluster managed by the customer

tasks/Resources/Provisioning/Hive/GCP/gcp-cluster.yaml
```diff
@@ -0,0 +1,88 @@
...
+        machineNetwork:
+        - cidr: 10.0.0.0/16
+        networkType: OpenShiftSDN
+        serviceNetwork:
+        - 172.30.0.0/16
+      platform:
+        gcp:
+          projectID: {{ gcp_project_id }}
+          region: {{ gcp_region|default("us-central1") }}
+      pullSecret: {{ pull_secret|default("") }}
```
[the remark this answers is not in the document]
pablofelix: Is recommendable, depending on the cluster configuration that we want to use. By default is an empty string that gave us a 60 days trial cluster

tasks/Resources/Provisioning/Hive/GCP/gcp-cluster.yaml
pablofelix: changed

bdattoma (approved):

lugi0 (approved):

PR #655, opened 2022-12-16
Smarter comparison for model inference output

lugi0:

When the comparison is successful:
image

When an exception is thrown during comparison:
image

When the comparison of two elements fails:
image

### DIFF
--- a/libs/Helpers.py
+++ b/libs/Helpers.py
@@ -160,3 +160,42 @@ class Helpers:
         ocm_client = OpenshiftClusterManager()
         ocm_client.cluster_name = cluster_name
         ocm_client.uninstall_managed_starburst_addon(exit_on_failure=False)
+
+    @keyword
+    def inference_comparison(self, expected, received, threshold=0.00001):
+        try:
+            import numbers, decimal, ast
+
+            # Cast from string to python type
+            expected=ast.literal_eval(expected)
+            received=ast.literal_eval(received)
+
+            failures=[]
+            
+            def _inference_object_comparison(expected, received, threshold):
+                if isinstance(expected, dict):
+                    # if current element is a dict, compare the keys and then each value
+                    if not expected.keys()==received.keys(): failures.append([expected.keys(),received.keys()])
+                    for k in expected.keys():
+                        _inference_object_comparison(expected[k], received[k], threshold)
+                elif isinstance(expected, list):
+                    # if current element is a list, compare each value 1 by 1
+                    for id, _ in enumerate(expected): _inference_object_comparison(expected[id],received[id], threshold)				
+                elif isinstance(expected, numbers.Number):
+                    # if current element is a number, compare each value with a rounding threshold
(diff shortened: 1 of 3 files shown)


jgarciao (dismissed):
lgtm

libs/Helpers.py
```diff
@@ -160,3 +160,42 @@ def uninstall_managed_starburst_using_addon_flow(self, cluster_name):
...
+            failures=[]
+            
+            def _inference_object_comparison(expected, received, threshold):
+                if isinstance(expected, dict):
+                    # if current element is a dict, compare the keys and then each value
+                    if not expected.keys()==received.keys(): failures.append([expected.keys(),received.keys()])
+                    for k in expected.keys():
+                        _inference_object_comparison(expected[k], received[k], threshold)
+                elif isinstance(expected, list):
+                    # if current element is a list, compare each value 1 by 1
```
bdattoma, line 182: you're using dictionary right? do you think this function will be used for list too?

lugi0: even though it starts from a dictionary, it then compares element by element, and the dictionary might contain lists

pablofelix (approved):
/lgtm

libs/Helpers.py
bdattoma: oh I missed the line where you call the same function in the inside for loop. Okay

bdattoma (dismissed):

pablofelix (approved):

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -39,6 +39,11 @@ Verify Model Can Be Deployed Via UI
     Click Button    Create server
     Wait Until Page Contains    Data science projects
     Wait Until Page Contains Element    //button[.="Create data science project"]
+    ${already_present} =    Run Keyword And Return Status    Page Should Contain    ${PRJ_TITLE}
```
bdattoma, line 42: if you want there is the `Project Should Be Listed`, nothing will change from functionality pov

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -39,6 +39,11 @@ Verify Model Can Be Deployed Via UI
     Click Button    Create server
     Wait Until Page Contains    Data science projects
     Wait Until Page Contains Element    //button[.="Create data science project"]
```
bdattoma, line 41: `Is Data Science Projects Page Open`

bdattoma (approved):
left some comments, nothing critical.

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: ty, need to do another pass on this kw and will fix this up then

PR #681, opened 2023-01-24
[WIP]Fix for the api url missing issue

tarukumar:


tarukumar:
Need to discuss with pablo before merging it

bdattoma:
@tarukumar I think this change is not needed anymore: we have now  `Export Variables From File` which I guess achieve the same result. Hence, I'm closing this PR but feel free to re-open or raise a new one if needed

PR #685, opened 2023-01-27
Implement CR deletion using OpenshiftLibrary instead of UI

bdattoma:
In Tier2 runs we see often failing the step of deleting ISV's CustomResource, so this PR wants to improve that by replacing UI navigation with CLI for performing that step.
For pachyderm, it also replaces project creation/deletion with OpenshftiLibrary since it is a ery easy and quick change.

This is just a first piece of the final implementation where the install/uninstall process of ISV will be completely done via CLI instead UI.

**A note about openvino**
You'll see the code for Openvino is slightly different, because the default CR name is tide to a version (i.e., v2022.3), so there is an higher probability this could change over time and our teardown would fails. So, until we get to Install using CLI (where we will input the CR name), the Openvino teardown will delete all the CR instances of openvino. In theory, there should be only one which was created by the test itself. 
Adding now the input of a custom CR name (or just fetching the default one from the UI) would add just an additional point of failure in the UI testing which is not very stable over OCP versions

### DIFF
--- a/tests/Resources/Page/ODH/AiApps/Aikit.robot
+++ b/tests/Resources/Page/ODH/AiApps/Aikit.robot
@@ -13,7 +13,7 @@ Remove AIKIT Operator
 
 Uninstall AIKIT Operator
     [Documentation]    Uninstall intel aikit operator and it's realted component
-    [Arguments]    ${cr_kind}=AIKitContainers    ${cr_name}=intel-aikit-container
+    [Arguments]    ${cr_kind}=AIKitContainer    ${cr_name}=intel-aikit-container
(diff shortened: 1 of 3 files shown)


lugi0 (dismissed):

bdattoma:
Jenkins job for ODS-760, ODS-675, ODS-1137 and ODS-1161: [1535](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1535/) _RUNNING_

jgarciao (approved):
lgtm

lugi0 (approved):

PR #699, opened 2023-02-17
Fix Automation Bug in nb-events test and add product bug tag

bdattoma:
This PR is to fix a small change in nb-events endpoint. While fixing it a product bug was discovered: RHODS-7150.


jgarciao (approved):
lgtm

mwaykole (approved):

PR #711, opened 2023-03-08
Fix Model Serving in RHODS 1.23

lugi0:
Need to fully test this, cannot do it at the moment since AWS credentials have rotated and model does not get pulled from the S3 bucket.


lugi0:
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1579/console
All pass

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -29,7 +30,12 @@
     Page Should Contain    Model Serving
     Click Link      Model Serving
     Wait Until Page Contains    Manage and view the health and performance of your deployed models.    timeout=30
-    Wait Until Page Contains Element    ${MS_HEADER_XP}    timeout=30
+    ${version-check}=  Is RHODS Version Greater Or Equal Than  1.23.0
```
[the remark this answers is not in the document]
jgarciao: In RHODS 1.23 there are several html changes and ods-ci is not preserving compatibility with RHODS 1.22. It would be better not to add the version check. Also, Diego is removing all the occurrences of `Is RHODS Version Greater Or Equal Than` because it's not compatible with ODH. 

We can discuss in the automation meeting how we can deal with this in the future without using that keyword

jgarciao (changes requested):

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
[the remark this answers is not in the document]
lugi0: Removed the version checks

lugi0:
Tested locally and working fine, will test with Jenkins once it's up again
image

lugi0:
Jenkins test: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1590/console

jgarciao (approved):
lgtm

lugi0:
Verified with Jenkins: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1600/console

tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
```diff
@@ -23,9 +23,10 @@ ${SPAWNER_LINK}=        xpath=//*[text()="Launch Jupyter"]
 Open Data Science Projects Home Page
     [Documentation]    Verifies submenu Settings > "Data Science Projects" is visible
     # Page Should Contain    Data Science Projects
-    Click Link      Data Science Projects
+    SeleniumLibrary.Wait Until Page Contains    Data Science Projects
```
bdattoma, line 26: can you remove the Lib name?

bdattoma: wait do you need to wait for this? in theory, if you run `Wait for RHODS Dashboard to Load` before calling `Open Data Science Projects Home Page` you should not get any failures

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -26,10 +27,16 @@ ${MS_TABLE_STATUS_FAILURE}=    //span[contains(@class,"pf-c-icon__content")][con
 *** Keywords ***
 Open Model Serving Home Page
     [Documentation]    Verifies submenu Settings > "Model Serving" is visible and navigates to page
-    Page Should Contain    Model Serving
-    Click Link      Model Serving
-    Wait Until Page Contains    Manage and view the health and performance of your deployed models.    timeout=30
-    Wait Until Page Contains Element    ${MS_HEADER_XP}    timeout=30
+    SeleniumLibrary.Page Should Contain    Model Serving
+    SeleniumLibrary.Click Link      Model Serving
+    Sleep    1s
```
bdattoma, line 32: could you replace the sleep with `Wait for RHODS Dashboard to Load`?

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -26,10 +27,16 @@ ${MS_TABLE_STATUS_FAILURE}=    //span[contains(@class,"pf-c-icon__content")][con
...
+    ${loading_error} =    Run Keyword And Return Status    SeleniumLibrary.Page Should Contain    Problem loading model serving page
```
bdattoma, line 33: why that would happen?

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -26,10 +27,16 @@ ${MS_TABLE_STATUS_FAILURE}=    //span[contains(@class,"pf-c-icon__content")][con
...
+    IF    ${loading_error}
+        Sleep    5s
+        SeleniumLibrary.Reload Page
+    END
```
bdattoma, line 37: isn't the library search order handled in the Suite Setup of the test suite which is calling this kw? I believe it shouldn't be necessary to do so in the kw code

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -26,10 +27,16 @@ ${MS_TABLE_STATUS_FAILURE}=    //span[contains(@class,"pf-c-icon__content")][con
...
+    SeleniumLibrary.Wait Until Page Contains    Manage and view the health and performance of your deployed models.    timeout=30
+    SeleniumLibrary.Wait Until Page Contains Element    ${MS_HEADER_XP}    timeout=30
```
bdattoma, line 39: I think you could replace this with `Wait for RHODS Dashboard to Load`

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -35,10 +37,12 @@ Verify Model Can Be Deployed Via UI
...
     Open Model Serving Home Page
-    # Verify No Models Are Present
-    Click Button    Create server
-    Wait Until Page Contains    Data science projects
-    Wait Until Page Contains Element    //button[.="Create data science project"]
+    Clean Up Model Serving Page
+    Clean Up DSP Page
+    Open Model Serving Home Page
+    Try Opening Create Server
+    SeleniumLibrary.Wait Until Page Contains    Data science projects
```
bdattoma, line 44: I think you could replace both lines with ` Wait for RHODS Dashboard to Load`

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -35,10 +37,12 @@ Verify Model Can Be Deployed Via UI
...
+    SeleniumLibrary.Wait Until Page Contains Element    //button[.="Create data science project"]
```
bdattoma, line 45: I think you could replace both lines with  Wait for RHODS Dashboard to Load

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -35,10 +37,12 @@ Verify Model Can Be Deployed Via UI
...
     ${already_present} =    Run Keyword And Return Status    Page Should Contain    ${PRJ_TITLE}
```
bdattoma, line 46: probably not needed anymore because `Clean Up DSP Page` is called at the beginning of the test. Right?

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -61,12 +65,48 @@ Test Inference With Token Authentication
...
     ${out}=    Get Model Inference   ${MODEL_NAME}    ${INFERENCE_INPUT}    token_auth=${FALSE}
     Should Contain    ${out}    <button type="submit" class="btn btn-lg btn-primary">Log in with OpenShift</button>
 
+Verify Openvino_IR Model Via UI
+    [Documentation]    Test the deployment of an openvino_ir model
+    [Tags]    Sanity    Tier1
+    ...    ODS-2054
+    Open Model Serving Home Page
+    Clean Up Model Serving Page
+    Clean Up DSP Page
```
bdattoma, line 80: these keywords are called by `Try Opening Create Server` 2 lines later. Are still necessary here?

tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
lugi0: Without the Wait it would fail while the page was still loading.
>can you remove the Lib name?
I prefer to keep it in, as without the name it's always marked as an issue in my IDE. I also think it's better this way, so that it doesn't fail when called from somewhere that didn't set the search order.

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
lugi0: I've seen it happen while testing locally, sometimes the user would get an error page saying that they couldn't list the resources, and a refresh always fixed it

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
lugi0: I am trying to specify the library everywhere so that the search order doesn't matter

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: This is not only about the page loading, but also about finding the expected element(s)

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
bdattoma: hm interesting. Could you add a warn and a screenshot here? So next time we capture and report it (I believe it should not happen)

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: You're probably right, but this check shouldn't take more than a few ms

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: Try opening create server would just click the button if the pages are clean, so it doesn't really matter

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
bdattoma: hmm I personally don't like that because it make reading more difficult and add can be solved by one line only. We could talk in the automation meeting so we find an agreement otherwise we'll have some places where it is used and some places where it is not

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
lugi0: Wouldn't that keyword only wait for Model Serving to appear?

tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
bdattoma: >Without the Wait it would fail while the page was still loading.


probably you need to call `Wait for RHODS Dashboard to Load` it before running `Open Data Science Projects Home Page`

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
lugi0: agreed. 
I don't like the search order approach because you won't ever be sure what is actually being run when the test runs by just reading the code.

tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
bdattoma: the Wait kw would look for the same element `xpath://h1[text()="${expected_page}"]` vs `    xpath=//h1[text()="Deployed models"]`

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
bdattoma: so you don't need to call those keywords before `Try Opening Create Server` because if the pages are not cleaned, will get cleaned in the ELSE part

tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: there's no such variable in this suite

bdattoma: you can import it from DS Projects resources

tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
lugi0: To me it makes more sense to keep the wait inside this keyword. If the link that the keyword has to click is already present it won't add any run time to the tests, otherwise it will prevent a failure that could've been avoided.

bdattoma (approved):

lugi0:
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1606/console all pass

jgarciao (approved):
lgtm

PR #715, opened 2023-03-13
Backport ldap fix

aloganat:


bdattoma (changes requested):
I think it would be better to cherry pick the desired commit, because this PR is fetching commit which are required only in 1.23

PR #718, opened 2023-03-14
fix the issue for upgrade verification

tarukumar:
ODS-233 on upgrade cluster always fails because there are two replica set present and for older replica set we can not find the readyreplica value. this fix the issue.


tarukumar:
upgrade cluster: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1607/

aloganat (approved):
lgtm

mwaykole (approved):

PR #695, opened 2023-02-08
Task disconnected

pablofelix:
Steps:
1- Create a cluster with Hive
2- Add the variables in infrastructure_configuration.yaml
registryCA: <CA certificate>
registry_pull_secret: <registry_pull_secret>
3- Execute the task: 
./run_robot_test.sh --test-case tasks/Tasks/provision_self_managed_cluster.robot --extra-robot-args '--variablefile infrastructure_configuration.yaml' --include self_managed_disconnect --skip-oclogin
4- Follow RHODS official documentation to install/upgrade RHODS in a disconnected cluster

PR #719, opened 2023-03-15
Hive tasks to provision clusters on openstack (PSI)

manosnoam:
The following tasks now support Openstack provider:
- Provision Self-Managed Cluster
- Deprovision Self-Managed Cluster

To use, first add Openstack values to the `infrastructure_configuration.yaml`. 
Here's an example for Openstack on PSI:
```
infrastructure_configurations:
    provider: OSP
    hive_cluster_name: ods-qe-007
    hive_claim_ns: rhods
    aws_domain: rhods.ccitredhat.com
    ocp_version: 4.12.5
    osp_cloud_name: openstack
    osp_network: shared_net_5
    osp_flavor: g.standard.xl
    osp_yaml_encoded: {PSI clouds.yaml base64 encoded string}
```

In case you want to reuse existing floating IPs on PSI, instead of creating new IPs, add them too:
```
infrastructure_configurations:
    ...
    fip_api: 10.0.188.220
    fip_apps: 10.0.188.30
```

To Provision cluster:
```
run_robot_test.sh --test-case tasks/Tasks/provision_self_managed_cluster.robot --extra-robot-args '-i self_managed_provision'
```
To De-provision cluster:
```
run_robot_test.sh --test-case tasks/Tasks/provision_self_managed_cluster.robot --extra-robot-args '-i self_managed_deprovision'
```


jgarciao (changes requested):
Could you address the warnings by Robocop? [Here](https://github.com/red-hat-data-services/ods-ci/blob/master/docs/check-code-style.md) you'll find how to test it locally and also how to exclude robocop rules if you thing any doesn't have sense

In https://github.com/red-hat-data-services/ods-ci/pull/695 there are two variables in infrastructure_configuration related to the registry certification authority and the pull secret. Is the purpose of QUAY_PULL_SEC and QUAY_SSH_KEY the same? If yes, could you use those parameters?

What is the purpose of PSI_CLOUD_YAML_ECODED and PSI_CLOUD_NAME_ECODED?  Should it say _ENCODED or _BASE64?

Are these changes affecting just PSI or also AWS and GCP provisioning? If the latter, can you test it before this PR is merged? Thanks!

tasks/Resources/Provisioning/Hive/OSP/create_fips.sh
```diff
@@ -0,0 +1,120 @@
...
+echo "Exporting Floating IPs of API '$FIP_API' and *APPS '$FIP_APPS', and saving into: $CLUSTER_FIPS"
+
+: > "$CLUSTER_FIPS"
+
+echo "export FIP_API=$FIP_API" >> "$CLUSTER_FIPS"
+echo "export FIP_APPS=$FIP_APPS" >> "$CLUSTER_FIPS"
+
+cat "$CLUSTER_FIPS"
+
+. "$CLUSTER_FIPS"
```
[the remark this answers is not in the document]
manosnoam: create_fips.sh is an independent script, that can be executed outside of Robot framework.
$CLUSTER_FIPS is the created file with the floating ips to be exported, in order to be used in the Hive template. In order to export the fips, one must source the $CLUSTER_FIPS.

We can later move away from calling bash scripts within Robot (we do it several times on other places as well). In the future we can decide to implement the script logic into Robot as well.

tasks/Resources/Provisioning/Hive/OSP/hive_osp_cluster_template.yaml
```diff
@@ -0,0 +1,102 @@
...
+            type: ${infrastructure_configurations}[osp_flavor]
+        replicas: 3
+      metadata:
+        name: ${infrastructure_configurations}[hive_cluster_name]-install
+      networking:
+        clusterNetwork:
+        - cidr: 10.128.0.0/14
+          hostPrefix: 23
+        machineNetwork:
+        - cidr: 192.169.0.0/16
```
bdattoma, line 56: should these value be stored into a variable too?

manosnoam:
> Could you address the warnings by Robocop? [Here](https://github.com/red-hat-data-services/ods-ci/blob/master/docs/check-code-style.md) you'll find how to test it locally and also how to exclude robocop rules if you thing any doesn't have sense.
Sure, I will use it, thanks for the heads up! 
Please note that most of the Robocop Warnings are not new. It complains about missing Doc strings for already existing keywords (before this PR). I think we can add a new PR to fix all Robocop warnings out there. 

> In #695 there are two variables in infrastructure_configuration related to the registry certification authority and the pull secret. Is the purpose of QUAY_PULL_SEC and QUAY_SSH_KEY the same? If yes, could you use those parameters?
#695 is for using internal Registry, but I don't see reference to Quay secret.
In test-variables.yml in gitlab, I did not find references to quay "pullSecret" and "sshKey", neither in ods-ci templates. 
@aloganat told me to embed these vars in the Jenkins pipeline for now, until we decide where to place them (as we want to move away from using test-variables.yml, right ?)

> What is the purpose of PSI_CLOUD_YAML_ECODED and PSI_CLOUD_NAME_ECODED? Should it say _ENCODED or _BASE64?
Yes, indeed typo. it's the clouds.yaml for openstack in a base64 string, which is required for Hive.

> Are these changes affecting just PSI or also AWS and GCP provisioning? If the latter, can you test it before this PR is merged? 
I did not change the templates for GCP and AWS, but I will test them as well, in the pipeline.

tasks/Resources/Provisioning/Hive/OSP/hive_osp_cluster_template.yaml
manosnoam: On other provisioning we do not externalize the CIDRs, since the benefit of it is just for use cases of Networking tests. 
We can add a new PR to externalize these CIDRs to be passed by configuration, and Jenkins later.

setup.py
jgarciao, line 1: I think setup.py shouldn't be included in the PR, as it was replaced by poetry

jgarciao (changes requested):

jgarciao:
Could you document in https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/infrastructure_configuration_variables.md the new fields introduced in infrastructure_configuration?

setup.py
manosnoam: ack!

manosnoam:
> Could you document in https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/infrastructure_configuration_variables.md the new fields introduced in infrastructure_configuration?

Done

bdattoma (dismissed):

bdattoma (dismissed):

jgarciao (dismissed):

bdattoma (dismissed):

bdattoma (dismissed):

bdattoma (approved):

jgarciao (approved):

PR #721, opened 2023-03-17
Retrieve Test tags from local branch changes

diegolovison:
Apply some changes inside `tests/**/*.robot` or `tests/**/*.resource`
Run the script `cd tests && python read_pr.py`
The output:
```
destructive_tests: ...
slow_tests: ...
fast_tests: ...
```

### DIFF
--- a/tests/read_pr.py
+++ b/tests/read_pr.py
@@ -1,0 +1,175 @@
+import glob
+
+
+class ReadPR:
+    """
+    Recursive find '*** Keywords ***' text relationship for the files that were changed.
+    For each match retrieve and report what are the tags that we should run the test suite.
+    """
+
+    def __init__(self):
+        self.keyword_key = '*** Keywords ***'
+        self.magic_tag_separator = '_______'
+        self.magic_tag = '[Tags]'
+
+    def run(self):
+        data = self.get_diff()
+        files_changed = []
+        for line in data:
+            if line.startswith('diff --git '):
+                files_changed.append(line)
+        infos = []
+        for f in files_changed:
+            is_resource = f and f.endswith('.resource')
+            is_robot = f.endswith('.robot')
+            if is_robot or is_resource:
+                infos.append(self.get_sections_info(f))
+
+        all_tags = set()
+        for info in infos:
+            tags = self.search_content_tag(info)
+            all_tags = all_tags.union(tags)
+        self.generate_arg_commands(all_tags)
+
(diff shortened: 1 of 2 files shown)


tests/read_pr.py
```diff
@@ -0,0 +1,175 @@
...
+        data = self.get_diff()
+        files_changed = []
+        for line in data:
+            if line.startswith('diff --git '):
+                files_changed.append(line)
+        infos = []
+        for f in files_changed:
+            is_resource = f and f.endswith('.resource')
+            is_robot = f.endswith('.robot')
+            if is_robot or is_resource:
```
tarukumar, line 25: do we need line 22 23 and 24. line can be be sorted in below code :

`    for f in files_changed:
        if f.endswith(('.robot', '.resource')):
            infos.append(self.get_sections_info(f))`

tests/read_pr.py
```diff
@@ -0,0 +1,175 @@
...
+        self.keyword_key = '*** Keywords ***'
+        self.magic_tag_separator = '_______'
+        self.magic_tag = '[Tags]'
+
+    def run(self):
+        data = self.get_diff()
+        files_changed = []
+        for line in data:
+            if line.startswith('diff --git '):
+                files_changed.append(line)
```
tarukumar, line 20: list comprehension could be good solution as well. 

` files_changed = [line for line in data if line.startswith('diff --git ')]`

tests/read_pr.py
```diff
@@ -0,0 +1,175 @@
...
+        infos = []
+        for f in files_changed:
+            is_resource = f and f.endswith('.resource')
```
tarukumar, line 23: why `f and f.endswith('.resource')` ?

tarukumar:
overall looks good to me, however i would suggest to use list comprehension to simplify the filtering process in the code

tests/read_pr.py
diegolovison: a mistake

diegolovison:
@tarukumar it is ready for review again. we must squash before merging
Thanks for your review

tests/read_pr.py
diegolovison: Done

tests/read_pr.py
diegolovison: Done

tarukumar (approved):
lgtm

bdattoma (approved):

PR #723, opened 2023-03-20
Test and end-to-end pipeline workflow via API

diegolovison:
Test and end-to-end pipeline workflow via API

```
oc login with_admin
./ods_ci/run_robot_test.sh --set-urls-variables true --test-case ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
```

```
oc login with_admin
cd ods_ci
sh run_robot_test.sh --set-urls-variables true --test-case tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot  --test-variables-file test-variables.yml --extra-robot-args '-L DEBUG' --skip-oclogin
```

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: qe-dspa-admin
  namespace: dlovison-example1
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: admin
subjects:
    - kind: User
      name: ldap-user2
```

### DIFF
--- a/ods_ci/libs/DataSciencePipelinesAPI.py
+++ b/ods_ci/libs/DataSciencePipelinesAPI.py
@@ -1,0 +1,245 @@
+import subprocess
+from json import JSONDecodeError
+import requests
+import os
+import json
+import time
+from robotlibcore import keyword
+import base64
+
+
+class DataSciencePipelinesAPI:
+
+    def __init__(self):
+        self.auth_url = self.retrieve_auth_url()
+        self.route = ''
+        self.sa_token = None
+
+    @keyword
+    def create_operator(self, file):
+        print(os.path.abspath(__file__))
+        current_path = __file__.index('/libs/')
+        current_path = __file__[:current_path]
+        response, error = self.run_oc(f"oc apply -f {current_path}/{file}")
+        print(f'create_operator({error}): {response}')
+
+    @keyword
+    def get_redhat_openshift_pipelines(self):
+        """
+        when creating at the first time, it can take like 1 minute to have the pods ready
+        """
+        deployment_count = 0
+        count = 0
+        while deployment_count != 1 and count < 30:
+            deployments = []
+            response, _ = self.run_oc(f"oc get deployment -n openshift-operators openshift-pipelines-operator -o json")
+            try:
+                response = json.loads(response)
+                if response['metadata']['name'] == 'openshift-pipelines-operator' and 'readyReplicas' in response['status'] and response['status']['readyReplicas'] == 1:
+                    deployments.append(response)
+            except JSONDecodeError:
+                pass
+            deployment_count = len(deployments)
+            time.sleep(1)
+            count += 1
+        return self.count_running_pods(f'oc get pods -n openshift-operators -l name=openshift-pipelines-operator -o json', 'openshift-pipelines-operator', 'Running', 1)
+
+    @keyword
+    def login_using_user_and_password(self, user, pwd, project):
+        print("Fetch token")
+        basic_value = f'{user}:{pwd}'.encode('ASCII')
+        basic_value = base64.b64encode(basic_value).decode('ASCII')
+        response = requests.get(self.auth_url, headers={
+            'Authorization': f'Basic {basic_value}'
+        }, verify=False, allow_redirects=False)
+        url_with_token = response.headers['Location']
+        access_token_key = 'access_token='
+        idx_token = url_with_token.index(access_token_key) + len(access_token_key)
+        token = url_with_token[idx_token:]
+        idx_token = token.index('&')
+        self.sa_token = token[:idx_token]
+
+        print("Fetch the dsp route")
+        self.route = ''
+        count = 0
+        while self.route == '' and count < 60:
+            self.route, _ = self.run_oc(f"oc get route -n {project} ds-pipeline-ui-sample --template={{{{.spec.host}}}}")
+            time.sleep(1)
+            count += 1
+
+        print(f"DSP route should be working: {self.route}")
+        status = -1
+        count = 0
+        while status != 200 and count < 30:
+            if count > 0:
+                print(f'({count}): HTTP Status: {status}')
+                time.sleep(1)
+            response, status = self.do_get(f'https://{self.route}/apis/v1beta1/runs', headers={
+                'Authorization': f'Bearer {self.sa_token}'
+            })
+            count += 1
+        return status
+
+    @keyword
+    def remove_pipeline_project(self, project):
+        print(f"We are removing the project({project}) because we could run the test multiple times")
+        self.run_oc(f"oc delete project {project} --wait=true --force=true")
+        print("Wait because it could be in Terminating status")
+        count = 0
+        while count < 30:
+            project_status, error = self.run_oc(f"oc get project {project} --template={{{{.status.phase}}}}")
+            print(f"Project status: {project_status}")
+            print(f"Error message: {error}")
+            if project_status == '':
+                break
+            time.sleep(1)
+            count += 1
+        self.run_oc(f"oc new-project {project}")
+
+    @keyword
+    def create_pipeline_project(self, project):
+        self.run_oc(f"oc new-project {project}")
+
+    @keyword
+    def create_pipeline(self, url_test_pipeline_run_yaml):
+        print("Creating a pipeline from data science pipelines stack")
+        test_pipeline_run_yaml, _ = self.do_get(url_test_pipeline_run_yaml)
+        filename = 'test_pipeline_run_yaml.yaml'
+        with open(filename, 'w', encoding='utf-8') as f:
+            f.write(test_pipeline_run_yaml)
+        with open(filename, 'rb') as f:
+            response, _ = self.do_upload(f'https://{self.route}/apis/v1beta1/pipelines/upload', files={
+                    'uploadfile': f
+                }, headers={
+                    'Authorization': f'Bearer {self.sa_token}'
+                })
+        os.remove(filename)
+        pipeline_json = json.loads(response)
+        pipeline_id = pipeline_json['id']
+        response, status = self.do_get(f'https://{self.route}/apis/v1beta1/pipelines/{pipeline_id}', headers={
+            'Authorization': f'Bearer {self.sa_token}'
+        })
+        assert status == 200
+        assert json.loads(response)['name'] == filename
+        return pipeline_id
+
+    @keyword
+    def create_run(self, pipeline_id):
+        print("Creating the run from uploaded pipeline")
+        response, status = self.do_post(f'https://{self.route}/apis/v1beta1/runs', headers={
+                'Authorization': f'Bearer {self.sa_token}',
+                'Content-Type': 'application/json'
+            }, json={
+                "name": "test-pipeline-run",
+                "pipeline_spec": {
+                    "pipeline_id": f"{pipeline_id}"
+            }
+        })
+        assert status == 200
+        run_json = json.loads(response)
+        run_id = run_json['run']['id']
+
+        response, status = self.do_get(f'https://{self.route}/apis/v1beta1/runs/{run_id}', headers={
+            'Authorization': f'Bearer {self.sa_token}'
+        })
+        assert status == 200
+
+        return run_id
+
+    @keyword
+    def check_run_status(self, run_id):
+        run_status = None
+        count = 0
+        while run_status != 'Completed' and count < 160:
+            print(f"Checking run status: {run_status}")
+            response, status = self.do_get(f'https://{self.route}/apis/v1beta1/runs/{run_id}', headers={
+                'Authorization': f'Bearer {self.sa_token}'
+            })
+            try:
+                run_json = json.loads(response)
+                if 'run' in run_json and 'status' in run_json['run']:
+                    run_status = run_json['run']['status']
+            except JSONDecodeError:
+                print(response, status)
+                pass
+            time.sleep(1)
+            count += 1
+        return run_status
+
+    @keyword
+    def delete_runs(self, run_id):
+        print('Deleting the runs')
+
+        response, status = self.do_delete(f'https://{self.route}/apis/v1beta1/runs/{run_id}', headers={
+            'Authorization': f'Bearer {self.sa_token}'
+        })
+        assert status == 200
+        response, status = self.do_get(f'https://{self.route}/apis/v1beta1/runs/{run_id}', headers={
+            'Authorization': f'Bearer {self.sa_token}'
+        })
+        assert status == 404
+
+    @keyword
+    def delete_pipeline(self, pipeline_id):
+        print("Deleting the pipeline")
+        response, status = self.do_delete(f'https://{self.route}/apis/v1beta1/pipelines/{pipeline_id}', headers={
+            'Authorization': f'Bearer {self.sa_token}'
+        })
+        assert status == 200
+        response, status = self.do_get(f'https://{self.route}/apis/v1beta1/pipelines/{pipeline_id}', headers={
+            'Authorization': f'Bearer {self.sa_token}'
+        })
+        assert status == 404
+
+    def count_running_pods(self, oc_command, name_startswith, status_phase, pod_criteria):
+        pod_count = 0
+        count = 0
+        while pod_count != pod_criteria and count < 30:
+            pods = []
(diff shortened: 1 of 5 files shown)


libs/OpenShiftPipelineAPI.py
```diff
@@ -0,0 +1,146 @@
...
+            })
+            time.sleep(1)
+            print(status)
+        assert status == 200
+        return status
+
+    @keyword
+    def create_pipeline(self):
+        print("Creating a pipeline from data science pipelines stack")
+        test_pipeline_run_yaml, _ = self.do_get('https://raw.githubusercontent.com/opendatahub-io/data-science-pipelines-operator/main/tests/resources/dsp-operator/test-pipeline-run.yaml')
```
jgarciao, line 37: url of the pipeline to load should be a parameter

libs/OpenShiftPipelineAPI.py
```diff
@@ -0,0 +1,146 @@
...
+
+class OpenShiftPipelineAPI:
+
+    def __init__(self):
+        self.project = "ods-ci-pipeline"
+        self.route = ''
+        self.sa_token = None
+
+    @keyword
+    def fetch_runs(self):
```
jgarciao, line 18: I think this method should be renamed to "authenticate" or "login", as is the only one setting the credentials and is really not returning a list of runs

How are you going to test with different user credentials? (This is probably something to discuss in a meeting and not here)

tests/Resources/Files/kubeflow-operator.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kfdef.apps.kubeflow.org/v1
+kind: KfDef
+metadata:
+   name: data-science-pipelines-operator
```
jgarciao, line 4: This file should be called data-science-pipelines-operator.yaml, not kubeflow-operator.yaml

tests/Tests/x__pipelines/x__data-science-pipelines-api.robot
```diff
@@ -0,0 +1,46 @@
...
+Resource            ../../Resources/ODS.robot
+Resource            ../../Resources/Common.robot
+Library             DateTime
+Library             ../../../libs/OpenShiftPipelineAPI.py
+Suite Setup         RHOSi Setup
+Suite Teardown      RHOSi Teardown
+
+
+*** Variables ***
+${KUBEFLOW_OPERATOR_YAML}=                     tests/Resources/Files/kubeflow-operator.yaml
```
jgarciao, line 13: Rename to ` ${DATA_SCIENCE_PIPELINES_OPERATOR}`

tests/Resources/Files/data-science-pipelines.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: datasciencepipelinesapplications.opendatahub.io/v1alpha1
+kind: DataSciencePipelinesApplication
```
jgarciao, line 2: This file could be renamed to data-science-pipelines-sample.yaml

libs/OpenShiftPipelineAPI.py
```diff
@@ -0,0 +1,146 @@
+import subprocess
+from json import JSONDecodeError
+import requests
+import os
+import json
+import time
+from robotlibcore import keyword
+
+
+class OpenShiftPipelineAPI:
```
jgarciao, line 10: I think this class and the file should be called DataSciencePipelinesAPI

tests/Tests/x__pipelines/x__data-science-pipelines-api.robot
```diff
@@ -0,0 +1,46 @@
...
+Install DataSciencePipelinesApplication CRD
+    [Documentation]    Install and verifies that DataSciencePipelinesApplication CRD is installed and working
+    [Tags]    Smoke
+    Run    oc new-project ods-ci-pipeline
+    Oc Apply    kind=DataSciencePipelinesApplication    src=${DATA_SCIENCE_PIPELINES_APPLICATION_YAML}     namespace=ods-ci-pipeline
+    ${generation_value}    Run    oc get datasciencepipelinesapplications -o json | jq '.items[0].metadata.generation'
+    Should Be True    ${generation_value} == 1    DataSciencePipelinesApplication created
+
+
+Test and end-to-end pipeline workflow via API
```
jgarciao, line 34: Tests should not be dependent on the execution of the previous one and should be repeatable (deletion of resources can be done in Test Teardown/Suite teardown)

tests/Tests/x__pipelines/x__data-science-pipelines-api.robot
```diff
@@ -0,0 +1,46 @@
...
+    [Documentation]    Create, run and double check the pipeline result. In the end, clean the pipeline resources.
+    [Tags]    Smoke
```
jgarciao, line 36: This test should be in Tier1 and Sanity  (will be in Smoke once is stable). Missing Polarion ID. 

Also, it could be refactored with keywords to simplify the first level of abstraction (see https://docs.google.com/document/d/11ZJOPI1uq-0Wl6a2V8fkAv_TQhfzp9t_IjXAheaJxmQ/edit#heading=h.7gognio1srsr and we can discuss in a meeting too)

jgarciao (changes requested):

bdattoma (changes requested):

ods_ci/libs/DataSciencePipelinesAPI.py
```diff
@@ -0,0 +1,235 @@
...
+            deployments = []
+            response, _ = self.run_oc(f"oc get deployment -n openshift-operators openshift-pipelines-operator -o json")
+            try:
+                response = json.loads(response)
+                if response['metadata']['name'] == 'openshift-pipelines-operator' and 'readyReplicas' in response['status'] and response['status']['readyReplicas'] == 1:
+                    deployments.append(response)
+            except JSONDecodeError:
+                pass
+            deployment_count = len(deployments)
+            time.sleep(1)
```
bdattoma, line 42: maybe you could use [oc wait](https://docs.openshift.com/container-platform/4.8/cli_reference/openshift_cli/developer-cli-commands.html#oc-wait) to spare a bit of lines

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
```diff
@@ -0,0 +1,60 @@
...
+    [Documentation]    Create, run and double check the pipeline result. In the end, clean the pipeline resources.
+    [Arguments]     ${username}    ${password}    ${project}
+    Create Pipeline Project    ${project}
+    Install DataSciencePipelinesApplication CR    ${project}
+    ${status}    Login Using User And Password    ${username}    ${password}    ${project}
+    Should Be True    ${status} == 200    DSP routing is working
+    ${pipeline_id}    Create Pipeline    ${URL_TEST_PIPELINE_RUN_YAML}
+    ${run_id}    Create Run    ${pipeline_id}
+    ${run_status}    Check Run Status    ${run_id}
+    Should Be True    '${run_status}' == 'Completed'    Run ends
```
bdattoma, line 38: would it make sense to check the pipeline output, if there's any output? These lines are checking the status only, which is ok, but if there is an expected result of the pipeline (e.g., a number, a text, whatever) we may want to check it too.

bdattoma (dismissed):
lgtm, left some comments

ods_ci/libs/DataSciencePipelinesAPI.py
```diff
@@ -0,0 +1,235 @@
...
+            })
+            time.sleep(1)
+            count += 1
+            print(status)
+        return status
+
+    @keyword
+    def create_pipeline_project(self, project):
+        print(f"We are removing the project({project}) because we could run the test multiple times")
+        self.run_oc(f"oc delete project {project} --wait=true --force=true")
```
bdattoma, line 86: wouldn't be safer to delete all the pipelines objects before deleting forcing the deletion of the project? It would ensure there are no dangling resources in the test cluster

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
```diff
@@ -0,0 +1,60 @@
...
+    Delete Runs    ${run_id}
+    Delete Pipeline    ${pipeline_id}
```
bdattoma, line 40: it would be safer to have a Teardown running the deletion, because the teardown is executed even if some of the test keywords fail. While, in this way, if the "Should Be True"  fails, the deletion is not run

ods_ci/libs/DataSciencePipelinesAPI.py
diegolovison: I believe we should investigate what resources are "running" that is preventing the project to be deleted. Also, removing all pipelines before removing the project seems like: "before removing the database we need to drop all tables"

ods_ci/libs/DataSciencePipelinesAPI.py
diegolovison: I believe `oc wait` doesn't support nested AND or OR
also, https://github.com/kubernetes/kubernetes/issues/83242

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
diegolovison: It will keep running the `Check Run Status`

```pyton
        while run_status != 'Completed':
            print(f"Checking run status: {run_status}")

```

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
diegolovison: How can I invoke Teardown with parameters?

ods_ci/libs/DataSciencePipelinesAPI.py
bdattoma: Probably you can avoid checking "response['metadata']['name'] == 'openshift-pipelines-operator'" because you're already getting it. So you may not need nested AND or OR.

Anyway, just a suggestion, I think it's okay what you implemented

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
bdattoma: okay, but what about the produced results of the pipeline? The pipelines runs some code doing something, does that have an output to the user, rather than the pipeline status.

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
```diff
@@ -0,0 +1,70 @@
...
+Resource            ../../../Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+Library             DateTime
+Library             ../../../../libs/DataSciencePipelinesAPI.py
+Suite Setup         Data Science Pipelines Suite Setup
+Suite Teardown      RHOSi Teardown
+
+
+*** Variables ***
+${REDHAT_OPENSHIFT_PIPELINES_YAML}=            tests/Resources/Files/redhat-openshift-pipelines.yaml
+${DATA_SCIENCE_PIPELINES_APPLICATION_YAML}=    tests/Resources/Files/data-science-pipelines-sample.yaml
```
lugi0, line 15: These need to be updated to `ods_ci/tests/Resources/...` in order to work when running ods_ci as `./ods_ci/run_robot_tests.sh`

ods_ci/run_robot_test.sh
```diff
@@ -288,9 +288,18 @@ if command -v yq &> /dev/null
...
   poetry install
+  echo "path after install"
+  python3 -c 'import sys;print(sys.executable)'
 fi
+echo "path before source activate"
+python3 -c 'import sys;print(sys.executable)'
 source $(poetry env info --path)/bin/activate
+echo "path after source activate"
+python3 -c 'import sys;print(sys.executable)'
+
```
lugi0, line 302: This has to be removed before merging

lugi0 (changes requested):

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
bdattoma: you can either:
- pass the parameter as a normal keyword when you declare the teardown, like:
```
My Test 1
      [Setup] Setup Keyword
      ${my_pipeline}=   Create Pipeline
      Second Keyword
      ....
      [Teardown]  Clean Up pipeline_to_delete=${my_pipeline}
```
- set some Suite Variables to use in the teardown keyword (I usually do this when using Suite Teardown instead of Test Teardown)

ods_ci/libs/DataSciencePipelinesAPI.py
```diff
@@ -0,0 +1,242 @@
...
+
+class DataSciencePipelinesAPI:
+
+    def __init__(self):
+        self.auth_url = self.retrieve_auth_url()
+        self.route = ''
+        self.sa_token = None
+
+    @keyword
+    def create_operator(self, file):
```
jgarciao, line 19: As a suggestion for a future PR, this method could be implemented as regular Robot Framework keyword using the robotframework-openshift library.

ods_ci/libs/DataSciencePipelinesAPI.py
```diff
@@ -0,0 +1,242 @@
...
+                if response['metadata']['name'] == 'openshift-pipelines-operator' and 'readyReplicas' in response['status'] and response['status']['readyReplicas'] == 1:
+                    deployments.append(response)
+            except JSONDecodeError:
+                pass
+            deployment_count = len(deployments)
+            time.sleep(1)
+        return self.count_running_pods(f'oc get pods -n openshift-operators -l name=openshift-pipelines-operator -o json', 'openshift-pipelines-operator', 'Running', 1)
+
+    @keyword
+    def get_data_science_pipelines_operator_controller_manager(self):
```
jgarciao, line 46: It should be renamed as `Wait Until Data Science Pipelines Controller Manager Is Deployed`
As a suggestion for a future PR, this method could be implemented as regular Robot Framework keyword using the robotframework-openshift library.

ods_ci/libs/DataSciencePipelinesAPI.py
```diff
@@ -0,0 +1,242 @@
...
+        return pipeline_id
+
+    @keyword
+    def create_run(self, pipeline_id):
+        print("Creating the run from uploaded pipeline")
+        response, status = self.do_post(f'https://{self.route}/apis/v1beta1/runs', headers={
+                'Authorization': f'Bearer {self.sa_token}',
+                'Content-Type': 'application/json'
+            }, json={
+                "name": "test-pipeline-run",
```
jgarciao, line 134: "test-pipeline-run": Should be a parameter, in order to be able to create multiple runs for the same user (also to test behavior of different run names)

jgarciao (approved):

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot
jgarciao: This is done

jgarciao (approved):
LGTM (leaving optional improvements for a future PR)

jgarciao:
Diego tested this code (successfully) here: https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1705/console

bdattoma (approved):

tests/Tests/x__pipelines/x__data-science-pipelines-api.robot
jgarciao: This was done

PR #726, opened 2023-03-23
add dependency install skip flag

lugi0:


ods_ci/run_robot_test.sh
```diff
@@ -287,7 +287,9 @@ if command -v yq &> /dev/null
         echo "we did not find yq, so not trying the oc login"
 fi
 
-poetry install
+if [[ ${SKIP_INSTALL} -eq 0 ]]; then
+  poetry install
+fi
 source $(poetry env info --path)/bin/activate
```
bdattoma, line 293: what would happen if I skip install and I didn't run poetry in previous runs?
Instead, we could skip the env creation if SKIP_INSTALL is 1 and the venv folder doesn't exist. WDYT?

lugi0: If you skip the install and you have not run Poetry before, the activate step would fail, which I think is the correct behaviour given that we want the dependencies to be installed.
This flag will be used only in known scenarios where we known Poetry doesn't need to run again IMHO, so I don't see the problem in leaving it as it is right now. We could certainly try to be a bit smarter about the venv being present or not, but then again the venv folder existing does not guarantee that all dependencies have been installed

bdattoma (approved):

PR #741, opened 2023-04-07
Fix DS project Sanity test for 1.25

bdattoma:
Keywords to check if workbenches, storages and data connection got broken because of a change in the HTML (from `<h3> `to `<h4>`). 
This PR wants to fix it and make the xpath a bit more robust against the change of h type


lugi0 (approved):
I think that instead of doing `/*/div` you could do `//div` directly, but LGTM regardless

jgarciao (approved):

bdattoma:
Jenkins Job against RHODS v1.25: [1661](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1661/console) _RUNNING_

bdattoma:
> I think that instead of doing `/*/div` you could do `//div` directly, but LGTM regardless

true, I think it should work. If we see this failing, we could switch to your suggestion

PR #743, opened 2023-04-11
fix for robocop action

tarukumar:
`could-be-forced-tags` rule has been removed and replaced by "could-be-test-tags" from the robocop 3.0.0 . Which was causing the robocop GitHub action to fail.


aloganat (approved):
LGTM

mwaykole (approved):

PR #754, opened 2023-04-18
RHODS-8014 Improve understanding of poetry activate bash script

diegolovison:
https://issues.redhat.com/browse/RHODS-8014

### DIFF
--- a/ods_ci/run_robot_test.sh
+++ b/ods_ci/run_robot_test.sh
@@ -287,10 +287,17 @@ if command -v yq &> /dev/null
         echo "we did not find yq, so not trying the oc login"
 fi
 
+# Activating the virtual environment
+# python3 -c help you to understand what is going on with environments
 if [[ ${SKIP_INSTALL} -eq 0 ]]; then
+  python3 -c 'import sys;print(sys.executable)'
+  poetry env use 3.9
+  source $(poetry env info --path)/bin/activate
+  python3 -c 'import sys;print(sys.executable)'
   poetry install
+else
+  echo "poetry install skipped"
 fi
-source $(poetry env info --path)/bin/activate
 
 #Create a unique directory to store the output for current test run
 if [[ ! -d "${TEST_ARTIFACT_DIR}" ]]; then


ods_ci/run_robot_test.sh
```diff
@@ -287,10 +287,17 @@ if command -v yq &> /dev/null
         echo "we did not find yq, so not trying the oc login"
 fi
 
+# Activating the virtual environment
+# python3 -c help you to understand what is going on with environments
 if [[ ${SKIP_INSTALL} -eq 0 ]]; then
+  python3 -c 'import sys;print(sys.executable)'
+  poetry env use 3.9
```
lugi0, line 294: why would you hardcode python 3.9 here?

ods_ci/run_robot_test.sh
```diff
@@ -287,10 +287,17 @@ if command -v yq &> /dev/null
...
+  source $(poetry env info --path)/bin/activate
```
lugi0, line 295: this has to go outside the IF block, otherwise skipping the install will also not activate the python virtual environment

ods_ci/run_robot_test.sh
```diff
@@ -287,10 +287,17 @@ if command -v yq &> /dev/null
         echo "we did not find yq, so not trying the oc login"
 fi
 
+# Activating the virtual environment
+# python3 -c help you to understand what is going on with environments
```
lugi0, line 291: You can just run `poetry env list` and `poetry env info` to get all the information you need

bdattoma:
@diegolovison I'm closing this old PR since the original ticket in the description has been closed. If you think we still need these changes, please re-open it

ods_ci/run_robot_test.sh
jiridanek: I wouldn't do it here, but I can imagine hardcoding singular Python version in `pyproject.toml`. Same as dependency pinning, this will might help prevent some inconsistencies.

bdattoma: @jiridanek the .toml file has already a python version pinned

jiridanek: @bdattoma no, not really

https://github.com/red-hat-data-services/ods-ci/blob/afa1de64ba5e632e794d70eebb4197956270e419/pyproject.toml#L9-L10

That means "use Python 3.8.1 or anything <4.0.0 that is also newer than 3.8.1", https://python-poetry.org/docs/dependency-specification/#caret-requirements

That's not much of a constrain at all, since I don't expect Python 4 to come any time soon.

bdattoma: you're right! Anyways, if we want to pin to a specific version, it should be done in the .toml

jiridanek: I'll propose a pin to Python 3.11.x, in the toml. It seems to me the most sensible, sufficiently available, and widely supported version, lets see how it goes.

I'm afraid it's going to be more trouble than it's worth, because it forces everybody (and their CI) to adjust, but, on the other hand, I like unifying infra that can be unified. I might leave this for a quiet period after a release, though.

bdattoma: sounds good. Do you mind raising a PR? if you don't have bandwitdh, raise the point in the automation sync up Agenda and we'll find someone to work on it

jiridanek: all right, I'll do it; it's a good opportunity to create a CI MR, which I did not do yet for ODH

* https://github.com/red-hat-data-services/ods-ci/pull/1118

PR #761, opened 2023-04-28
Upgrade ODS-CI image to centos stream 9

bdattoma:
Image upgrade comes with additional changes in the Dockerfile:
1. replace `lsb_release` with python pkg `distro` ([details](https://access.redhat.com/solutions/6973382))
2. upgrade to python 3.9 (installed by default in centos stream 9)
3. upgrade fedora `epel` to avoid issue with chromium installation

Plus, 3 more changes not related to OS upgrade:
1. add security hotspot fix as suggested by SonarCloud in the cURL steps
2. join 2 RUN steps into one single one ([docker best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers))
3. fix the argument `--skip-pip-install` in `run.sh` script which was preventing the scripts to run

### DIFF
--- a/ods_ci/build/Dockerfile
+++ b/ods_ci/build/Dockerfile
@@ -14,21 +14,16 @@ ARG OC_VERSION=4.10
 ARG OC_CHANNEL=stable
 
 
-RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm &&\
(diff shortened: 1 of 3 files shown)


manosnoam:
lgtm

manosnoam (approved):
lgtm

PR #772, opened 2023-05-11
Improve uninstallation process (waiting until a project is deleted)

tarukumar:


ods_ci/tasks/Resources/RHODS_OLM/uninstall/oc_uninstall.robot
```diff
@@ -30,5 +30,14 @@ Verify Project Does Not Exists
   ${project_exists}=  Run Keyword and return status
   ...  Oc Get  kind=Namespace  field_selector=metadata.name=${project}
   IF  ${project_exists}
-  ...  wait until project does not exists  ${project}  3600
```
bdattoma, line 33: I think we should also fix this keyword in the openshiftlibrary

bdattoma (approved):

PR #778, opened 2023-05-15
Remove ProductBug from ODS-909 and ODS-507

bdattoma:
2 Test Cases which run as part of our known-issues pipeline are actually passing. This PR is removing the ProductBug tag from:
- ODS-909
- ODS-507


lugi0 (approved):

bdattoma:
dry run failure seems related to DS Pipelines test suites. @diegolovison investigating it. I don't think it should block this PR

tarukumar (approved):
/lgtm

PR #787, opened 2023-05-19
Set RHODS_OSD_INSTALL_REPO directory as a suite variable

manosnoam:
Set a suite variable OLM_DIR = rhodsolm
for the local directory name of RHODS_OSD_INSTALL_REPO 
(The Gitlab repo data-hub/olminstall)

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -1,6 +1,7 @@
 *** Settings ***
 Library    String
 Library    OpenShiftLibrary
+
 *** Keywords ***
 Install RHODS
   [Arguments]  ${cluster_type}     ${image_url}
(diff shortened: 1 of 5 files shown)


ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -64,20 +65,19 @@ Verify Builds In redhat-ods-applications
 
 Clone OLM Install Repo
    [Documentation]   Clone OLM git repo
-   ${return_code}	  Run And Return Rc    git clone ${RHODS_OSD_INSTALL_REPO} rhodsolm
+   ${return_code}	  Run And Return Rc    git clone ${RHODS_OSD_INSTALL_REPO} ${EXECDIR}/${OLM_DIR}
    Should Be Equal As Integers	${return_code}	 0
-   Set Test Variable     ${filename}    rhodsolm
```
manosnoam, line 69: @lugi0 take a look at the root cause - We're setting the test variable "filename" here:
1) "filename" for test variable that is being used other places is unmeaning-full and wrong name for global variable that is being called in many places.
2) It is the directory name of the local clone of the repo data-hub/olminstall - it should not be called "filename".
3) Since it was set here, and not globaly as suite variable, it is not beeing invoked when "Clone OLM Install Repo" task is skipped - when `${is_operator_installed} =  Is RHODS Installed` returns true (which can and have happened on env where RHODS was partly installed).

manosnoam:
> The only changes I see are formatting and using the new `${OLM_DIR}` variable, I'm not sure this fixes the issue encountered before opening the PR. I'd suggest implementing an actual fix before merging.

var namethe only change, see my comment bellow.
But even if it was the only change - fixing a

> After discussing with Noam offline I understand the change that is being applied here, and it LGTM. Will wait for @tarukumar 's opinion as well before merging.

Here's the latest job run:
http://10.0.146.152:8080/job/rhods-ci-pr-test-hive/24/

tarukumar (approved):
Approving the pr . my comment are present in slack

jgarciao (approved):
LGTM: what I understood in the automation meeting and looking at the code is that this PR is making it more legible and it's been tested successfully. Further improvements in future PRs

PR #790, opened 2023-05-22
Increase the timeout to double check if Pipeline API is working

diegolovison:


jgarciao (approved):

ldimaggi:
One thing to be careful about regarding timeouts - let's not mask product bugs and or regressions by extending timeouts.

In this case - are we seeing a regression in throughput that is requiring the test to have a longer timeout? Thx!

PR #799, opened 2023-05-30
New common keyword 'Run And Watch Command'

manosnoam:
Add a new keyword in Common.robot to run any command and watch its output.

This is extremely useful for those RF keywords that calls a sub-process
which runs more than few seconds, like calling an external shell script.
The 'Run And Watch Command' keyword arguments are:
- Full command including args (single String).
- Optional: Timeout in minutes: 10 by default.
- Optional: An expected text: Verify a search phrase to exist in command output.


manosnoam:
Here's a test where this keyword was used:
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/cluster-operations/23/console

You can see that instead of just printing:
Uninstalling RHODS operator in selfmanaged
[Waiting for ever without any indcation what's going on]

Now the output, which is printed in **realtime**,
without the need to wait for the long process of RHODS uninstall to finish:

```
Rhods Olm :: Perform and verify RHODS OLM tasks                               
Can Uninstall RHODS Operator                                          Cluster type: selfmanaged
Uninstalling RHODS operator in selfmanaged
Watching command output: cd /home/jenkins/workspace/cluster-operations/23/ods-ci/rhodsolm && ./cleanup.sh -t operator
Shell process started in the background
.Removing RHODS... Be patient...
.No resources found
kfdef.kfdef.apps.kubeflow.org "modelmesh-monitoring" deleted
kfdef.kfdef.apps.kubeflow.org "monitoring" deleted
............................................Unable to connect to the server: EOF
| FAIL |
Error occured while running: cd /home/jenkins/workspace/cluster-operations/23/ods-ci/rhodsolm && ./cleanup.sh -t operator: 1 != 0
------------------------------------------------------------------------------
Rhods Olm :: Perform and verify RHODS OLM tasks                       | FAIL |
1 task, 0 passed, 1 failed
```

ods_ci/tests/Resources/Common.robot
```diff
@@ -305,3 +305,38 @@ Extract URLs From Text
...
+Run And Watch Command
+  [Documentation]    Run any shell command (including args) with optional:
+  ...    Timeout in minutes: 10 by default.
+  ...    Expected text: Verify the text exists in command output.
+  [Arguments]    ${command}    ${timeout_min}=10    ${expected_text}=${NONE}
+  Log    Watching command output: ${command}   console=True
+  @{args} =    Split String    ${command}
+  ${process_log} =    Set Variable    ${OUTPUT DIR}/${args}[0].log
+  Create File    ${process_log}
+  Create File    ${process_log}.old
```
tarukumar, line 318: what is the idea behind old and new log file?

ods_ci/tests/Resources/Common.robot
```diff
@@ -305,3 +305,38 @@ Extract URLs From Text
...
     RETURN    ${urls}
+
+Run And Watch Command
+  [Documentation]    Run any shell command (including args) with optional:
+  ...    Timeout in minutes: 10 by default.
+  ...    Expected text: Verify the text exists in command output.
+  [Arguments]    ${command}    ${timeout_min}=10    ${expected_text}=${NONE}
+  Log    Watching command output: ${command}   console=True
+  @{args} =    Split String    ${command}
+  ${process_log} =    Set Variable    ${OUTPUT DIR}/${args}[0].log
```
tarukumar, line 316: I would recommend against using the log name "cd.log" as it is not considered a good practice. Instead, I suggest using a more descriptive and relevant name, such as "uninstall.log" or something similar. Using meaningful and contextual log names helps improve readability and clarity when reviewing or analyzing log file.

Also, If you agree with above suggestion i don't think there is a need of Split string in line no 315

ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
```diff
@@ -24,34 +28,32 @@ Uninstall RHODS In OSD
...
-   ${return_code}    ${output}    Run And Return Rc And Output   oc create configmap delete-self-managed-odh -n redhat-ods-operator
-   Should Be Equal As Integers	${return_code}	 0   msg=Error creation deletion configmap
-   ${return_code}    ${output}    Run And Return Rc And Output   oc label configmap/delete-self-managed-odh api.openshift.com/addon-managed-odh-delete=true -n redhat-ods-operator
-   Should Be Equal As Integers	${return_code}	 0   msg=Error observed while adding label to configmap
-   Verify Project Does Not Exists  redhat-ods-applications
-   Verify Project Does Not Exists  redhat-ods-monitoring
-   Verify Project Does Not Exists  rhods-notebooks
-   ${return_code}    ${output}    Run And Return Rc And Output   oc delete namespace redhat-ods-operator
+  [Documentation]   Uninstall rhods on self-managed cluster for operatorhub installtion
+  ${return_code}    ${output}    Run And Return Rc And Output   oc create configmap delete-self-managed-odh -n redhat-ods-operator
```
bdattoma, line 52: wouldn't we need to put `Run and Watch Command` here too?

ods_ci/tests/Resources/Common.robot
```diff
@@ -305,3 +305,38 @@ Extract URLs From Text
...
+  Create File    ${process_log}
+  Create File    ${process_log}.old
+  ${process} =    Start Process    ${command}    shell=True    stdout=${process_log}    stderr=STDOUT    # robocop: disable
+  Log    Shell process started in the background   console=True
+  ${timeout_result} =    Wait Until Keyword Succeeds    ${timeout_min} min    10 s
+  ...    Read Command Log    ${process}    ${process_log}
```
bdattoma, line 322: what is it doing? does it wait until the first log line is written?

bdattoma: oh it is kind of "wait until the process finished to write logs". Is that correct? 
If that so, I think the Wait Until Kw Succeeds is not necessary here, unless there is the chance it fails when the file is empty, hmm

bdattoma: I think the set of naming here can be misleading, because I wouldn't expect a kw named `Read Command Log` to also to assess if the process is ended

ods_ci/tests/Resources/Common.robot
manosnoam: Good question: We need to direct stdout and stderr to external log (new file), but each few seconds track the new lines appended to the file. For that I added old file that is a copy of new file - before new data was appended to the new file (that is constantly getting proccess output). Each interval I compare between old file and new file data, and only print the diff into console in real time.

lugi0 (dismissed):

ods_ci/tests/Resources/Common.robot
tarukumar: I think it better you directly monitor the process output and log to console instead of open file and write to it.That would be easy path to go and also save some time might be in second

ods_ci/tests/Resources/Common.robot
tarukumar: @manosnoam let me know you though on it

ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
manosnoam: We should defently start to use "Run and Watch Command" everywhere in RF if a command takes more than few seconds. 
Is that the case here ?

ods_ci/tests/Resources/Common.robot
```diff
@@ -305,3 +305,38 @@ Extract URLs From Text
...
+  ${proc_result} =	    Wait For Process    ${process}    timeout=3 secs
```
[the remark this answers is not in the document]
manosnoam: Since the previous step was "Wait Until Keyword Succeeds" with a timeout of 10 minutes by default, I excpect the command to end before 10 minutes passed. 
Then I also give it a "gesture" of 3 more seconds to complete, as it has already passed the allowed timeout, or ended before the timeout. But it will return the process results, and that's what matters. to evaluate.

ods_ci/tests/Resources/Common.robot
manosnoam: Makes sense! I changed it to set the log name according to the test name.

ods_ci/tests/Resources/Common.robot
tarukumar: You don't need multiple file ..you can monitor the out just via alias and print the out put
` alias= prep
  ${result}   Wait For Process   prep   timeout=10 secs
  ${status}   Is Process Running   prep`

ods_ci/tests/Resources/Common.robot
manosnoam: ack, I renamed "Read Command Log" to "Check Process Output and Status"

ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
manosnoam: I've also added it to the RHODS CLI install test, here's an an output example:
http://10.0.146.152:8080/job/rhods-ci-pr-test/12/console

bdattoma: > Is that the case here ?

hm probably not

bdattoma (commented):
lgtm, i'll wait until the other conversations are closed to send the approval :)

ods_ci/tests/Resources/Common.robot
manosnoam: I'm not sure I'm following your suggestion how to compare stdout of external process, without using a reference to the stdout before the new stdout was sent to. 
Anyway, I have changed the logic to use a temporary file for the old stdout/stderr, and now keep just the latest process log in artifacts.

jgarciao (approved):

ods_ci/tests/Resources/Common.robot
```diff
@@ -305,3 +305,39 @@ Extract URLs From Text
     [Arguments]    ${text}
     ${urls}=     Get Regexp Matches   ${text}   (?:(?:(?:ftp|http)[s]*:\/\/|www\.)[^\.]+\.[^ \n]+)
     RETURN    ${urls}
+
+Run And Watch Command
+  [Documentation]    Run any shell command (including args) with optional:
+  ...    Timeout in minutes: 10 by default.
+  ...    Expected text: Verify the text exists in command output.
+  [Arguments]    ${command}    ${timeout_min}=10    ${expected_text}=${NONE}
```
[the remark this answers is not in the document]
manosnoam: Added your suggestions, thanks!

ods_ci/tests/Resources/Common.robot
```diff
@@ -305,3 +305,39 @@ Extract URLs From Text
...
+  Create File    ${temp_log}
+  ${process_id} =    Start Process    ${command}    shell=True    stdout=${process_log}    stderr=STDOUT    # robocop: disable
+  Log    Shell process started in the background   console=True
+  Wait Until Keyword Succeeds    ${timeout_min} min    10 s
+  ...    Check Process Output and Status    ${process_id}
+  ${proc_result} =	    Wait For Process    ${process_id}    timeout=3 secs
+  Terminate Process    ${process_id}    kill=true
+  Should Be Equal As Integers	    ${proc_result.rc}    0    msg=Error occured while running: ${command}
+  Should Contain    ${process_log}    ${expected_text}
+  RETURN    ${proc_result.rc}
```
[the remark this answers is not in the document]
manosnoam: The output could be huge, so I only return the exit code.
But the output is saved both into {TEST NAME}.log and in the RF output report.

ods_ci/tests/Resources/Common.robot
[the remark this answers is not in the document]
manosnoam: Tested again on https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1849/console

```
Can Install RHODS Operator                                            Cluster type: selfmanaged
Installing RHODS operator in selfmanaged
Watching command output: cd /home/jenkins/workspace/rhods-ci-pr-test/1849/ods-ci/ods_ci/rhodsolm && ./setup.sh -t operator -u beta -i brew.registry.redhat.io/rh-osbs/iib:470465
Shell process started in the background
..namespace/redhat-ods-operator created
Context "default/api-ods-qe-07-rhods-ccitredhat-com:6443/htpasswd-cluster-admin-user" modified.
catalogsource.operators.coreos.com/rhods-catalog-dev created
operatorgroup.operators.coreos.com/rhods-operator-dev created
subscription.operators.coreos.com/rhods-operator-dev created
```

@jgarciao , @bdattoma  can you review to merge ?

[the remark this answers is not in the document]
bdattoma: the new changes looks good to me. Only one thing I believe it's better to change, sorry I didn't notice before. Could you replace all the calls to `Run Keyword If` with an `IF` statement? that kw is deprecated and going to be removed in next RF upgrade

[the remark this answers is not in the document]
jgarciao: @manosnoam , in https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1850/console I see this error:
```
Multiple keywords with name 'Get RHODS Version' found. Give the full name of the keyword you want to use:
    InstalledOperators.Get RHODS version
    oc_version.Get RHODS Version
```

I believe it's not introduced in this PR but warning you just in case

jgarciao (approved):
LGTM (added a comment)

bdattoma (approved):

PR #809, opened 2023-06-02
Fix tekton test route

jgarciao:


diegolovison (changes requested):
Seems like we are renaming the variable instead of fixing.
It's late in Europe now, I am setting this as a Request change because we need to continue talking on Tue. This will prevent someone merging the PR

Thanks

jgarciao:
Currently the test is using the route [ds-pipeline-ui-sample](
https://github.com/red-hat-data-services/ods-ci/blob/93fd10fd71fdbe1e248a81e7adb23c233e457b93/ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/432__data-science-pipelines-tekton.robot#L36), which is the Kubeflow UI route, not the Pipelines backend route (it should be ds-pipeline-sample, without ui).

The main purpose of this PR is to fix that. Additionally, it is changing the name of the route, to be ds-pipeline-pipelines-definition instead of ds-pipeline-sample, in order to have the same naming than the pipelines created from the UI

jgarciao:
Closed in favor of #839

PR #816, opened 2023-06-08
Increase default timeout for Start Workbench

diegolovison:
The default timeout is to fast
image


bdattoma:
we probably need a way to check why the workbench didn't start in 30s, either bia CLI by checking the pod status or via UI by checking the event log under the "Starting.." status. 

However, I also think it could be a resources-related issue as @jgarciao mentioned. In which type of cluster you were facing the issue? @diegolovison

diegolovison:
Try to run the test locally on PSI

bdattoma:
> Try to run the test locally on PSI

then it could be a resources-related issue. Could you try on another one?

bdattoma:
@diegolovison I'm closing this PR because the timeout has been already increased in some other PR. Thanks

PR #817, opened 2023-06-09
Add poetry dependency to python 'distro' pkg

bdattoma:
In order to solve the error during our CI runs, we need to add the 'distro' package to our poetry dependency manager.
```
run_robot_test.sh: line 194: distro: command not found
```

the applied changes are:
1. add distro to poetry
2. move poetry install before calls to distro
3. remove yq from poetry

Notes on point 3:
- poetry installs the old version that mostly does not work for us
- there are few tests where the old version might be working and that could be replaced `jq`
- was not able to install the github version with poetry...can be further investigated
- this one can also be solved with jenkins nodes in containers

However, point 2 is not entirely right as choice because we should not install dependency before checking if the system is compatible with ods-ci (thx @tarukumar for the discussion). 
The PR does the work, but I would suggest not to merge it for the moment. I believe the issue might be solved when we'll get the Jenkins nodes in containers. 

**DO NOT MERGE FOR NOW**


bdattoma:
Jenkins run: [817](https://github.com/red-hat-data-services/ods-ci/pull/817) _SUCCESS_ but do not merge

bdattoma:
closing this one because it's not high in the priority queue, besides this is not causing issues.

PR #823, opened 2023-06-13
Update Poetry with latest packages

manosnoam:
The updated Poetry includes robotframework-openshift fix for Python 3.8:
https://github.com/red-hat-data-services/robotframework-openshift/pull/6


manosnoam:
For example, this job console shows the SyntaxWarning before the fix:
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1865/console

```
/home/centos/.cache/pypoetry/virtualenvs/ods-ci-urWBp4V3-py3.8/lib/python3.8/site-packages/OpenShiftLibrary/keywords/generic.py:379: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if str_idx is not '*' and str_idx is not '[':
```

mwaykole (approved):
lgtm

mwaykole:
if its verified could  u plz add a proper label

bdattoma (changes requested):
it would be better to keep the poetry toml file consistent with the changes here. That would also avoid replacing the pkgs version with the old ones when a poetry install is executed (after deleting the lock).

is that correct @lugi0 ?

jgarciao:
> it would be better to keep the poetry toml file consistent with the changes here. That would also avoid replacing the pkgs version with the old ones when a poetry install is executed (after deleting the lock).
> 
> is that correct @lugi0 ?

Yes, the change needs to be done in pyproject.toml and then update poetry.lock following the steps at https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/release.md

manosnoam:
> > it would be better to keep the poetry toml file consistent with the changes here. That would also avoid replacing the pkgs version with the old ones when a poetry install is executed (after deleting the lock).
> > is that correct @lugi0 ?
> 
> Yes, the change needs to be done in pyproject.toml and then update poetry.lock following the steps at https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/release.md The PR needs to update both files

@jgarciao , @bdattoma  I did not change pyproject.toml at all ;-) 
I only run `poetry update`with the **existing toml**, that built all packages with their latest allowed versions, include fetching latest robotframework-openshift.

The only thing I noticed is that pyproject.toml includes `robotframework = "^5"` instead of `robotframework = "^6"`as suggested in [readme](https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/release.md).

https://github.com/red-hat-data-services/ods-ci/blob/78a8f6707e9aada3869488e982426c1115a81b0e/pyproject.toml#LL12C18-L12C18

jgarciao (approved):
We cannot use robot framework 6 yet because it's not compatible with another dependency we are using (rpaframework)

I understand it now. I think the PR can be merged. Thanks  @manosnoam !

lugi0:
I believe @manosnoam is correct, @bdattoma if you were to delete your lock and then run a poetry install _today_, I'm almost certain you'd get the same lock file Noam is committing, since in this case you're simply getting the latest version(s) allowed by your pyproject.toml file.
The same thing would happen if you were to run `poetry updated` in your local env with the current toml file.

lugi0 (approved):

lugi0:
btw this is the issue for RF6 support in RPA: https://github.com/robocorp/rpaframework/issues/673
it's not likely they'll support it anytime soon

PR #833, opened 2023-06-16
Remove workaround for DS Project creation RHODS-8477

bdattoma:
In RHODS 1.28 the issue with user not being redirected to DS Project page has been fixed. Hence, it's worth to remove the workaround in the code (even if it was conditional)

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
@@ -103,6 +103,15 @@ Create Data Science Project
     Click Button    ${GENERIC_CREATE_BTN_XP}
     Wait Until Generic Modal Disappears
     Wait Until Project Is Open    project_title=${title}
+    ${open}=    Run Keyword And Return Status    Wait Until Project Is Open    project_title=${title}
+    IF    ${open} == ${FALSE}
+        ${is_home_open}=    Is Data Science Projects Page Open
+        IF    ${is_home_open} == ${TRUE}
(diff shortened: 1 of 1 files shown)


bdattoma:
test against 1.28: [1881](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1881/console) _RUNNING_

lugi0:
I'm ok with the change, but are we absolutely sure the issue has been fixed and won't appear ever again?
If it's possible we'll encounter it in the future, IMHO it'd be worth it to leave the conditional workaround in, with a Log message explicitly stating that the issue has reappeared if it's ever executed

bdattoma:
> I'm ok with the change, but are we absolutely sure the issue has been fixed and won't appear ever again? If it's possible we'll encounter it in the future, IMHO it'd be worth it to leave the conditional workaround in, with a Log message explicitly stating that the issue has reappeared if it's ever executed

the workaround was already conditional, but I fear like from now to 5 releases we'll stop looking for the warning and a possible regression would slip in with higher probability. That's the reason why I thought to remove it at all

lugi0:
We could force a failure in case the conditional block is entered to make sure it won't get missed

mwaykole (approved):

bdattoma:
new jenkins test after latest changes: [1884](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods-ci-pr-test/1884/TestReport/) _OK_ there are 2 failures: 1 is expected because there were other DS projects in the cluster for manual testing reasons. 1 happened for a delay in loading the web page (loading spinner can be seen in the screenshot)

jgarciao (approved):

nehachopra27 (approved):
LGTM

PR #836, opened 2023-06-20
fix SSL validation for model serving tests in self-managed clusters

lugi0:


diegolovison (changes requested):
You will need to implement a logic like: https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/libs/DataSciencePipelinesKfpTekton.py#L24-L39

First you should try, then if failed: https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/libs/DataSciencePipelinesKfpTekton.py#L42-L53

Maybe we could have a generic way to call the endpoint that requires a certificate and an unique way to retrieve the certificate by changing DataSciencePipelinesKfpTekton to use a different python code

lugi0:
@diegolovison rather than trying to assume the cert is valid and fetching the self-signed certificate only if that fails, I think it's better to use the `Is RHODS Self-Managed` keyword.
We expect the calls to fail only in that scenario, and this allows us to see a failure in case the connection appears to be insecure in other environments. Your logic works around this and never exposes the problem if it's encountered.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -209,13 +209,26 @@ Get Model Inference
...
+        ELSE
+            ${inference_output} =    Run    curl -ks ${url} -d ${INFERENCE_INPUT} -H "Authorization: Bearer ${token}"
+        END
     ELSE
-        ${inference_output} =    Run    curl -ks ${url} -d ${INFERENCE_INPUT}
+        IF  ${self_managed}==${TRUE}
+            Fetch Openshift CA Bundle
+            ${inference_output} =    Run    curl -s ${url} -d ${INFERENCE_INPUT} --cacert openshift_ca.crt
+        ELSE
+            ${inference_output} =    Run    curl -s ${url} -d ${INFERENCE_INPUT}
```
bdattoma, line 230: Probably this block of code could have been optimized in this way
```
${curl_cmd}=     Set Variable    curl -s ${url} -d ${INFERENCE_INPUT}
IF    ${token_auth}
        ${project_title}=    Get Model Project    ${model_name}
        ${token}=    Get Access Token via UI    ${project_title}
        ${curl_cmd}=     Catenate    ${cmd}    -H "Authorization: Bearer ${token}"
END
IF  ${self_managed}==${TRUE}
    Fetch Openshift CA Bundle
    ${curl_cmd}=     Catenate    ${cmd}    --cacert openshift_ca.crt
END
${inference_output} =    Run    ${curl_cmd}
```

jgarciao (approved):
LGTM (I like @bdattoma suggestion for this or a future PR. Also, I agree with Luca on checking if RHODS is Self-Managed)

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -119,6 +120,12 @@ Verify Tensorflow Model Via UI
     Run Keyword And Continue On Failure  Wait Until Keyword Succeeds  5 min  10 sec  Verify Serving Service
     Verify Model Status    ${MODEL_NAME}    success
     Set Suite Variable    ${MODEL_CREATED}    True
+    # If running on self-managed, fetch the CA bundle for the cluster
+    # so it can be used when making the inference request
+    ${self_managed} =    Is RHODS Self-Managed
+    IF  ${self_managed}==${TRUE}
+        Fetch Openshift CA Bundle
+    END
```
bdattoma, line 128: we usually avoid putting conditions in the test itself. Could this be moved outside? maybe in the setup? 
Alternatively, `Fetch Openshift CA Bundle` kw can be implemented in the helpers.py so can be used directly in the send_random_inference_request

bdattoma (dismissed):
left a couple of comments but in general it looks good

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: >Could this be moved outside? maybe in the setup?
This can be done

>Fetch Openshift CA Bundle kw can be implemented in the helpers.py so can be used directly in the send_random_inference_request
I don't like this idea because I don't think send_random_inference_request should be fetching the CA as well. The logic in that function already account for the CA if it exists, otherwise it functions as it always has

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
lugi0: I like this proposal, will do

diegolovison:
> We expect the calls to fail only in that scenario, and this allows us to see a failure in case the connection appears to be insecure in other environments. Your logic works around this and never exposes the problem if it's encountered.

In this case, create one test for that and for the others let the "framework" decide what certificate should be used.

bdattoma (approved):

diegolovison (changes requested):

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -133,6 +134,7 @@ Model Serving Suite Setup
     RHOSi Setup
     Launch Dashboard    ${TEST_USER.USERNAME}    ${TEST_USER.PASSWORD}    ${TEST_USER.AUTH_TYPE}
     ...    ${ODH_DASHBOARD_URL}    ${BROWSER.NAME}    ${BROWSER.OPTIONS}
+    Fetch CA Certificate If RHODS Is Self-Managed
```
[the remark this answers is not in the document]
lugi0: I don't think this is correct, `Fetch CA certificate` is a keyword that fetches the CA regardless, while `Fetch CA Certificate If RHODS Is Self-Managed` leverages that keyword but only if RHODS is in a self-managed environment.
This way, `Fetch CA certificate` can also be used by other keywords and other tests for whatever purposes we want.

The changes have already been tested against OSD and Self-Managed, you can find the links in our slack

PR #840, opened 2023-06-23
Validates OdhDashboardConfig in 'Is RHODS Installed' keyword

manosnoam:
`Is RHODS Installed` only checks if all RHODS namespaces exist
(redhat-ods-monitoring, redhat-ods-applications, redhat-ods-operator),
but not if RHODS actually installed.

This can lead to false positive, 
for example when RHODS operator was uninstalled in Openshift UI, 
while RHODS namespaces still exist.

Validating OdhDashboardConfig existence 
can prevent such false positive of 'Is RHODS Installed' keyword.

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
@@ -3,6 +3,7 @@ Is RHODS Installed
   IF  "${cluster_type}" == "selfmanaged"
       ${result}=  Run Keyword And Return Status
       ...  Run Keywords
+      ...  Oc Get  kind=OdhDashboardConfig  namespace=redhat-ods-applications  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-monitoring  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-applications  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-operator  AND
@@ -11,6 +12,7 @@ Is RHODS Installed
   ELSE IF  "${cluster_type}" == "managed"
       ${result}=  Run Keyword And Return Status
       ...  Run Keywords
+      ...  Oc Get  kind=OdhDashboardConfig  namespace=redhat-ods-applications  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-monitoring  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-applications  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-operator  AND


ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
```diff
@@ -3,6 +3,7 @@ Is RHODS Installed
   IF  "${cluster_type}" == "selfmanaged"
       ${result}=  Run Keyword And Return Status
       ...  Run Keywords
+      ...  Oc Get  kind=OdhDashboardConfig  namespace=redhat-ods-applications  AND
```
bdattoma, line 6: probably, more reliable ways would be:
- check the operator is installed (i.e., subscription object or catalog source)
- check there are any pods in one of the namespaces

bdattoma:
> `Is RHODS Installed` only checks if all RHODS namespaces exist (redhat-ods-monitoring, redhat-ods-applications, redhat-ods-operator), but not if RHODS actually installed.
> 
> This can lead to false positive, for example when RHODS operator was uninstalled in Openshift UI, while RHODS namespaces still exist.
> 
> Validating OdhDashboardConfig existence can prevent such false positive of 'Is RHODS Installed' keyword.

did you face such an issue during your testing? 

How did you perform the uninstall from UI? RHODS is not supposed to be uninstalled from Openshift console UI

manosnoam:
> did you face such an issue during your testing?

Yes, many times when the reinstall of RHODS failed, for example when operator has been removed, but kfdef pods hangs, so those namespaces where not removed.

> How did you perform the uninstall from UI? RHODS is not supposed to be uninstalled from Openshift console UI.

It happen when running our automation scripts. Uninstalling from UI is just another example that can cause it (you can also run `oc delete operator ...` and then this keyword would return false positive too.

The point of this PR is to improve current implementation, since now `IS RHODS Installed` is actually `ARE RHODS NAMESPACES EXISTS`, which leads to false positive.

bdattoma:
> > did you face such an issue during your testing?
> 
> Yes, many times when the reinstall of RHODS failed, for example when operator has been removed, but kfdef pods hangs, so those namespaces where not removed.
> 
> > How did you perform the uninstall from UI? RHODS is not supposed to be uninstalled from Openshift console UI.
> 
> It happen when running our automation scripts. Uninstalling from UI is just another example that can cause it (you can also run `oc delete operator ...` and then this keyword would return false positive too.
> 
> The point of this PR is to improve current implementation, since now `IS RHODS Installed` is actually `ARE RHODS NAMESPACES EXISTS`, which leads to false positive.

I'm okay with merging this one, it's a small change. However I wonder if there are cases where even adding the check on OdhDashboardConfig would bring to false positive.

jgarciao (approved):

bdattoma:
@tarukumar why is it set as "do not merge"?

tarukumar:
@bdattoma 
I have reservations about all the comments provided here, but I prefer not to delve into them at this moment. That's why I included that tag. I would like to discuss these concerns during the automation meeting if I am able to attend.

PR #841, opened 2023-06-26
Add option to use custom user settings in ods-ci image

bdattoma:
This PR wants to add a new feature of ods-ci image to adapt it to accept custom user configuration instead of using only randomly created users and passwords.

It required some refactor of the previous code plus adding new one. The user configuration can be set up with a JSON file and a template file is attached in the PR.

The new code allows us to:
1. use fix usernames and password
2. use usernames with a fix prefix + random suffix
3. define the mapping btw IDP users and TEST_USER_* in the test-variables.yaml
4. define group mapping

Note. Point 2 is necessary to mantain compatibility with the Interop team use-case (on openshift-ci) without asking them to maintain a configuration file + allows us not to publish internal configuration.

UPDATES:
- now the PR contains the updated documentation markdown file
- the install script contains a validation script checking the json fields and values before proceeding with actual installation.


manosnoam:
Hi @bdattoma Can it be run on RHODS Cluster that already created IDP users?
Will it update the existing users, or add new users ?

bdattoma:
> Hi @bdattoma Can it be run on RHODS Cluster that already created IDP users? Will it update the existing users, or add new users ?

it will detect the presence of either LDAP and HTP and will skip the installation. If the IDPs are installed, you're supposed to use the image without `SET_ENVIRONMENT=1`

jgarciao:
Could you document how to use it in the different scenarios in https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/ODS-CI-IMAGE-README.md or another document?

bdattoma:
> Could you document how to use it for the different scenarios in https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/ODS-CI-IMAGE-README.md or another document?

yes, it's part of the TO DOs for this PR. I'm currently testing the scenarios

bdattoma:
> > Could you document how to use it for the different scenarios in https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/ODS-CI-IMAGE-README.md or another document?
> 
> yes, it's part of the TO DOs for this PR. I'm currently testing the scenarios

@jgarciao it's documented now! I'd suggest reviewing the doc before the code :)

bdattoma:
I reached a final version for it. I've tested on PSI and OSD, with or without OCM. It looks good to me. There is one missing scenario I'll try to get covered soon. It would be good if I can get some reviews

manosnoam (approved):
Thank you Berto for this patch.

jgarciao (approved):

PR #842, opened 2023-06-28
Update model serving xpaths for RHODS 1.29

lugi0:


bdattoma:
tested on OSD, pr test job #1900 @lugi0

jgarciao (approved):

PR #851, opened 2023-07-10
Wait until pipeline route is ready to avoid firing pipeline alert

jgarciao:
If the pipeline route is tested before it is ready, alert "Data Science Pipelines Application Route Error Burn Rate" is fired.

This PR adds a sleep and increases the timeout between each test of the route to avoid it.

### DIFF
--- a/ods_ci/libs/DataSciencePipelinesAPI.py
+++ b/ods_ci/libs/DataSciencePipelinesAPI.py
@@ -83,7 +83,8 @@ class DataSciencePipelinesAPI:
             count += 1
 
         assert self.route != "", "Route must not be empty"
-        print(f"DSP route should be working: {self.route}")
+        print(f"Waiting for Data Science Pipeline route to be ready: {self.route}")
+        time.sleep(30)
         status = -1
         count = 0
         while status != 200 and count < timeout:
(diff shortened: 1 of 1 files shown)


manosnoam (approved):
lgtm

jgarciao:
Before the changes in this PR, we could see in  rhods-ci-pr-test/1912/TestReport that the alerts weren't firing in the Suite Setup but they were firing in the Suite Teardown:

```
Executing keyword 'Alerts Should Not Be Firing' failed:
Suite Trdwn: Tests.Ods Dashboard.Data Science Pipelines.Data-Science-Pipelines-Api: Alerts should not be firing: ['Data Science Pipelines Application Route Error Burn Rate (for:120, state:pending)', 'Data Science Pipelines Application Route Error Burn Rate (for:900, state:pending)', 'Data Science Pipelines Application Route Error Burn Rate (for:3600, state:pending)', 'Data Science Pipelines Application Route Error Burn Rate (for:10800, state:pending)']
```

When running this PR in rhods-ci-pr-test/1913, there was still the alert firing from the previous tests (for:3600 and for:10800) but didn't fire again for intervals 120 or 900 and in prometheus we could see no error 500s in the pipeline route
image

bdattoma (approved):

PR #859, opened 2023-07-14
First keywords for operator rearch

lugi0:


ods_ci/tasks/Tasks/patch_dsc_cr_operatorv2.robot
```diff
@@ -0,0 +1,237 @@
...
+
+# Verify None Installation Profile
+#     [Documentation]
+#     Run Keyword And Continue On Failure    Verify Component Resources    component=none
+#     Component Should Not Be Enabled    datasciencepipelines
+#     Component Should Not Be Enabled    dashboard
+#     Component Should Not Be Enabled    distributedWorkloads
+#     Component Should Not Be Enabled    kserve
+#     Component Should Not Be Enabled    modelmeshserving
+#     Component Should Not Be Enabled    workbenches
```
bdattoma, line 118: these are not valid anymore right?

ods_ci/tasks/Tasks/patch_dsc_cr_operatorv2.robot
```diff
@@ -0,0 +1,237 @@
...
+    ${expected} =    Get File    ${filepath}${component}.txt
+    Run    oc get $(oc api-resources --namespaced=true --verbs=list -o name | awk '{printf "%s%s",sep,$0;sep=","}') --ignore-not-found -n opendatahub -o=custom-columns=KIND:.kind,NAME:.metadata.name | sort -k1,1 -k2,2 | grep -v "PackageManifest\\|Event\\|ClusterServiceVersion" > ${filepath}${component}_runtime.txt  # robocop: disable
+    Process Resource List    filename_in=${filepath}${component}_runtime.txt
+    ...    filename_out=${filepath}${component}_processed.txt
+    ${actual} =    Get File    ${filepath}${component}_processed.txt
+    Remove File    ${filepath}${component}_processed.txt
+    Remove File    ${filepath}${component}_runtime.txt
+    Should Be Equal As Strings    ${expected}    ${actual}
+
+Verify If Component Is Enabled
```
bdattoma, line 146: I'd call it sth like `Is Component Enabled` since it is not actually verifying it

ods_ci/tasks/Tasks/patch_dsc_cr_operatorv2.robot
```diff
@@ -0,0 +1,237 @@
...
+    RETURN    ${status}
+
+Patch DataScienceCluster CustomResource Using Test Variables
+    [Documentation]  Enables a mix of components based on the values provided in test-variables.yml
+    [Arguments]    ${dsc_name}=default
+    ${len} =    Get Length    ${COMPONENTS}
+    ${patch} =    Set Variable    ${PATCH_PREFIX}
+    FOR    ${index}    ${cmp}    IN ENUMERATE    @{COMPONENTS}
+        IF    ${COMPONENTS.${cmp}} == ${True}
+            ${sub} =    Change Component Status    component=${cmp}    run=${FALSE}
```
bdattoma, line 159: can you set enable=true explicitly? to improve readability

ods_ci/tasks/Tasks/patch_dsc_cr_operatorv2.robot
lugi0: Will be moved to a different test file rather than a task

ods_ci/tasks/Tasks/patch_dsc_cr_operatorv2.robot
```diff
@@ -0,0 +1,185 @@
...
+#         IF    "${COMPONENTS.${cmp}}" == "True"
+#             Component Should Be Enabled    ${cmp}
+#         ELSE IF    "${COMPONENTS.${cmp}}" == "False"
+#             Component Should Not Be Enabled    ${cmp}
+#         ELSE
+#             Fail    msg=Invalid parameters in test-variables.yml
+#         END
+#     END
+
+# Switch To Core Installation Profile
```
tarukumar, line 33: Could we remove profile text from the lists since we are not officially saying profile anymore

lugi0: Yes, this code will be moved to test cases

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -21,6 +27,9 @@ Install RHODS
           FAIL    Provided test envrioment is not supported
       END
   END
+  IF  "${UPDATE_CHANNEL}" != "stable"
```
mwaykole, line 30: for managed we use beta and self managed we use stable so i think we should change condition

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -21,6 +27,9 @@ Install RHODS
           FAIL    Provided test envrioment is not supported
       END
   END
+  IF  "${UPDATE_CHANNEL}" != "stable" and "${UPDATE_CHANNEL}" != "beta"
```
tarukumar, line 30: This should be OR not AND

lugi0: My understanding is that it should be `and`, since we want to create the CR only if the channel is neither stable nor beta (i.e. alpha or fast).
If I use `or` it would create the DSC CR also when the channel is `stable` (false or true == true) or when it's `beta` (true or false == true), and of course when it's `alpha`/`fast` (true or true == true). With the `and` as it is now it would be (false and true == false) (true and false == false) (true and true == true).

lugi0:
Pending additional validation through Jenkins

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -111,3 +132,57 @@ Wait For Pods Numbers
...
+    ${file_path} =    Set Variable    ods_ci/tasks/Resources/Files/
+    Log to Console    Requested Configuration:
+    FOR    ${cmp}    IN    @{COMPONENTS}
+        Log To Console    ${cmp} - ${COMPONENTS.${cmp}}
+    END
+    Create DataScienceCluster CustomResource Using Test Variables
+    ${yml} =    Get File    ${file_path}dsc_apply.yml
+    Log To Console    Applying DSC yaml
+    Log To Console    ${yml}
+    Run    oc apply -f ${file_path}dsc_apply.yml
```
bdattoma, line 148: you could embed these lines into `Create DataScienceCluster CustomResource Using Test Variables`

lugi0: `Create` only creates the file, I don't think it should also apply it

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -111,3 +132,57 @@ Wait For Pods Numbers
...
+            Component Should Not Be Enabled    ${cmp}
+        ELSE
+            Fail    msg=Invalid parameters in test-variables.yml
+        END
+    END
+
+Create DataScienceCluster CustomResource Using Test Variables
+    [Documentation]
+    [Arguments]    ${dsc_name}=default
+    ${file_path} =    Set Variable    ods_ci/tasks/Resources/Files/
```
tarukumar, line 163: @lugi0 you need to give path from task not from ods_ci since jenkins executed task from ods_ci folder

lugi0: fixed

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/131__rhods_operatorv2.robot
```diff
@@ -0,0 +1,180 @@
...
+    Log    ${status}
+    Sleep    30s
+
+Change Component Status
+    [Documentation]    Enables or disables a single component. Can either run the patch command directly (if `run` is
+    ...    set to true) or return the patch string to be combined later for a bigger patch command.
+    [Arguments]    ${component}    ${run}=${TRUE}    ${enable}=true
+    IF    "${component}" not in @{COMPONENTS}
+        Log    unknown component: ${component}
+        RETURN
```
bdattoma, line 139: should we fail in this case?

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/131__rhods_operatorv2.robot
```diff
@@ -0,0 +1,180 @@
...
+        END
+        IF    ${index} ==0
+            ${patch} =    Catenate    SEPARATOR=    ${patch}    ${sub}
+        ELSE
+            ${patch} =    Catenate    SEPARATOR=,${SPACE}    ${patch}    ${sub}
+        END
+    END
+    ${patch} =    Catenate    SEPARATOR=    ${patch}    }}}'
+    Log    ${patch}
+    ${status} =    Run    ${patch}
```
bdattoma, line 178: should we check and fail in case the run fails?

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/131__rhods_operatorv2.robot
bdattoma: right right. Maybe, could you set the log level to WARN? so we may see soon if there's something wrong with the test itself in the future

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/131__rhods_operatorv2.robot
bdattoma: it might be good to fail as soon as possible instead of going forward, but it depends on whats next in the test. Anyway, I don't think it's critical now, feel free to ignore

lugi0: Added this check

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/131__rhods_operatorv2.robot
lugi0: updated to log level warn

bdattoma (dismissed):

manosnoam (approved):
Tested on GCP self-managed both with 1.31 V1 and V2 operator. 
rhods-smoke/3058/ passed successfully.

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/131__rhods_operatorv2.robot
```diff
@@ -0,0 +1,167 @@
...
+    Run Keyword And Continue On Failure    Verify Component Resources    component=none
+    Component Should Not Be Enabled    datasciencepipelines
+    Component Should Not Be Enabled    dashboard
+    Component Should Not Be Enabled    kserve
+    Component Should Not Be Enabled    modelmeshserving
+    Component Should Not Be Enabled    workbenches
+
+
+*** Keywords ***
+Verify Component Resources
```
bdattoma, line 75: what about naming it `Enable Component And Verify Resources?`

bdattoma (approved):
lgtm. Left few comment (could be implemented in future PRs if you prefer)

ods_ci/test-variables.yml.example
```diff
@@ -47,3 +47,11 @@ GITHUB_USER:
   USERNAME: USER
   EMAIL: email@gmail.com
   TOKEN: token
+# Components to set as enabled in the operator v2.0 DSC CR
+COMPONENTS:
+  dashboard: true
+  datasciencepipelines: true
+  # Right now only one between modelmesh and kserve can be enabled
+  kserve: false
```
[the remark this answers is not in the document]
lugi0: yes

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -29,44 +36,67 @@ Verify RHODS Installation
...
-  ...                   namespace=redhat-ods-monitoring
-  ...                   label_selector=prometheus=rhods-model-monitoring
-  ...                   timeout=400
-  Wait For Pods Status  namespace=redhat-ods-applications  timeout=60
-  Log  Verified redhat-ods-applications  console=yes
   Wait For Pods Status  namespace=redhat-ods-operator  timeout=1200
   Log  Verified redhat-ods-operator  console=yes
-  Wait For Pods Status  namespace=redhat-ods-monitoring  timeout=1200
-  Log  Verified redhat-ods-monitoring  console=yes
+  IF  "${UPDATE_CHANNEL}" != "stable" and "${UPDATE_CHANNEL}" != "beta"
```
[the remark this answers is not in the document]
lugi0: No, that will be in channel alpha so we need to find another way to check for it

mwaykole (approved):

PR #867, opened 2023-07-31
First round of automation for new serving stack

bdattoma:
The scope of this PR is to set the basic keyword to get the test cluster ready with the new model serving stack.
It's implementing 2 test cases:
- ODS-2326
- ODS-2325 (possible overalp with #859)

It also applies come improvements to the keyword which installs openshift operators from CLI. Due to this last change, Starburst Enterprise suite gets impacted, but this PR applies the necessary change.

The code developed here is likely going to change while development on the product side continue


lugi0:
Wouldn't this be already covered by https://github.com/red-hat-data-services/ods-ci/pull/859/files#diff-25ea385229e112b810b2d156db9f2dca6a5fa8fa7ceb5fc689ec1b7ee91ad8d5R59 ? 
Also the test does not make sure that the kserve component is enabled / does not enable it before running the test

bdattoma:
> Wouldn't this be already covered by https://github.com/red-hat-data-services/ods-ci/pull/859/files#diff-25ea385229e112b810b2d156db9f2dca6a5fa8fa7ceb5fc689ec1b7ee91ad8d5R59 ? Also the test does not make sure that the kserve component is enabled / does not enable it before running the test

true, actually the entire suite "post-install" might get broken with the new operator because those run with the assumption that everything is deployed. We can discuss offline

ods_ci/tests/Resources/OCP.resource
```diff
@@ -97,3 +97,8 @@ Fetch CA Certificate If RHODS Is Self-Managed
     IF  ${self_managed}==${TRUE}
         Fetch Openshift CA Bundle
     END
+
+Wait For Pods To Be Ready
+    [Arguments]    ${label_selector}    ${namespace}    ${timeout}=300s
+    ${rc}    ${out}=    Run And Return Rc And Output
+    ...    oc wait --for=condition=ready pod -l ${label_selector} -n ${namespace} --timeout=${timeout}
```
tarukumar, line 104: It will be better if you add check for the output in case timeout is exhausted

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -243,3 +243,8 @@ Clean Up Model Serving Page
         ${name} =    SeleniumLibrary.Get Text   ${model}
         Delete Model Via UI    ${name}
     END
+
+Add Namespace To ServiceMeshMemberRoll
+    [Arguments]    ${namespace}    ${servicemesh_ns}=istio-system
+    ${rc}    ${out}=    Run And Return Rc And Output
+    ...    oc patch smmr/default -n ${servicemesh_ns} --type='json' -p="[{'op': 'add', 'path': '/spec/members/-', 'value': \"${namespace}\"}]"
```
tarukumar, line 250: same as above

ods_ci/tests/Tests/100__deploy/100__installation/110__model_serving_watsonx.robot
```diff
@@ -0,0 +1,101 @@
+*** Settings ***
```
lugi0, line 1: As we said this can possibly be moved to the operator reach test cases, but we'll talk more about it offline. For now it should not cause issues having this code running as well

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -0,0 +1,205 @@
...
+${JAEGER_SUB_NAME}=    jaeger-product
+${KSERVE_NS}=    kserve    # will be replaced by redhat-ods-applications
+${CAIKIT_FILEPATH}=    ${LLM_RESOURCES_DIRPATH}/caikit_servingruntime.yaml
+${TEST_NS}=    watsonx
+
+
+*** Test Cases ***
+Verify External Dependency Operators Can Be Deployed
+    [Tags]    ODS-2326    WatsonX
+    Pass Execution    message=Installation done as part of Suite Setup.
```
lugi0, line 38: So this test suite is only going to validate the installation of additional operators for now? I feel it would make sense to also have a basic happy path functional test here

lugi0 (dismissed):
IMHO it could be slightly improved but it's also ok to merge as is and updated later

ods_ci/tests/Resources/OCP.resource
bdattoma: in general, i didn't mean to block execution untilt the install code gets stability, since there might be some changes in near future. I would add the blocking check in critical part like this one

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -0,0 +1,204 @@
...
+Resource          ../../../Resources/Page/Operators/ISVs.resource
+Suite Setup       Install Model Serving Stack Dependencies
+# Suite Teardown
+
+
+*** Variables ***
+${DEFAULT_OP_NS}=    openshift-operators
+${LLM_RESOURCES_DIRPATH}=    ods_ci/tests/Resources/Files/llm
+${SERVERLESS_OP_NAME}=     serverless-operator
+${SERVERLESS_SUB_NAME}=    serverless-operator
```
tarukumar, line 14: Do we need sub namespace as another variable it will be always be same as operator namespace.

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: it is in plan of course, but for allowing new PR for the same project I will implement the serving test in the next PR

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: >Do we need sub namespace


it's the subscription name, not the namespace. Initially I thought of assigning a different name for subscription in a way we can easily identity if the operator was previously installed or installed by our script (on debug time on any cluster). Wdyt?

ods_ci/tests/Resources/OCP.resource
bdattoma: in fact, in another run after the blocking check was added, the test was stopped because a delay in getting the pods started. 

The waiting should be improve to wait for the pod first of all. We actually have another custom keyword `Wait Until Operator Pods Are Running` which should do the job. The problem with that one is that it requires a lot of input (list of all the expected pods and containers etc) which are a lot in this scope...Wdyt?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
bdattoma: my comment as above

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: as of now it has the same name as the operator because I was testing on a cluster where the subscriptions where already in place (manually installed) and using a different name would have broken everything

tarukumar: Ohh sorry missed it..we can have same name to be more accurate with Manual install. ..what I was actually saying is in the below keyword where you are using the variable you don't need two varible since they are having the same name we can just use single variable at both the places.

ods_ci/tests/Resources/OCP.resource
tarukumar: Hmm ..I would suggest to check the rc outpu with run keyword and continue on failure keyword until we get the stability

tarukumar:
Overall looks good to me added some small suggestions. We can merge the pr

ods_ci/tests/Resources/OCP.resource
bdattoma: installation is now more stable, I extended the kw to check pods

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: yeah got it. As I said, I did so to give the option to have different names which i thought would have been good to have, but yeah we can change this, np

jgarciao (approved):

lugi0 (approved):
lgtm

PR #877, opened 2023-08-23
Add test for Canary model upgrade

tarukumar:


ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -65,6 +66,32 @@ Verify User Can Serve And Query A Model
...
+    ...    model_storage_uri=s3://ods-ci-wisdom/flan-t5-small/
+    ...    model_name=${model_name}
+    ...    namespace=canary-model-upgrade
+    Log To Console    Applying Canary Tarffic for Model Upgrade
+    Compile And Query LLM Model   isvc_name=${flan_isvc_name}
+    ...    sa_name=${DEFAULT_BUCKET_SA_NAME}
+    ...    model_storage_uri=s3://ods-ci-wisdom/bloom-560m/
+    ...    model_name=${model_name}
+    ...    canaryTrafficPercent=20
+    ...    namespace=canary-model-upgrade
```
bdattoma, line 87: probably, we could query multiple times and check the answer arrived 20% of the time from the new model and 80% from the previous one

bdattoma:
is this empty file needed? `ods_ci/tests/Resources/Files/llm/caikit_canary_model_upgrade.yaml`

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
tarukumar: We can or may be i will do it in future prs.  howver..the current problem i am facing is that bloom model is not getting downloaded.

PR #888, opened 2023-08-28
fix uninstall for v2 operator

lugi0:


manosnoam (approved):
This has been tested and allows the uninstall of V2, thanks!

mwaykole (approved):

PR #885, opened 2023-08-25
Add tests for deploying models with multiple replicas

bdattoma:
This PR is adding automation for ODS-2376 and applying bucket variables

PR #889, opened 2023-08-29
Add test for auto scale using Concurrency

tarukumar:
I have not update kind: KnativeServing file since that is not needed for this TC

### DIFF
--- a/ods_ci/tests/Resources/Files/llm/caikit_isvc.yaml
+++ b/ods_ci/tests/Resources/Files/llm/caikit_isvc.yaml
@@ -8,7 +8,7 @@ metadata:
   name: {{INFERENCE_SERVICE_NAME}}
 spec:
   predictor:
-    scaleMetric: {{SCALE_METRIC}
+    scaleMetric: {{SCALE_METRIC}}
     minReplicas: {{MIN_REPLICAS}}
     scaleTarget: {{SCALE_TARGET}}
     canaryTrafficPercent: {{CanaryTrafficPercent}}
--- a/ods_ci/tests/Resources/Files/llm/caikit_isvc.yaml
+++ b/ods_ci/tests/Resources/Files/llm/caikit_isvc.yaml
@@ -8,7 +8,7 @@ metadata:
   name: {{INFERENCE_SERVICE_NAME}}
 spec:
   predictor:
-    scaleMetric: concurrency
+    scaleMetric: {{SCALE_METRIC}
     minReplicas: {{MIN_REPLICAS}}
     scaleTarget: {{SCALE_TARGET}}
     canaryTrafficPercent: {{CanaryTrafficPercent}}
--- a/ods_ci/tests/Resources/Files/llm/caikit_isvc.yaml
+++ b/ods_ci/tests/Resources/Files/llm/caikit_isvc.yaml
@@ -8,6 +8,9 @@ metadata:
   name: {{INFERENCE_SERVICE_NAME}}
 spec:
   predictor:
+    scaleMetric: concurrency
+    minReplicas: {{MIN_REPLICAS}}
+    scaleTarget: {{SCALE_TARGET}}
     canaryTrafficPercent: {{CanaryTrafficPercent}}
     serviceAccountName: {{SA_NAME}}
     model:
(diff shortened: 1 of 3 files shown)


ods_ci/tests/Resources/Files/llm/caikit_isvc.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
   name: {{INFERENCE_SERVICE_NAME}}
 spec:
   predictor:
+    scaleMetric: concurrency
```
bdattoma, line 11: shall we use a variable for this too? and we set to concurrency by default

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -136,6 +136,43 @@ Verify Model Pods Are Deleted When No Inference Service Is Present
...
+    ${cmd} =     Set Variable     grpcurl -d
+    ${cmd} =    Catenate   ${cmd} ${body}
+    ${cmd} =    Catenate   ${cmd} -H ${header}
+    ${cmd} =    Catenate   ${cmd} -insecure
+    ${cmd} =    Catenate   ${cmd}  ${host}:443
+    ${cmd} =    Catenate   ${cmd}  "caikit.runtime.Nlp.NlpService/TextGenerationTaskPredict"
+    FOR    ${index}    IN RANGE    30
+            ${rc}  ${response}=   	Run And Return Rc And Output    ${cmd}&
+            Run Keyword And Continue On Failure  Should Be Equal As Integers     ${rc}  ${0}
+    END
```
bdattoma, line 164: there is already a keyword to query and check response multiple times, could you re-use that one?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -136,6 +136,43 @@ Verify Model Pods Are Deleted When No Inference Service Is Present
...
+    @{pod_lists}=    Oc Get    kind=Pod    namespace=autoscale-con
+        ...    label_selector=serving.kserve.io/inferenceservice=${flan_model_name}
+    ${count}=    Get Length   ${pod_lists}
+    IF   ${count} > ${1}
+         Log      Autoscale Using Concurrency is completed.Model Pod has been scaled up from 1 to $count
+    ELSE
+         FAIL     msg= Autoscale Using Concurrency has failed and Model pod has not been scaled up
```
bdattoma, line 171: the keyword `Wait For Pods To Be Ready` from #885 has the feature to check replicas number

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
tarukumar: nope, that keyword wait for response i am running the command in background so that it can almost run parllel. here is used "&"

ods_ci/tests/Resources/Files/llm/caikit_isvc.yaml
tarukumar: done

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: hmm did you try using it? was it preventing from scaling up?

In any case, I would avoid rewriting the entire keyword in the test case. Maybe we can still re-use one of the keyword and changing the behavior with a param. Wdyt?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: hm that keyword would need to be slighlty changed to accept a generic check like > 1

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
tarukumar: Indeed, I attempted it, but the outcome wasn't as anticipated during my local verification. As a result, I transitioned to this approach, which seems more appropriate in this context. The reason behind this shift is that when aiming for parallel loading, initiating the process in the background allows for swift commencement of new runs without unnecessary waiting period

tarukumar: Done

bdattoma (dismissed):

bdattoma (approved):

mwaykole (approved):

PR #890, opened 2023-08-29
Modify DSC CR install for 1.32 RC4

lugi0:


mwaykole:
seems to be small change /lgtm

mwaykole (approved):

jgarciao (dismissed):

tarukumar:
@lugi0 Below code need to be changed also why we are comparing managed == True with new changes, could you explain please.: 

`https://github.com/red-hat-data-services/ods-ci/pull/890/files#diff-11c2f02f8419b8b4ee8bf5f117645cfe5da320a349f21813d8da399c47a7ffb3R172-R176`

to 

`        ELSE IF    '${COMPONENTS.${cmp}}' == '${True}'
            Component Should Be Enabled    ${cmp}
        ELSE IF    '${COMPONENTS.${cmp}}' == '${False}'
            Component Should Not Be Enabled    ${cmp}
        END `

lugi0:
@tarukumar fix is up
The issue is that before RC4 the components were set to enabled with a true/false flag, instead now they use the strings `Managed` and `Removed`. I missed to update the check in that spot.

lugi0:
Verified install with Smoke #3169

tarukumar (approved):
/lgtm

mwaykole (approved):

PR #894, opened 2023-08-30
fix: Inconsisten path in Readme

FedeAlonso:
In the first step of the Readme there is an inconsistency in the path, as it assumes that the user has moved to the "ods-ci" folder instead of being in the root folder. 
In step 3 the paths are correct.


mwaykole (approved):

tarukumar (approved):
/lgtm

PR #895, opened 2023-08-30
Wait 30 seconds to check if pods are equals to 0 when deleting pipeline server

diegolovison:
Greater than 30 seconds I would say we should investigate


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataSciencePipelines.resource
```diff
@@ -110,5 +110,8 @@ Delete Pipeline Server
     Click Element    //li[contains(a/text(), 'Delete pipeline server')]
     Handle Deletion Confirmation Modal    ${data_science_project_name} pipeline server    pipeline server
     Wait Until Page Contains Element    xpath://h2[contains(text(), 'No pipeline server')]
-    ${pod_count}    Run    oc get pods -n ${data_science_project_name} | wc -l
```
jgarciao, line 113: This keyword is assuming you won't have any other pod running in the project, but that might not be always true.
You can narrow it adding a label:
```
${pod_count}    Run    oc get pods -n ${data_science_project_name} -l component=data-science-pipelines | wc -l
```

Also, the loop can be avoided with this command:
`oc wait --for=delete pod -n  ${data_science_project_name} -l component=data-science-pipelines  --timeout=60s`

diegolovison: This is not part of this test scenario.

diegolovison: Done

jgarciao (approved):

lugi0 (approved):

PR #897, opened 2023-08-30
Mark folders in utils as python packages and add pylint

Frawless:
This PR adds several init files into `utils` folder to make sure that PyCharm properly load the project. As a part of it I also added pylint and needed changes to the code.


ods_ci/libs/Helpers.py
```diff
@@ -167,12 +176,7 @@ def uninstall_managed_starburst_using_addon_flow(self, cluster_name):
     @keyword
     def inference_comparison(self, expected, received, threshold=0.00001):
         try:
-            import ast
-            import decimal
-            import numbers
-            import re
```
lugi0, line 173: Why did this move the imports out of the function?
Usually when one of these keywords is called it's a one off and no other keywords get called, so IMHO it's more efficient to only import inside the function so that it's triggered only when actually needed

ods_ci/libs/Helpers.py
```diff
@@ -225,29 +231,27 @@ def _inference_object_comparison(expected, received, threshold):
...
+        # pylint: disable=W0102
         value_range=[0, 255],
+        # pylint: disable=W0102
         shape={"B": 1, "C": 3, "H": 512, "W": 512},
         no_requests=100,
     ):
-        import os
-        import random
-        import requests
-        from pathlib import Path
```
lugi0, line 249: Same comment as above, why move the imports out?

poetry.lock
```diff
@@ -13,14 +13,14 @@ files = [
```
lugi0, line 13: This file will need to be recreated to include the changes that have been made to master as well as the additional libraries this PR adds

ods_ci/libs/Helpers.py
Frawless: The reason behind it is PEP8 recommendation that are used as part of pylint - https://peps.python.org/pep-0008/#imports . My personal preference is to have all imports and the top level. I am not sure if this somehow influence performance in the bad way to see the real difference, but if you want to keep it on function level, I can revert it and remove the check for it.

lugi0: we're talking about microseconds so it's probably not a big deal, but yeah there is a performance impact to importing libraries that do not get used. I'm not sure how Robot Framework behaves with this helper file specifically, I had put them inside the function under the assumption that the import would happen only if the keyword gets called during execution. We can leave it as is if PEP8 recommends top level imports.

Frawless: I think it works exactly as you describe. We can wait for some other reviews and thoughts about it.

kornys (dismissed):

ods_ci/libs/DataSciencePipelinesAPI.py
```diff
@@ -294,27 +299,37 @@ def retrieve_auth_url(self):
...
+            if (
+                storage_class["metadata"]["annotations"][
+                    "storageclass.kubernetes.io/is-default-class"
+                ]
+                == "true"
+            ):
+                storage_class_name = storage_class["metadata"]["name"]
                 break
-        return storage_class['metadata']['name']
+        return storage_class_name
```
bdattoma, line 320: was the linter complaining about this?

ods_ci/utils/scripts/SplitSuite.py
```diff
@@ -18,7 +18,7 @@ def chunked(lst, nchunk):
 
 
 def fetch_subsuites(parent):
-    children = [child for child in parent.suites]
+    children = set(parent.suites)
```
bdattoma, line 21: hm could you use list instead? I think there is technically the possibility to have suites with same name (under different subfolders). It shouldn't happen, but it could.

ods_ci/libs/DataSciencePipelinesAPI.py
Frawless: IIRC the reason for this was that the method returned data under some condition, but when the loop finish there were no return value. Another possible solution would be to return null after for loop.

ods_ci/utils/scripts/SplitSuite.py
Frawless: I don't remember why I put set here, will use list instead.

bdattoma: thanks!

bdattoma (dismissed):

kornys (approved):

lugi0 (dismissed):

kornys (dismissed):

kornys:
@lugi0 @bdattoma I had to do rebase and some fixes, could you please check it again?

kornys (dismissed):

jiridanek:
I'm submitting a separate PR containing only the `__init__.py` files addition. Merging the changes here piecemeal like that should be hopefully smoother

* https://github.com/red-hat-data-services/ods-ci/pull/1098

PR #903, opened 2023-09-03
Update global variables of Prometheus if required

manosnoam:
Prometheus might be created on another namespace like "istio-system" for WatsonX, but not always on redhat-ods-monitoring. 
Using a new keyword "Set Prometheus Variables" it will update the required global variables: RHODS_PROMETHEUS_URL and RHODS_PROMETHEUS_TOKEN, 
if they were not set in ods_ci/test-variables.yml (and only if Prometheus is running).


jgarciao (approved):

ods_ci/tests/Resources/Page/ODH/Prometheus/Prometheus.robot
```diff
@@ -335,3 +335,26 @@ Run Query Range
     ...    headers=${pm_headers}    verify=${False}
     Request Should Be Successful
     RETURN    ${resp}
+
+Set Prometheus Variables
+    [Documentation]    Update Global Variables RHODS_PROMETHEUS_URL and RHODS_PROMETHEUS_TOKEN if missing
+    ${result} 	Run Process 	oc get route -l app\=prometheus -A -o jsonpath\='{.items[0].metadata.namespace}'    shell=yes
+    IF    ${result.rc} == 0
+        ${prom_namespace}=    Set Variable    ${result.stdout}
+        Log    Prometheus is running in the namespace: ${prom_namespace}    console=True
```
bdattoma, line 344: is it assuming that the first instance of prometheus it find in the cluster is the correct one?

ods_ci/tests/Resources/Page/ODH/Prometheus/Prometheus.robot
```diff
@@ -335,3 +335,26 @@ Run Query Range
...
+        END
+        IF    "${RHODS_PROMETHEUS_TOKEN}" == "" or "${RHODS_PROMETHEUS_TOKEN}" == "PROMETHEUS_TOKEN"
+            ${result} 	Run Process 	oc create token prometheus -n ${prom_namespace} --duration 6h    shell=yes
+            Should Be True    ${result.rc} == 0    Error creating Prometheus token: ${result.stderr}
+            ${RHODS_PROMETHEUS_TOKEN}=    Set Variable    ${result.stdout}
+            Set Global Variable   ${RHODS_PROMETHEUS_TOKEN}
+        END
+    ELSE
+        Log    message=No Prometheus namespace found   level=WARN
+    END
```
bdattoma, line 360: could you please move the lines about the token into a separate kw and set the kw as Protected Keyword in the RHOSi.resource?

manosnoam:
Verified:
All monitoring tests passed, even when Prometheus vars where not defined in test variables:
image

ods_ci/tests/Resources/Page/ODH/Prometheus/Prometheus.robot
bdattoma: have you tried it in a OSD/ROSA cluster?

manosnoam: On OSD, not on ROSA.

bdattoma: hm okay, I'm unsure if ROSA might have any additional prometheus which could break the flow. I guess we will find out

bdattoma (approved):

jgarciao (approved):

PR #904, opened 2023-09-05
Update watsonX tests

bdattoma:
This PR is fixing the following:
- ODS-2325: post-install test which needed to be fixed after changes in the product
- Test Setup of the main test suite, updating the SMMR tasks to work with metrics - we have also commented the lines, since odh-model-controller should now do it automatically once the isvc got created


tarukumar (approved):
lgtm

PR #905, opened 2023-09-05
Bump kfp-tekton to 1.5.8

diegolovison:
Bump kfp-tekton to 1.5.8


diegolovison:
rhods-ci-pr-test/1993/TestReport/

ods_ci/libs/DataSciencePipelinesAPI.py
```diff
@@ -300,8 +299,9 @@ def get_default_storage(self):
         result, _ = self.run_oc("oc get storageclass -A -o json")
         result = json.loads(result)
         for storage_class in result['items']:
-            if storage_class['metadata']['annotations']['storageclass.kubernetes.io/is-default-class'] == 'true':
-                break
+            if 'annotations' in storage_class['metadata']:
```
jgarciao, line 302: Shouldn't you also check if  storage_class is null/empty?

ods_ci/libs/DataSciencePipelinesAPI.py
```diff
@@ -88,7 +88,6 @@ def login_and_wait_dsp_route(
 
         assert self.route != "", "Route must not be empty"
         print(f"Waiting for Data Science Pipeline route to be ready: {self.route}")
-        time.sleep(30)
```
jgarciao, line 91: I think this sleep was necessary to avoid pipeline alerts firing because of the test querying a route that is not ready yet

jgarciao (changes requested):

ods_ci/libs/DataSciencePipelinesAPI.py
diegolovison: so far, no.

ods_ci/libs/DataSciencePipelinesAPI.py
diegolovison: It is firing. It will depend on the cluster. Why firing the alert is a problem?

bdattoma:
I see more changes rather than only bumping the kfp version. Are these changes caused by the library change? Could you please extend the PR name/description?

diegolovison:
> I see more changes rather than only bumping the kfp version. Are these changes caused by the library change? Could you please extend the PR name/description?

all related

ods_ci/libs/DataSciencePipelinesKfpTekton.py
```diff
@@ -81,15 +81,18 @@ def kfp_tekton_create_run_from_pipeline_func(
             f"{current_path}/ods_ci/tests/Resources/Files/pipeline-samples/{source_code}"
         )
         pipeline = getattr(my_source, fn)
+        default_image = 'registry.redhat.io/ubi8/python-39@sha256:3523b184212e1f2243e76d8094ab52b01ea3015471471290d011625e1763af61'
```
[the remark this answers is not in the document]
diegolovison: If a test never failed in disconnected. then, yes
see: https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tests/Resources/Files/pipeline-samples/iris-pipeline-compiled.yaml#L144

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
```
lugi0, line 1: this file might need to be recreated to include changes already made in master

diegolovison: Done

bdattoma (dismissed):

ods_ci/libs/DataSciencePipelinesAPI.py
jgarciao: If the alert fires in our test when using the API because we don't wait for the route to be available but doesn't fire when using the Dashboard UI I think we are introducing noise in the test results

jgarciao:
Diego, kfp_tekton 1.5.9 was released on Sep 12. Should we use that one instead?

ods_ci/libs/DataSciencePipelinesAPI.py
diegolovison: Added it back

jgarciao (approved):

bdattoma (approved):

PR #917, opened 2023-09-07
[Fix] Update the expected CUDA version to 12.2

jstourac:
After we moved from the GPU Add-on installation to the GPU operator installation, there is a different version of CUDA installed and provided to the affected JupyterLab Notebooks.


lugi0 (approved):
lgtm

jgarciao (approved):

PR #922, opened 2023-09-10
Multiple fixes for GPU scripts

manosnoam:
In provision-gpu.sh:
- Fix shabang error: /bash/bin: bad interpreter
- Skip creating machineset if there's already one with annotations of GPU > 0
- Apply the manifest instead of create, in case machineset already exists
- Fix renaming of machineset that appends multiple "-gpu" to its name
- Remove the incorrect GPU label flag when creating the machineset
- Add GPU label to the new machineset

In gpu_deploy.sh:
- Fix GPU_INSTALL_DIR to point to the relative directory
- Chmod +x gpu_deploy.sh


manosnoam:
Output example:

image

lugi0 (approved):

lugi0 (approved):

tarukumar (approved):

PR #926, opened 2023-09-12
util.execute_command() real-time output and ocm.py verbosity level option

manosnoam:
This PR includes two improvements for real-time verbosity in:

- util.py execute_command()
Every call to `execute_command(<shell command>)` will not wait for the process to end to print the command output, but instead print the output during execution.

- ocm.py create/update/delete commands
With a new flag option to `ocm.py -v <Number>` (or `--ocm_verbose_level`), we can set verbosity level for OCM commands. The default verbosity level is set to 0, the highest is 5.
See more info at:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use

These changes are useful to see realtime log output of long running processes (for example when creating a cluster via OCM), as well as any shell command that being executed within ods-ci custom python functions.


jstourac:
Just wonder - I can see there are more occurrences of `ocm` (or `/bin/ocm` - would probably deserve to unify too?) command than just creating a cluster.

Shouldn't we be consistent and add this verbosity flag also on other occurrences too? I understand that adding this verbose flag doesn't make sense for some commands. But after very quick review e.g. `delete_cluster`, `hibernate_cluster`, `resume_cluster`, `add_machine_pool`, `uninstall_addon`, `install_addon`, `create_idp`? If we really don't want to do that, then we may want to update the help text for this new flag like this:

```
help="ocm logging verbosity level during OSD cluster creation",
```

manosnoam:
> Just wonder - I can see there are more occurrences of `ocm` (or `/bin/ocm` - would probably deserve to unify too?)

Yes, not sure why it is done so, maybe for situations where a different version of ocm is required. We can fix it in another PR.

> Shouldn't we be consistent and add this verbosity flag also on other occurrences too? 

Sure, I added it to all relevant ocm commands that either creates, updates, or deletes resources, and skipped it for list/get commands.

Help says now:
`ocm logging verbosity level for create/update/delete commands`

manosnoam:
An example for the ocm.py output when using --v=2:

image

manosnoam:
Verified:
image

bdattoma:
did you try creating a cluster with the code from this PR?

manosnoam:
> did you try creating a cluster with the code from this PR?

Yes I did, see:
https://github.com/red-hat-data-services/ods-ci/pull/926#issuecomment-1720151894

bdattoma:
> > did you try creating a cluster with the code from this PR?
> 
> Yes I did, see: [#926 (comment)](https://github.com/red-hat-data-services/ods-ci/pull/926#issuecomment-1720151894)

right, sorry I  missed that.

bdattoma (approved):

aloganat:
Changes looks good to me. We can merge it, if we have tested the workflows.

aloganat (approved):

PR #927, opened 2023-09-13
Add support for CodeFlare operator in RHODS task

winklerm:
Adds Robot framework resources for installing/uninstalling downstream CodeFlare operator as a part of RHODS operator installation/uninstallation.

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
@@ -1,3 +1,10 @@
+*** Settings ***
+Documentation    RHODS operator installation top-level keywords
+Resource   ../pre-tasks/oc_is_operator_installed.robot
+Resource   oc_install.robot
+Resource   codeflare_install.resource
+
+
 *** Keywords ***
 Installing RHODS Operator ${image_url}
   ${is_operator_installed} =  Is RHODS Installed
@@ -6,6 +13,9 @@ Installing RHODS Operator ${image_url}
   ...  Set Suite Variable  ${image_url}  AND
   ...  Set Test Variable  ${RHODS_OSD_INSTALL_REPO}  AND
   ...  Install RHODS   ${cluster_type}    ${image_url}
+  ${is_codeflare_managed} =    Is CodeFlare Managed
+  Log  Will install CodeFlare operator: ${is_codeflare_managed}  console=yes
+  IF  ${is_codeflare_managed}    Installing CodeFlare Operator
 
 RHODS Operator Should Be installed
   Verify RHODS Installation
--- a/ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
@@ -1,53 +1,86 @@
 *** Keywords ***
 Is RHODS Installed
+  Log   Checking if RHODS is installed with "${clusterType}" "${UPDATE_CHANNEL}" "${INSTALL_TYPE}"      console=yes
   IF  "${cluster_type}" == "selfmanaged"
       IF  "${UPDATE_CHANNEL}" == "odh-nightlies"
           ${result}=  Run Keyword And Return Status
           ...  Run Keywords
-          ...  Oc Get  kind=OdhDashboardConfig  namespace=opendatahub  AND
+          ...  Check A RHODS Family Operator Is Installed  namespace=opendatahub
+          ...                                              subscription=rhods-operator-dev  AND
           ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-operator  AND
           ...  Oc Get  kind=CatalogSource  namespace=openshift-marketplace
           ...          field_selector=metadata.name=redhat-operators
       ELSE
-          ${result}=  Run Keyword And Return Status
-          ...  Run Keywords
-          ...  Oc Get  kind=OdhDashboardConfig  namespace=redhat-ods-applications  AND
-          ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-monitoring  AND
-          ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-applications  AND
-          ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-operator  AND
-          ...  Oc Get  kind=CatalogSource  namespace=openshift-marketplace
-          ...          field_selector=metadata.name=redhat-operators
+          IF  "${INSTALL_TYPE}" == "CLi"
+              ${result}=  Run Keyword And Return Status
+              ...  Run Keywords
+              ...  Check A RHODS Family Operator Is Installed  namespace=redhat-ods-operator
+              ...                                              subscription=rhods-operator-dev  AND
+              ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-monitoring  AND
+              ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-applications  AND
+              ...  Oc Get  kind=Namespace  field_selector=metadata.name=redhat-ods-operator  AND
+              ...  Oc Get  kind=CatalogSource  namespace=openshift-marketplace
+              ...          field_selector=metadata.name=rhods-catalog-dev
+          ELSE IF  "${INSTALL_TYPE}" == "OperatorHub"
+              ${result}=  Run Keyword And Return Status
+              ...  Run Keywords
+              ...  Check A RHODS Family Operator Is Installed  namespace=redhat-ods-operator
+              ...                                              subscription=rhods-operator  AND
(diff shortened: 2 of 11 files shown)


winklerm:
Going to address the warnings and test it.

ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
```diff
@@ -1,3 +1,11 @@
+*** Settings ***
+Documentation    RHODS operator installation top-level keywords
+
+
+*** Variables ***
+${MANAGE_CODEFLARE_BOOL} =    Convert To Boolean    ${MANAGE_CODEFLARE_OPERATOR}
```
bdattoma, line 6: where `MANAGE_CODEFLARE_OPERATOR` defined?

ods_ci/tasks/Resources/RHODS_OLM/uninstall/codeflare_uninstall.resource
```diff
@@ -0,0 +1,57 @@
...
+
+Uninstall CodeFlare
+    [Documentation]   Removes the RHODS CodeFlare operator subscription and CSV
+    ${return_code}    ${output}    Run And Return Rc And Output
+    ...    oc delete subscription rhods-codeflare-operator -n openshift-operators
+    Should Be Equal As Integers	${return_code}	 0   msg=Error deleting CodeFlare subscription
+    ${return_code}    ${output}    Run And Return Rc And Output
+    ...    oc delete csv `oc get csv -n openshift-operators |grep rhods-codeflare-operator |awk '{print $1}'`
+    ...    -n openshift-operators
+    Should Be Equal As Integers	${return_code}	 0   msg=Error deleting CodeFlare CSV
```
bdattoma, line 27: these might be covered by `Uninstall ISV Operator From OperatorHub Via CLI` keyword. Do you mind trying?

ods_ci/tasks/Tasks/rhods_olm.robot
```diff
@@ -15,7 +15,11 @@ ${RHODS_OSD_INSTALL_REPO}    None
 ${TEST_ENV}                  AWS
 ${INSTALL_TYPE}              OperatorHub
 ${UPDATE_CHANNEL}            odh-nightlies
+${MANAGE_CODEFLARE_OPERATOR} False
```
bdattoma, line 18: need one more space before the value. 

You can set it to `${FALSE}` so you can avoid `Convert to boolean` later

ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
winklerm: It should be defined in the `deployRhods` pipeline library method here: https://gitlab.cee.redhat.com/ods/jenkins/-/merge_requests/451/diffs#75c1d85aec243f843c64b69364a3b76b88614656_127_131

Not sure if I need to declare/define it in some other place as well?

ods_ci/tasks/Tasks/rhods_olm.robot
winklerm: Nice! Just checking, will it work when I set the value in the Jenkins job as follows?
```
--test-variable MANAGE_CODEFLARE_OPERATOR:false
```
or
```
--test-variable MANAGE_CODEFLARE_OPERATOR:true
```
I mean, if the value is set as the above string, will it be converted into a boolean automatically?

bdattoma: the RF UserGuide says:
`Strings TRUE, YES, ON and 1 are converted to True,` so probably if we pass "TRUE" instead of "true" it should work.

Probably, to be on safer side, we could use Convert to Boolean as you did.

winklerm: Updated, will test it.

winklerm: I have found that Keywords cannot be called from the Variables definition as I did: https://groups.google.com/g/robotframework-users/c/salOiBEzPXE

I have introduced a new keyword `Is CodeFlare Managed` and will use that instead, testing it..

winklerm: Looks good so far.

winklerm:
Overall, the changes seem to work, I am checking the operator uninstallation (seemed not to remove everything but might have been related to an older operator installation on the cluster) and I would like to add verification for `ray` and `codeflare` components after `DataScienceCluster` is created.

ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
bdattoma: hm well, it depends on the desired final result, meaning:
1. we could decide to set a default value 'FALSE' and if anybody runs the code from local or without setting the variable at runtime, the code works.
2. the code is not supposed to run if the variable don't get set at runtime.

winklerm:
The uninstallation seems not to work because of [1] - the `OdhDashboardConfig` does not need to be present when only ray and/or codeflare components are installed, I will create a new keyword for a more appropriate check as discussed here [2].

[1] https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot#L14
[2] https://github.com/red-hat-data-services/ods-ci/pull/840#discussion_r1241831366

winklerm:
Detection of RHODS installation fixed, ready for review.

winklerm:
Hmm, in fact it seems the installation detection is still not working correctly, investigating.

ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
```diff
@@ -1,38 +1,86 @@
...
+  Log   RHODS is installed result: ${result}      console=yes
+  RETURN  ${result}
+
+Is RHODS Family Operator Installed
+  [Documentation]   Returns if an operator with given pod label selector is currently installed in given namespace
+  [Arguments]    ${namespace}  ${subscription}  ${pod_selector}
+  Log   Getting CSV from subscription ${subscription} namespace ${namespace}      console=yes
+  ${rc}    ${current_csv_name} =    Run And Return Rc And Output
+  ...    oc get subscription ${subscription} -n ${namespace} -ojson | jq '.status.currentCSV' | tr -d '"'
+  Log   Got CSV ${current_csv_name} from subscription ${subscription}, result: ${rc}      console=yes
```
bdattoma, line 68: you could reuse this kw: `Wait Until Operator Subscription Last Condition Is`

ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
```diff
@@ -1,38 +1,86 @@
...
+  ${result} =  Run Keyword And Return Status
+  ...  Run Keywords
+  ...  Oc Get  kind=Csv  namespace=${namespace}  name=${current_csv_name}   AND
+  ...  Oc Get  kind=Pod  namespace=${namespace}  label_selector=${pod_selector}
```
bdattoma, line 72: you could also use `Wait For Pods To Be Ready `

ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
winklerm: Good point! Just one doubt, is it ok to wait in this case? We are checking that the operator is installed or not (typically from a previous job run on the cluster), so waiting might not make sense in this case and just prolong the job duration?

On the other hand, it might help in some corner case, when the operator would have been installed/installing at the time this RHODS operator installation check was being made.. not sure if that is expected, though.

If you feel like reusing the waiting here would make sense, I will add it - just double checking :)

bdattoma: hm it depends on the expectations: do you expect that when this lines gets executed, the operator subscription must be ready? If yes, then it's ok not to wait

winklerm:
Tested with the rhods-smoke job, builds:
- 3434: v1 on managed AWS cluster - all tests passed,
- 3435: v2 on managed AWS cluster - 17 tests failed - confirmed by @tarukumar and @lugi0 that the failures are either a known issue or teardown/metrics/alerts related which are not supported in v2 at the moment,
- 3436: v1 on self-managed cluster - all tests passed,
- 3437: v2 on self-managed cluster - 1 test failed, the same test failed in 3435 so it is expected or not supported.

ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
winklerm: Yes - my understanding is that this code is checking for an existing operator installation which had either:
- been installed some time ago (such as in previous build of the job) so the subscription is already ready,
- or the operator is not installed at all so the code would be just waiting while no installation had been requested.

The original code seemed to use this assumption, so I would vote for keeping it and if we see any issues, we can make it more resilient (and maybe slower). Would that be fine?

tarukumar (approved):

PR #934, opened 2023-09-19
Fix always passing namespace deletion verification

winklerm:
Fix the `Wait Until Project Is Deleted` keyword always passing after timeout even when the project/namespace still exists.

Please note that the keyword `Wait Until Project is Deleted` will fail fast when the PR is merged.

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/uninstall/oc_uninstall.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/uninstall/oc_uninstall.robot
@@ -39,5 +39,6 @@ Wait Until Project Is Deleted
         ${project_exists}=  Run Keyword and return status
         ...  Oc Get  kind=Namespace  field_selector=metadata.name=${project}
         Exit For Loop If     not ${project_exists}
-
    END
+   IF  ${project_exists}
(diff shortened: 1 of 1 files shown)


ods_ci/tasks/Resources/RHODS_OLM/uninstall/oc_uninstall.robot
```diff
@@ -39,5 +39,7 @@ Wait Until Project Is Deleted
         ${project_exists}=  Run Keyword and return status
         ...  Oc Get  kind=Namespace  field_selector=metadata.name=${project}
         Exit For Loop If     not ${project_exists}
-
+   END
+   IF    ${project_exists}
+        Fail    Project ${project} has not been deleted!
```
tarukumar, line 44: could you update the message. Something like Project has not been deleted after ${timeout} attempts

winklerm: Done, thanks!

bdattoma (approved):

winklerm:
Testing revealed that this change requires also a few other fixes for operator v2 uninstallation from #927. I am going to include this change in #927 and close this PR.

PR #936, opened 2023-09-19
Small change in README + addition of a flag to open the test reports in a browser after test execution

jstourac:


jstourac:
@jgarciao JFYI

ods_ci/run_robot_test.sh
```diff
@@ -20,6 +20,8 @@ EMAIL_SERVER_USER="None"
 EMAIL_SERVER_PW="None"
 EMAIL_SERVER_SSL=false
 EMAIL_SERVER_UNSECURE=false
+OPEN_REPORT_IN_BROWSER=false
+REPORT_BROWSER="firefox" # Default browser to open reports in
```
lugi0, line 24: Shouldn't this be exposed as well like `--open-report`?

README.md
```diff
@@ -3,10 +3,12 @@ ODS-CI is a framework to test [Red Hat Open Data Science](https://www.redhat.com
 and its upstream project, [Open Data Hub](https://opendatahub.io/).
 
 # Requirements
-  Linux distribution that supports Selenium automation of a chromium web browser using [ChromeDriver](https://chromedriver.chromium.org)
-  * chromedriver binaries can be downloaded from https://chromedriver.chromium.org/downloads. The chromedriver version must match the installed version of chromium/google-chrome
+  1. Linux distribution that supports Selenium automation of a chromium web browser using [ChromeDriver](https://chromedriver.chromium.org):
+     * chromedriver binaries can be downloaded from https://chromedriver.chromium.org/downloads
+     * the chromedriver version must match the installed version of chromium/google-chrome
```
lugi0, line 8: Can you update this to point to the new chromedriver website? https://googlechromelabs.github.io/chrome-for-testing/
I'd also add Firefox/geckodriver compatibility as an option here

lugi0 (approved):

jstourac:
Ah, I can see the comments now. Sorry, I'll update.

ods_ci/run_robot_test.sh
jstourac: hm, well, I can add another flag for it, but as described in comment for `--open-report` you can use following now:

`--open-report true` - Firefox will be used to open the reports
`--open-report false` - nothing is opened after the tests execution
`--open-report google-chrome` - Google Chrome will be used to open the reports
`--open-report non-existent` - this will fail to open anything :slightly_smiling_face:

So at the moment, you don't need to use another flag to define your desired browser/app to open the directory with test reports.

README.md
jstourac: * ChromeDriver link to downloads updated
* I'm not sure whether I got your comment correctly but I added a Firefox geckodriver to the descriptions too - please review whether it's what you meant and that it's correct.

jstourac:
@lugi0 thank you for your review - I addressed your comments. This is ready for re-review. Thanks!

ods_ci/run_robot_test.sh
lugi0: Ah, sorry, totally missed that! It's good as it is then.

lugi0 (approved):

README.md
lugi0: Yep, everything looks good to me, thanks :D

ods_ci/run_robot_test.sh
```diff
@@ -290,7 +301,8 @@ fi
 if [[ ${SKIP_INSTALL} -eq 0 ]]; then
   poetry install
 fi
-source $(poetry env info --path)/bin/activate
+# shellcheck disable=SC1091
```
bdattoma, line 304: why?

ods_ci/run_robot_test.sh
```diff
@@ -317,19 +329,30 @@ echo "${exit_status}"
 
 # send test artifacts by email
 if ${EMAIL_REPORT}
- then
-     tar cvzf rf_results.tar.gz ${TEST_ARTIFACT_DIR} &> /dev/null
+  then
+     tar cvzf rf_results.tar.gz "${TEST_ARTIFACT_DIR}" &> /dev/null
```
bdattoma, line 333: was it failing?

ods_ci/run_robot_test.sh
```diff
@@ -317,19 +329,30 @@ echo "${exit_status}"
...
-                        -l ${EMAIL_SERVER_SSL} -d ${EMAIL_SERVER_UNSECURE}
+     python3 ods_ci/utils/scripts/Sender/send_report.py send_email_report -s "${EMAIL_FROM}" -r "${EMAIL_TO}" -b "ODS-CI: Run Results" \
+                        -v "${EMAIL_SERVER}" -a "rf_results.tar.gz" -u "${EMAIL_SERVER_USER}" -p "${EMAIL_SERVER_PW}" \
+                        -l "${EMAIL_SERVER_SSL}" -d "${EMAIL_SERVER_UNSECURE}"
+fi
+
+
+if test "${OPEN_REPORT_IN_BROWSER}" != "false"
+  then
+    if test "${OPEN_REPORT_IN_BROWSER}" != "true"
```
bdattoma, line 349: how many values do you expect this variable to have? "false", "true" and?

ods_ci/run_robot_test.sh
```diff
@@ -317,19 +329,30 @@ echo "${exit_status}"
...
+      then
+        REPORT_BROWSER="${OPEN_REPORT_IN_BROWSER}"
```
bdattoma, line 351: why?

ods_ci/run_robot_test.sh
bdattoma: imo i'ts ambiguous what value this variable can have. It'd be probably better to use it as boolean and control the default browser with another option

ods_ci/run_robot_test.sh
lugi0: I think he was just fixing shellcheck warning about word splitting https://github.com/koalaman/shellcheck/wiki/SC2046

ods_ci/run_robot_test.sh
lugi0: https://github.com/koalaman/shellcheck/wiki/SC1091

ods_ci/run_robot_test.sh
lugi0: https://github.com/red-hat-data-services/ods-ci/pull/936/files#diff-127a3d7ae5cdfa3e5e9cadb9daf76ea4763da79227ceffecd571fb764b93603fR163

bdattoma: @lugi0 yeah I saw that comment, but that still sounds ambiguous to me: "if not false (hence true?); if true (is it different from not false?)"

ods_ci/run_robot_test.sh
bdattoma: what happens without? sorry I don't know much about shellchek

ods_ci/run_robot_test.sh
lugi0: I'm not sure I follow your logic; if you give it false, it's not going to enter that block at all, if you give it true, it opens the default browser, if you give it a different string `xyz` it's going to open browser `xyz` (so, valid for valid browsers, fails for random strings). It looks ok to me, or am I missing something?

ods_ci/run_robot_test.sh
lugi0: I think it just prints out a warning here in GitHub. Annoying but not the end of the world.

ods_ci/run_robot_test.sh
bdattoma: just saying that the variable seem ambiguous to me, it starts as boolean but then you can assign other string values. I know it works, not saying differently. Just ambiguous to me. If it's only me seeing this it's ok to merge as it is, since it works

ods_ci/run_robot_test.sh
jstourac: Yeah - this is to disable the warning, since the warning is irrelevant for us.

Also, per configuration - if there are more than 10 warnings reported by shellcheck, then relevant CI job is marked as failed, which isn't nice for our PRs.

bdattoma: thanks!

ods_ci/run_robot_test.sh
jstourac: Well, I can update to introduce a new flag for it, but TBH - it seems simple to me to use this one flag directly so that you don't have to remember another flag name for this feature - that was my main motivation to implement it this way.

bdattoma: hmm okay, please document it in https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/docs/RUN_ARGUMENTS.md

bdattoma (approved):

lugi0 (approved):
lgtm

jstourac:
Thank you very much, @lugi0 and @bdattoma for your reviews! I just rebased, no other changes.

jstourac:
Thank you, @bdattoma for merging this one into `master`.

I can see that the commits have been squashed - is it done always for each PR? If so, is it preferred to create separate PRs for separate things? I tried to keep commits as a complete separate part of work so it's easy to cherry-pick or revert it, which is effectively discarded if these are squashed.

bdattoma:
> Thank you, @bdattoma for merging this one into `master`.
> 
> I can see that the commits have been squashed - is it done always for each PR? If so, is it preferred to create separate PRs for separate things? I tried to keep commits as a complete separate part of work so it's easy to cherry-pick or revert it, which is effectively discarded if these are squashed.

we squash for every PR.

> is it preferred to create separate PRs for separate things
how would this avoid the squash? every PR could have multiple commits and we prefer to keep the history of the main repo clean

jstourac:
> we squash for every PR.

Okay, noted - I'll create separate PRs for things like this in the future then.

> how would this avoid the squash? every PR could have multiple commits and we prefer to keep the history of the main repo clean

Yeah, every PR can have multiple commits and not squashing doesn't necessarily mean that it will pollute the `main` branch if these commits are separate part of work with appropriate commit message.

bdattoma:
> > we squash for every PR.
> 
> Okay, noted - I'll create separate PRs for things like this in the future then.
> 
> > how would this avoid the squash? every PR could have multiple commits and we prefer to keep the history of the main repo clean
> 
> Yeah, every PR can have multiple commits and not squashing doesn't necessarily mean that it will pollute the `main` branch if these commits are separate part of work with appropriate commit message.

imagine if we need to revert the changes of a PR, it becomes more effort and error prone to revert single commits against reverting one single commit. PR could also have many commits. Btw, this is just my opinion, feel free to bring this up during the automation meeting

jstourac:
> imagine if we need to revert the changes of a PR, it becomes more effort and error prone to revert single commits against reverting one single commit. PR could also have many commits. Btw, this is just my opinion, feel free to bring this up during the automation meeting

As I wrote - `if these commits are separate part of work`. But nevermind, I'll simply create separate PR for even small, but separate change, then.

bdattoma:
> if these commits are separate part of work

I feel it's a bit an ideal situation

PR #943, opened 2023-09-21
[Fix] run_robot_test.sh - OC login and return value

jstourac:
This fixes OC login on my local Fedora 38 machine. Without this change,
the strings contain also `"`.

Also missing `-` in front of the switch isn't allowed.

Tested with Python based `yq` tool:

```
$ yq --version # https://github.com/kislyuk/yq
yq 3.2.3

$ yq --help
...
  -e               set the exit status code based on the output;
...
  -r               output raw strings, not JSON texts;
...
```
and also with GoLang based `yq` tool:
```
$ ./yq_linux_amd64 --version # https://github.com/mikefarah/yq
yq (https://github.com/mikefarah/yq/) version v4.35.2
$ ./yq_linux_amd64 --help
...
  -e, --exit-status                   set exit status if there are no matches or null or false is returned
...
  -r, --unwrapScalar                  unwrap scalar, print the value with no quotes, colors or comments. Defaults to true for yaml (default true)
...
```

### DIFF
--- a/ods_ci/run_robot_test.sh
+++ b/ods_ci/run_robot_test.sh
@@ -263,15 +263,15 @@ if command -v yq &> /dev/null
                     then
                         oc_host=${api_server}
                     else
-                        oc_host=$(yq  e '.OCP_API_URL' "${TEST_VARIABLES_FILE}")
+                        oc_host=$(yq -er '.OCP_API_URL' "${TEST_VARIABLES_FILE}")
                 fi
 
 
                 if [ -z "${SERVICE_ACCOUNT}" ]
                     then
                         echo "Performing oc login using username and password"
-                        oc_user=$(yq  e '.OCP_ADMIN_USER.USERNAME' "${TEST_VARIABLES_FILE}")
-                        oc_pass=$(yq  e '.OCP_ADMIN_USER.PASSWORD' "${TEST_VARIABLES_FILE}")
+                        oc_user=$(yq -er '.OCP_ADMIN_USER.USERNAME' "${TEST_VARIABLES_FILE}")
+                        oc_pass=$(yq -er '.OCP_ADMIN_USER.PASSWORD' "${TEST_VARIABLES_FILE}")
                         oc login "${oc_host}" --username "${oc_user}" --password "${oc_pass}" --insecure-skip-tls-verify=true
                     else
                         echo "Performing oc login using service account"
--- a/ods_ci/run_robot_test.sh
+++ b/ods_ci/run_robot_test.sh
@@ -263,15 +263,27 @@ if command -v yq &> /dev/null
                     then
                         oc_host=${api_server}
                     else
-                        oc_host=$(yq -er '.OCP_API_URL' "${TEST_VARIABLES_FILE}")
+                        oc_host=$(yq -er '.OCP_API_URL' "${TEST_VARIABLES_FILE}") || {
+                            echo "Couldn't find '.OCP_API_URL' variable in provided '${TEST_VARIABLES_FILE}'."
+                            echo "Please either provide it or use '--skip-oclogin true' (don't forget to login to the testing cluster manually then)."
+                            exit 1
+                        }
                 fi
 
 
                 if [ -z "${SERVICE_ACCOUNT}" ]
                     then
                         echo "Performing oc login using username and password"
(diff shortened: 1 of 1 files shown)


jstourac:
Note - regarding the shellcheck, there is also this PR opened #936.

kornys (approved):

jstourac:
@tarukumar regarding our discussion whether to remove the `e` - if we remove it, then the script returns `0` even if the searched key isn't present in the provided variables file, see e.g.:

```
$ yq -r '.OCP_API_URLf' "ods_ci/test-variables.yml" ; echo $?
null
0
$ yq -er '.OCP_API_URLf' "ods_ci/test-variables.yml" ; echo $?
null
1
```

So, I guess we want to keep it there? Truth is that looking on the code, it won't stop execution even in case of failure. I can update that.

jstourac:
I added a commit with the check for the `yq` return value. Also rebased against `master`.

kornys (approved):

ods_ci/run_robot_test.sh
```diff
@@ -251,32 +251,46 @@ if command -v yq &> /dev/null
                     then
                         oc_host=${api_server}
                     else
-                        oc_host=$(yq  e '.OCP_API_URL' "${TEST_VARIABLES_FILE}")
+                        oc_host=$(yq -er '.OCP_API_URL' "${TEST_VARIABLES_FILE}") || {
```
bdattoma, line 254: I  believe this breaks on latest yq version. Please hold on mergin this change. Unfortunately poetry installs yq v3 which is old. We need to install yq from github.

bdattoma (changes requested):

ods_ci/run_robot_test.sh
jstourac: Hm, does `yq` from GitHub has any advantages for us? Trying locally to update my `yq` via `pip` - looks like this is the latest one available there. So I suppose they changed the way how they distribute this tool? Still, do we need it in newer version? :upside_down_face:

bdattoma: yeah, there were singificant changes from v3 to v4. We leverage on some changes present in v4 for out code. You can check the `install_idp.sh` script. Plus, it's better to use more recent version of libraries rather than very old ones for sec reason

jstourac: Okay, I'll try to check and verify this works with newer version also. Note - I am not able to easily test this on Mac, so I'll probably ask somebody to do this for me then.

bdattoma: what issue do you face?

jstourac: not having a Mac machine :wink:

bdattoma: ah sorry, I thought you were seeing issues testing in on Mac.

jstourac: @bdattoma, could you, please, be a bit more specific on what breaking issue did you expect, Berto? I downloaded the latest `yq` release from [GitHub](https://github.com/mikefarah/yq/releases/tag/v4.35.2) one and looks that it behaves as expected on my Fedora machine. I updated commit message appropriately so you can check details there.

Do you have a Mac machine? Can you test it too?

bdattoma: if I remember correctly, lines like `yq --inplace '.KEY=env(value)' <yaml_filepath>` wasn't working fine on v3. 
> looks that it behaves as expected on my Fedora machine


`yq -er '.OCP_API_URL' "${TEST_VARIABLES_FILE}"` does this work with latest version for you?

jstourac: Yeah, as I stated above - it works fine for me both with `3.2.3` and also with the latest `4.35.2`.

The reason that `--inplace` doesn't work is that such flag isn't available in `3.2.3` version.

Although, as you can see in the `run_robot_test.sh` script, we use the `yq` only to grab three variables from the `test-variables.yaml` file. Nothing more. So I don't think it breaks anything else here.

bdattoma: talked offline, it's ok to keep the -r which has been introduced in newer versions of the lib

bdattoma (approved):

jstourac:
Rebased.

Also created #953.

PR #947, opened 2023-09-22
Add option to install KServe stack from Script instead of simulating step-by-step

bdattoma:
Upstream repo has a scripted tool for installing KServe stack and this PR is adding the option to leverage on that script to install the stack for our tests.

Notes.
- the script tries to install RHODS, but if it is already installed, it shouldn't break it (it's not optimal, but it is thought as internal tool)
- had to change `Run And Watch Command` to make it able to run outside a test (e.g., in a Suite Setup) cc: @manosnoam


ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -764,3 +769,10 @@ Compile And Query LLM model
           ...    json_body=${body}    json_header=${header}
           ...    insecure=${TRUE}
     END
+
+Run Install Script
+    ${rc}=    Run And Return Rc    git clone https://github.com/opendatahub-io/caikit-tgis-serving
+    Should Be Equal As Integers    ${rc}    ${0}
+    ${rc}=    Run And Watch Command    TARGET_OPERATOR=rhods CHECK_UWM=false ./scripts/install/kserve-install.sh
+    ...    cwd=caikit-tgis-serving/demo/kserve
```
lugi0, line 779: Given
```the script tries to install RHODS, but if it is already installed, it shouldn't break it (it's not optimal, but it is thought as internal tool)```
Can you add a check on RHODS being present (`Is RHODS Installed`) before this and abort the execution if it is?

Comment (bdattoma) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot:779]: nope. I will try raising a PR for the instal script in order to make the RHODS deployment optional

Comment (lugi0) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot:779]: Yeah, that's the point :D since there's a risk it would mess up the installation if already present, and since we don't control the script directly, I would prefer to have this run only on a clean cluster. 
You could also trigger an Uninstall before proceeding with execution, but that complicates the flow quite a bit.

Comment (bdattoma) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot:779]: let me see what I can do. However consider that this is not the default option. BY default the tests will use step-by-step installation which is implemented in RF

Comment (lugi0) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot:779]: Then why not just make it Fail if RHODS is already installed, given that it is an optional flow?

Comment (bdattoma) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot:779]: because I expect RHODS installed in most of the cases. Think of a jenkins job which installs rhods but then failed for other reasons like connection issues or RHODS not installed by the jenkins timeout etc. Then you could not re-run the test because if would fail

Comment (lugi0) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot:779]: But in this specific case this script would be the one installing RHODS, so you would run it in a clean cluster and you wouldn't enable rhods installation.
If the script fails to install RHODS then the test run should also fail, since that's part of what this is trying to do, no?

Comment (bdattoma) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot:779]: >But in this specific case this script would be the one installing RHODS,

I'm assuming that we may be running the tests in a cluster where RHODS is already installed or installed via jenkins. In this PR i'm not making our RF test to set the RHODS version to install but the script will try to install always rhods from operatorhub. So if you want a different version you'll need to install it differently.

It's not ideal neither case

Comment (lugi0) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot:779]: I understand it's not ideal, but AFAIK the documentation will point to the script and will tell users to run it, so IMHO we should be running it and expecting it work with its own install.
The only correct option here would be to remove the installation steps from the script, but I don't know if that's feasible / something we want to do

Comment (bdattoma) [reply ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot]: fyi: https://github.com/opendatahub-io/caikit-tgis-serving/pull/91 
PR to make the rhods deployment optional in the install script

File: ods_ci/tests/Resources/Common.robot [hunk 2]
```diff
@@ -342,17 +342,29 @@ Run And Watch Command
...
+  IF    ${is_test} == ${FALSE}
+    ${incremental}=    Generate Random String    5    [NUMBERS]
+    ${TEST NAME}=    Set Variable    testlogs-${incremental}    
+  END
   ${process_log} =    Set Variable    ${OUTPUT DIR}/${TEST NAME}.log
   ${temp_log} =    Set Variable    ${TEMPDIR}/${TEST NAME}.log
-  Set Test Variable    ${process_log}
-  Set Test Variable    ${temp_log}
+  Set Variable    ${process_log}
+  Set Variable    ${temp_log}
```
manosnoam, line 355: I don't think you need to call "Set Variable", since it is already called in lines 352-353.

bdattoma: true, I can just remove the lines. thanks

ods_ci/tests/Resources/Common.robot
```diff
@@ -342,17 +342,29 @@ Run And Watch Command
...
   Create File    ${process_log}
   Create File    ${temp_log}
-  ${process_id} =    Start Process    ${command}    shell=True    stdout=${process_log}    stderr=STDOUT    # robocop: disable
+  IF    "${cwd}" == "${NONE}"
+      ${process_id} =    Start Process    ${command}    shell=True    stdout=${process_log}
+      ...    stderr=STDOUT
+  ELSE
+      ${process_id} =    Start Process    ${command}    shell=True    stdout=${process_log}
+      ...    stderr=STDOUT    cwd=${cwd}
+  END
```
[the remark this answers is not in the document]
bdattoma: I tried with setting cwd equal to ${EMPTY} and it failed. I guess it cannot access the argument without value. I didn't try with ${NONE} but I'd expect the same.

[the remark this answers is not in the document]
manosnoam: I don't think you need neither empty or none - Don't set any default value to `cwd` arg.
Then, when calling  `Start Process` simply pass  cwd="${cwd}"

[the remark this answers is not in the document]
bdattoma: >Don't set any default value to cwd arg.


Calls to that kw would fail in that case.
I will try with ${CURDIR}

manosnoam (dismissed):

bdattoma:
Verified in 2 scenarios:
1. against a cluster where the entire stack was already installed (PASS)
2. against a cluster where only RHODS was already installed (PASS) - however it creates a new subscription, catalog source and operator group for RHODS if we install RHODS using olminstall script. It doesn't break existing installation, though

manosnoam (approved):

lugi0 (approved):

PR #949, opened 2023-09-25
Add support to validate that image provide is based on digest not tags

tarukumar:


mwaykole (approved):

ods_ci/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
```diff
@@ -235,3 +235,18 @@ Get Instance Type Of Node
     [Arguments]    ${node}
     ${type} =    Run    oc get Node -o json ${node} | jq '.metadata.labels["beta.kubernetes.io/instance-type"]'
     RETURN    ${type}
+
+Container Image Url Using Image Digest Instead Of Tags Based On Project Name
```
bdattoma, line 239: I'd probably rename it to sth like `Container Image Url Shuld Use [...]`

ods_ci/tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -16,6 +16,8 @@ Resource            ../../../Resources/Common.robot
 Suite Setup         RHOSi Setup
 Suite Teardown      RHOSi Teardown
 
+*** Variables ***
+${OPERATOR_NAMESPACE}      redhat-ods-operator
```
bdattoma, line 20: shouldn't we check all the RHODS namespaces?

ods_ci/tests/Resources/Page/OCPDashboard/Pods/Pods.robot
tarukumar: Done

ods_ci/tests/Tests/100__deploy/100__installation/102__post_install.robot
tarukumar: they are using all namespace. Below command gives all the namespace for rhods

`oc get ns -l opendatahub.io/generated-namespace -o jsonpath='{.items[*].metadata.name}'`

mwaykole (approved):

ods_ci/tests/Tests/100__deploy/100__installation/102__post_install.robot
tarukumar: why ..this solution is more independent and we don't need test variable

tarukumar: i would rather use this instead of dependent on test variable

tarukumar: "I've implemented the changes; however, I'm not seeing significant added value. As I mentioned earlier, I intended to keep this independent because we know that the operator namespace is where the pod will be deployed for both normal RHODs and even for ODH nightly. If your thought was on the upstream ODH, then I believe most aspects need to be reevaluated in our whole suite.

mwaykole (approved):

ods_ci/tests/Tests/100__deploy/100__installation/102__post_install.robot
bdattoma: I was suggesting to use that variable because it was decided to declare the RHODS namespaces in the test-variables.yml so I thought it would have been more consistent to use those. Eventually, we should have have no mention of RHODS/ODH namespaces in the tests, but they should be controlled by the test-variables.yml (this is my understanding)

bdattoma (approved):

PR #924, opened 2023-09-11
Automate embedded RHODS install

lugi0:
To run this test case in Jenkins, you should point to **an empty cluster** (no rhods installed)

The parameters you will need to change are:

**Disable** DEPLOY_RHODS_OPERATOR
Update the ODS_CI_RUN_SCRIPT_ARGS to:
```
--test-case ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/133__embedded_operator.robot
```
If you want to install the latest build in the embedded channel in production, or
```
--test-case ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/133__embedded_operator.robot --test-variable image_url:brew.registry.redhat.io/rh-osbs/iib:IIB_NUMBER
```
If you want to install a custom IIB, in which case the cluster will also need to have an ICSP  and pull secret set.

PR #951, opened 2023-09-26
Add support for CodeFlare operator in managed env

winklerm:
Added CodeFlare-specific cluster-wide CatalogSource to handle the issue with the RHODS Addon CatalogSource being scoped to `redhat-ods-operator` namespace only, while the CodeFlare operator is expected to be installed in `openshift-operators`.

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/uninstall/codeflare_uninstall.resource
+++ b/ods_ci/tasks/Resources/RHODS_OLM/uninstall/codeflare_uninstall.resource
@@ -21,9 +21,16 @@ Uninstall CodeFlare
     [Documentation]   Removes the RHODS CodeFlare operator subscription and CSV
     # needs to be deleted first to avoid getting stuck on CodeFlare finalizer
     ${return_code}    ${output}    Run And Return Rc And Output
-    ...    oc delete mcads.codeflare.codeflare.dev mcad -n redhat-ods-applications --ignore-not-found
+    ...    oc delete mcads.codeflare.codeflare.dev --all -n redhat-ods-applications --ignore-not-found
     Should Be Equal As Integers	${return_code}	 0   msg=Error deleting CodeFlare MCAD CR
     Uninstall ISV Operator From OperatorHub Via CLI     rhods-codeflare-operator    openshift-operators
+    Uninstall CodeFlare CatalogSource
+
+Uninstall CodeFlare CatalogSource
+    [Documentation]    Removes the CodeFlare cluster-wide catalog source created for managed env, if it exists
+    ${return_code}    ${output}    Run And Return Rc And Output
+    ...    oc delete catalogsource rhods-codeflare-catalog -n openshift-marketplace --ignore-not-found
+    Should Be Equal As Integers	${return_code}	 0   msg=Error deleting CodeFlare CatalogSource
 
 CodeFlare Operator Should Be Uninstalled
     [Documentation]   Verifies and logs that the RHODS CodeFlare operator is uninstalled
--- a/ods_ci/tasks/Resources/Files/codeflare_cs_template.yaml
+++ b/ods_ci/tasks/Resources/Files/codeflare_cs_template.yaml
@@ -1,0 +1,14 @@
+kind: List
+metadata: {}
+apiVersion: v1
+items:
+    -   apiVersion: operators.coreos.com/v1alpha1
+        kind: CatalogSource
+        metadata:
+            name: rhods-codeflare-catalog
+            namespace: openshift-marketplace
+        spec:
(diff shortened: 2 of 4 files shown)


winklerm:
Testing in progress, will mark as ready when tested.

ods_ci/tasks/Resources/RHODS_OLM/uninstall/codeflare_uninstall.resource
```diff
@@ -24,6 +24,13 @@ Uninstall CodeFlare
     ...    oc delete mcads.codeflare.codeflare.dev mcad -n redhat-ods-applications --ignore-not-found
     Should Be Equal As Integers	${return_code}	 0   msg=Error deleting CodeFlare MCAD CR
     Uninstall ISV Operator From OperatorHub Via CLI     rhods-codeflare-operator    openshift-operators
+    Uninstall CodeFlare CatalogSource
+
+Uninstall CodeFlare CatalogSource
+    [Documentation]    Removes the CodeFlare cluster-wide catalog source created for managed env, if it exists
+    ${return_code}    ${output}    Run And Return Rc And Output
+    ...    oc delete catalogsource rhods-codeflare-catalog -n openshift-marketplace --ignore-not-found
```
tarukumar, line 32: I would suggest you to delete the subscription of codeflare before deleting the catalog source

ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
```diff
@@ -95,6 +95,8 @@ Uninstall RHODS V2
     ...    oc delete operatorgroup --all -n redhat-ods-operator --ignore-not-found
     Should Be Equal As Integers	${return_code}	 0   msg=Error deleting operatorgroup
     ${return_code}    ${output}    Run And Return Rc And Output    oc delete ns -l opendatahub.io/generated-namespace --ignore-not-found
+    # TODO: the RHODS namespaces do not have the above label, deleting explicitely - is this ok?
```
tarukumar, line 98: in which case you saw that these namespace are not having the same label because if any namespace created by rhods it should have that label

ods_ci/tasks/Resources/RHODS_OLM/uninstall/codeflare_uninstall.resource
winklerm: Sure, good point!

ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
winklerm: @tarukumar It was the Managed cluster on AWS, installation type CLI, the latest RHODS RC.

ods_ci/tasks/Resources/RHODS_OLM/uninstall/codeflare_uninstall.resource
winklerm: @tarukumar Not sure if I miss something, but it seems to me that the catalogsource is already being deleted after the subscription is removed? Please see the lines https://github.com/red-hat-data-services/ods-ci/pull/951/files#diff-f757092a94298927c14b51bf57cfd6bdc2bc6779d5690ef73482aa2a5d35f29aR26-R27

ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
tarukumar: The v2 operator is not compatible with the Addon. If you attempt to test it, you will notice abnormal behavior. This is likely why you observed that it was not deleted.

winklerm: @tarukumar When I checked again on self-managed, the label is there as you expected. I am going to remove the explicit namespace deletion and test it again on both managed and self-managed.

ods_ci/tasks/Resources/RHODS_OLM/uninstall/codeflare_uninstall.resource
tarukumar: correct, i overlooked

winklerm:
Tested with selfmanaged PSI and AWS. Managed with AWS fails to uninstall the RHODS operator v2, but it is expected for v2 at this moment.

winklerm:
The failing PR check was also failing at [1] which has been merged despite that - so the failure is not introduced by this PR.

[1] https://github.com/red-hat-data-services/ods-ci/pull/924/checks

sutaakar:
Not relevant any more due to changes in CodeFlare, closing.

PR #956, opened 2023-09-28
fix the machine pool exiting function

tarukumar:
Observed that this function always retrun True evnthough machine poll doesn't exist. Whne machine pool doesn't exists ocm return empty string


aloganat (approved):
lgtm

FedeAlonso (approved):
LGTM

PR #958, opened 2023-09-29
Fix starburst enteprise tests

bdattoma:
**Problem**
Test suite for SEP is often failing because the robot tries to deploy the CustomResource before the CustomResourceDefinition is available in the cluster.
**Solution**
Wait for the CRD to be available before proceeding with the rest of the test case.


bdattoma:
verified on `rhods-ci-pr-test/2017` job

jgarciao (approved):

lugi0 (approved):

PR #972, opened 2023-10-09
Fix check uninstall osd cluster

kornys:
I saw several times in jenkins that unstall osd job failed due to wrong check if cluster exists. I saw that cluster was unstalled but `ocm describe` returned non zero ecode but contains string about exception and it is not evaluated as None.


jgarciao (approved):

kornys:
@jgarciao thanks for approve but Im not able to merge it due to rights on this repo🙂

diegolovison (approved):

diegolovison:
@kornys I believe there weren't enough approvals for this PR ( It should be >= 2 )

PR #979, opened 2023-10-16
Update OpenShift Pipelines to latest version

diegolovison:
```
Tests.Ods Dashboard.Data Science Pipelines.Data-Science-Pipelines-... | PASS |
1 test, 1 passed, 0 failed
Tests.Ods Dashboard.Data Science Pipelines                            | PASS |
1 test, 1 passed, 0 failed
Tests.Ods Dashboard                                                   | PASS |
1 test, 1 passed, 0 failed
Tests                                                                 | PASS |
1 test, 1 passed, 0 failed
Output:  /home/dlovison/github/rhods/ods-ci/ods_ci/test-output/ods-ci-2023-10-16-12-36-vbGMaVNjl6/output.xml
XUnit:   /home/dlovison/github/rhods/ods-ci/ods_ci/test-output/ods-ci-2023-10-16-12-36-vbGMaVNjl6/xunit_test_result.xml
Log:     /home/dlovison/github/rhods/ods-ci/ods_ci/test-output/ods-ci-2023-10-16-12-36-vbGMaVNjl6/log.html
Report:  /home/dlovison/github/rhods/ods-ci/ods_ci/test-output/ods-ci-2023-10-16-12-36-vbGMaVNjl6/test_report.html
```

### DIFF
--- a/ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
+++ b/ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
@@ -41,7 +41,7 @@ Install Red Hat OpenShift Pipelines
         ELSE IF    ${is_self_managed} == ${FALSE} and "${platform}" == "GCP"
             ${openshift_pipelines_version}=    Set Variable    pipelines-1.10
         ELSE
-            ${openshift_pipelines_version}=    Set Variable    pipelines-1.11
+            ${openshift_pipelines_version}=    Set Variable    pipelines-1.12
         END
     ELSE IF    "4.14" in "${oc_version}"
         ${openshift_pipelines_version}=    Set Variable    pipelines-1.12
--- a/ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
+++ b/ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
@@ -20,19 +20,16 @@ Install Red Hat OpenShift Pipelines
     [Documentation]    Install Red Hat OpenShift Pipelines operator depending on
     ...     OpenShift version, platform (AWS/GCP/...) and RHODS flavor (Managed/Self-Managed) with the objective of
     ...     testing all the operator versions without having to test all possible combinations:
-    ...     - OpenShift 4.10:                      OpenShift Pipelines 1.8
     ...     - OpenShift 4.11:                      OpenShift Pipelines 1.9
     ...     - OpenShift 4.12, 4.13 (AWS and GCP):  OpenShift Pipelines 1.10 / 1.11 (depending on platform and flavor)
     ...     - OpenShift >= 4.12 (OpenStack):       OpenShift Pipelines 1.11
-    ...     - OpenShift 4.14:                      OpenShift Pipelines 1.11
+    ...     - OpenShift 4.14:                      OpenShift Pipelines 1.12
     ...     Note: the complete operator compatiblity matrix is published at
     ...      https://docs.openshift.com/container-platform/4.13/cicd/pipelines/op-release-notes.html
     ${oc_version}=    Get OpenShift Version
     ${is_self_managed}=    Is RHODS Self-Managed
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
```diff
@@ -46,8 +43,10 @@ Install Red Hat OpenShift Pipelines
         ELSE
             ${openshift_pipelines_version}=    Set Variable    pipelines-1.11
```
jgarciao, line 44: OpenShift 4.14 is not available yet, so OpenShift Pipelines 1.12 won't be tested so much

One option could be test 1.12 also in OpenStack:
`- OpenShift >= 4.12 (OpenStack):       OpenShift Pipelines 1.12`

ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
jgarciao, line 1: We should try to detect if the operator is already installed and do nothing in that case (to make this keyword faster)

Once the OpenShift Pipelines operator is installed, we need to wait a little bit more (e.g. adding Sleep 30s) so RHODS Dashboard will properly detect that it is available. We have seen the pipeline tests in Smoke failing because of this.

We could use the  existing keywords `Install ISV Operator From OperatorHub Via CLI`, `Wait Until Operator Subscription Last Condition Is` and `Wait For Pods To Be Ready` as it is done by `Install Service Mesh Stack` in ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot

jgarciao (changes requested):

apodhrad:
Hi @jgarciao is there any reason why we don't use the channel latest at
https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tests/Resources/Files/redhat-openshift-pipelines.yaml#L7
and then avoid guessing a version at
https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource#L33-L51 ?

ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
diegolovison: Sleep is not good. We need to find the right way and replace https://github.com/red-hat-data-services/ods-ci/blob/8008f0619953cc510aea3a9d1ba01af520ab0abc/ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource#L58-L59 with the proper solution

apodhrad: > We should try to detect if the operator is already installed and do nothing in that case (to make this keyword faster)

Hi @diegolovison this is something we need to implement for Component Dependency Testing (something like interop testing) where all dependent operators (e.g. pipelines) are already installed. Please let me know if that's something you are already working on or if some help from my side is needed (e.g. due to your other assignments).

jgarciao:
@apodhrad Here is the support matrix for OpenShift Pipelines: https://docs.openshift.com/pipelines/1.12/about/op-release-notes.html. As they support multiple versions of their operator, we try to test as many options as possible, but in an automated way (we didn't want to add another parameter in Jenkins). That's why the logic is hardcoded like this.

ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
diegolovison: done

ods_ci/tests/Resources/Page/Operators/OpenShiftPipelines.resource
diegolovison: Should be addressed in other PR.
This is for update the version.

diegolovison:
Done. Ready for review

jgarciao (approved):

apodhrad (approved):

PR #983, opened 2023-10-19
Retag all current Sanity tests to Tier1

asanzgom:


bdattoma:
I would not merge this before defining the list of the test to move into the new Sanity

bdattoma:
you will probably need to re-do the change for the same tests, so I propose not to replace all the Sanity with Tier1, but starting replacing only the sanity tests from the following suites:
- API access test
- Jupyterhub

Wdyt?

jgarciao (approved):

PR #985, opened 2023-10-24
Add WatsonX metrics test

bdattoma:
This PR is adding a test for checking model metrics for WatsonX/Kserve. 

The PR checks:
1. metrics with `tgi_`, `caikit_` and `istio_` are present in the Openshift metrics
2. user can get `# of inference requests over defined time period` , `# of successful inference requests over defined time period` and `CPU Utilization`

This is a base for the final test which should check that all the expected metrics are present (_waiting for a final list_) and will check also more derived metrics (_waiting for a final list and their formulas)

### DIFF
--- a/ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
+++ b/ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
@@ -988,11 +988,9 @@ TGI Caikit And Istio Metrics Should Exist
     ${caikit_metrics_names}=    Get Thanos Metrics List    thanos_url=${thanos_url}    thanos_token=${thanos_token}
     ...    search_text=caikit
     ${caikit_metrics_names}=    Split To Lines    ${caikit_metrics_names}
-    Set Suite Variable    ${CAIKIT_METRICS_NAMES}    ${caikit_metrics_names}
     ${istio_metrics_names}=    Get Thanos Metrics List    thanos_url=${thanos_url}    thanos_token=${thanos_token}
     ...    search_text=istio
     ${istio_metrics_names}=    Split To Lines    ${istio_metrics_names}
-    Set Suite Variable    ${ISTIO_METRICS_NAMES}    ${istio_metrics_names}
     ${metrics}=    Append To List    ${tgi_metrics_names}    @{caikit_metrics_names}    @{istio_metrics_names}
     RETURN    ${metrics}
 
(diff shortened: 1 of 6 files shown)


ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -671,12 +713,21 @@ Deploy Caikit Serving Runtime
     ...    oc apply -f ${CAIKIT_FILEPATH} -n ${namespace}
 
 Set Project And Runtime
-    [Arguments]    ${namespace}
+    [Documentation]    Creates the DS Project (if not exists), creates the data connection for the models,
+    ...                creates caikit runtime. This can be used as test setup
+    [Arguments]    ${namespace}    ${enable_metrics}=${FALSE}
+    ${test_var}=    Set Variable    hello there!
```
FedeAlonso, line 719: Is this variable used? Is the value right?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -429,13 +431,53 @@ Verify User Can Serve And Query Flan-t5 Large
     [Teardown]    Clean Up Test Project    test_ns=${test_namespace}
     ...    isvc_names=${models_names}
 
+Verify User Can Access Model Metrics From UWM
+    [Documentation]    Verifies that model metrics are available for users in the
+    ...                OpenShift monitoring system (UserWorkloadMonitoring)
+    ...                PARTIALLY DONE: it is checking number of requests, number of successful requests
+    ...                and model pod cpu usage. Waiting for a complete list of expected metrics and
+    ...                derived metrics.
+    [Tags]    ODS-XYZ    WatsonX
```
FedeAlonso, line 440: ODS-XYZ means that the test is not defined in Polarion?

FedeAlonso (changes requested):

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: no it was used for debug reasons and removed

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: correct, i've replace it already with a real value

FedeAlonso (dismissed):
LGTM

bdattoma:
verified at `rhods-ci-pr-test/2064` SUCCESS

tarukumar:
@bdattoma why don't you enable metric in Suite setup. I don't see any harm in that those are cluster wide configuration and having it enable by default will be good idea.

bdattoma:
> @bdattoma why don't you enable metric in Suite setup. I don't see any harm in that those are cluster wide configuration and having it enable by default will be good idea.

I thought to have scenarios running without metrics enabled, since users won't be forced to enable them. But yeah I can add it to suite setup, do you mind if we move it there in another PR? this one has been already tested

FedeAlonso (approved):
LGTM

tarukumar (approved):

PR #987, opened 2023-10-24
Reduce the scope of the mlpipeline_minio_artifact_secret parameter value and dynamic bucket name for Tekton test

diegolovison:
Please keep the two commits
job/rhods/job/rhods-ci-pr-test/2060/console


bdattoma:
> lease keep the two commits
> job/rhods/job/rhods-ci-pr-test/2060/console
@diegolovison 
Are you asking of not squashing commits?

Could you pls write some lines in the description explaining what's the PR is aiming for?

bdattoma (approved):

diegolovison:
Hi @bdattoma 

Reducing the scope of the mlpipeline_minio_artifact_secret parameter value means reducing the parameter value string length. Before the change, the content was too broad and now it is limited to what is being used inside the code.

Related to the dynamic bucket name for the Tekton test, before the change, the parameter value was static, now it is retrieved from OpenShift. We support minio and S3. A special condition was added to handle that.

PR #988, opened 2023-10-25
Fix custom image tests

lugi0:


FedeAlonso (dismissed):
LGTM

ods_ci/tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -17,7 +17,7 @@ Force Tags       JupyterHub
 *** Variables ***
 ${YAML} =         tests/Resources/Files/custom_image.yaml
 ${IMG_NAME} =            custom-test-image
-${IMG_URL} =             quay.io/thoth-station/s2i-lab-elyra:v0.1.1
+${IMG_URL} =             quay.io/opendatahub-contrib/workbench-images:jupyter-datascience-c9s-py311_2023c_latest
```
diegolovison, line 20: It should be using the hash

diegolovison (changes requested):

ods_ci/tests/Tests/500__jupyterhub/custom-image.robot
lugi0: Better to use the tag in this case IMHO as it will keep testing an updated image without having to change the test every time

FedeAlonso (dismissed):

lugi0:
verified in rhods-ci-pr-test/2088

jstourac (approved):

jgarciao (approved):

PR #993, opened 2023-10-27
[Stabilize] Let's wait for DOM ready before we start to edit workbenches

jstourac:
ODS-1931

The test that edits workbench name and description fails intermittently. I wasn't able to reproduce but based on the data, it looks that the page is updated before all the fields are populated. This leads to the situation where there are both old and new value appended in the text box and test failure.

Waiting for the DOM to be ready [1] should avoid these issues.

[1] https://html.spec.whatwg.org/multipage/dom.html#reporting-document-loading-status

Tested by job - rhods-ci-pr-test/2079


jstourac:
I don't believe that test failures in `dryrun` github action are related. Thus this is ready for review, thanks.

jgarciao (approved):

PR #995, opened 2023-10-29
Test Pipeline YAML context menu and verify read only

manosnoam:
- Test that the context menu (right click) of a Pipeline YAML has a command menu, and responds to user typing "Fold all" yaml tags.
- Verify YAML text is readonly (Bug https://github.com/opendatahub-io/odh-dashboard/issues/1689)

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
@@ -112,6 +112,24 @@ Click Action From Actions Menu
     Wait Until Page Contains Element    //table//tr[td[*[a[text()='${pipeline_name}']]]]//td[contains(@class, 'pf-c-table__action')]/div/ul/li/button[contains(text(), '${action}')]
     Click Element    //table//tr[td[*[a[text()='${pipeline_name}']]]]//td[contains(@class, 'pf-c-table__action')]/div/ul/li/button[contains(text(), '${action}')]
 
+Pipeline Context Menu Should Be Readonly
+    [Documentation]   Test Pipeline YAML context menu and verify read only (https://github.com/opendatahub-io/odh-dashboard/issues/1689)
+    [Arguments]    ${pipeline_name}
+    ${orig wait}=    SeleniumLibrary.Set Selenium Implicit Wait    10s
+    Wait Until Element is Visible     //button[contains(text(), "Data Science Pipelines")]          timeout=30
+    Click Button      Data Science Pipelines
+    Click Link      Pipelines
+    Wait Until Element is Visible     //a[text()="${pipeline_name}"]          timeout=30s
+    Click Link      ${pipeline_name}
+    Click Element     //button[@aria-label="Pipeline YAML Tab"]
+    Open Context Menu    //div[contains(@class, 'lines-content')]
+    Click Element     //button[@aria-label="Pipeline YAML Tab"]
+    Press Keys    //div[contains(@class, 'lines-content')]    F1+fold+SPACE+all+ENTER
+    Press Keys    //div[contains(@class, 'lines-content')]    cannot_enter_read_only
+    Wait Until Element Is Visible    //div[@class="pf-c-code-editor pf-m-read-only odh-dashboard__code-editor"]    timeout=10
+    Capture Page Screenshot
+    SeleniumLibrary.Set Selenium Implicit Wait    ${orig wait}
+
 Pipeline Should Be Listed
     [Documentation]    Checks a pipeline is listed in the DS Project details page
     [Arguments]     ${pipeline_name}    ${pipeline_description}
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
@@ -112,19 +112,28 @@ Click Action From Actions Menu
     Wait Until Page Contains Element    //table//tr[td[*[a[text()='${pipeline_name}']]]]//td[contains(@class, 'pf-c-table__action')]/div/ul/li/button[contains(text(), '${action}')]
     Click Element    //table//tr[td[*[a[text()='${pipeline_name}']]]]//td[contains(@class, 'pf-c-table__action')]/div/ul/li/button[contains(text(), '${action}')]
 
-Pipeline Context Menu Should Be Readonly
-    [Documentation]   Test Pipeline YAML context menu and verify read only (https://github.com/opendatahub-io/odh-dashboard/issues/1689)
+Pipeline Context Menu Should Be Working
+    [Documentation]   Test Pipeline YAML context menu works with mouse right-click and with keyboard
     [Arguments]    ${pipeline_name}
     ${orig wait}=    SeleniumLibrary.Set Selenium Implicit Wait    10s
-    Wait Until Element is Visible     //button[contains(text(), "Data Science Pipelines")]          timeout=30
-    Click Button      Data Science Pipelines
-    Click Link      Pipelines
+    Menu.Navigate To Page    Data Science Pipelines    Pipelines
     Wait Until Element is Visible     //a[text()="${pipeline_name}"]          timeout=30s
     Click Link      ${pipeline_name}
     Click Element     //button[@aria-label="Pipeline YAML Tab"]
     Open Context Menu    //div[contains(@class, 'lines-content')]
     Click Element     //button[@aria-label="Pipeline YAML Tab"]
     Press Keys    //div[contains(@class, 'lines-content')]    F1+fold+SPACE+all+ENTER
+    Capture Page Screenshot
+    SeleniumLibrary.Set Selenium Implicit Wait    ${orig wait}
+
+Pipeline Yaml Should Be Readonly
+    [Documentation]   Verify Pipeline Yaml is read only (https://github.com/opendatahub-io/odh-dashboard/issues/1689)
+    [Arguments]    ${pipeline_name}
+    ${orig wait}=    SeleniumLibrary.Set Selenium Implicit Wait    10s
+    Menu.Navigate To Page    Data Science Pipelines    Pipelines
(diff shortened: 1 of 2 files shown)


manosnoam:
This new test has been verified:
image

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
```diff
@@ -112,6 +112,24 @@ Click Action From Actions Menu
     Wait Until Page Contains Element    //table//tr[td[*[a[text()='${pipeline_name}']]]]//td[contains(@class, 'pf-c-table__action')]/div/ul/li/button[contains(text(), '${action}')]
     Click Element    //table//tr[td[*[a[text()='${pipeline_name}']]]]//td[contains(@class, 'pf-c-table__action')]/div/ul/li/button[contains(text(), '${action}')]
 
+Pipeline Context Menu Should Be Readonly
+    [Documentation]   Test Pipeline YAML context menu and verify read only (https://github.com/opendatahub-io/odh-dashboard/issues/1689)
+    [Arguments]    ${pipeline_name}
+    ${orig wait}=    SeleniumLibrary.Set Selenium Implicit Wait    10s
+    Wait Until Element is Visible     //button[contains(text(), "Data Science Pipelines")]          timeout=30
+    Click Button      Data Science Pipelines
+    Click Link      Pipelines
```
bdattoma, line 121: you can use this keyword to move in the desired section`Menu.Navigate To Page`

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
```diff
@@ -112,6 +112,24 @@ Click Action From Actions Menu
...
+    Wait Until Element is Visible     //a[text()="${pipeline_name}"]          timeout=30s
+    Click Link      ${pipeline_name}
+    Click Element     //button[@aria-label="Pipeline YAML Tab"]
+    Open Context Menu    //div[contains(@class, 'lines-content')]
+    Click Element     //button[@aria-label="Pipeline YAML Tab"]
```
bdattoma, line 126: by looking at the GH issue, i understood that there are 2 different scenario:
1. the pipeline YAML editor should be read only
2. the Command Palette should not allow to run commands. I guess this is the "Context Menu" this keyword is referring.

Is my understanding correct? If yes, I would expect this "Open Context Menu" to be placed either in a separated keyword or after checking the Pipeline YAML editor is read-only

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
```diff
@@ -112,6 +112,24 @@ Click Action From Actions Menu
...
+    Press Keys    //div[contains(@class, 'lines-content')]    F1+fold+SPACE+all+ENTER
+    Press Keys    //div[contains(@class, 'lines-content')]    cannot_enter_read_only
+    Wait Until Element Is Visible    //div[@class="pf-c-code-editor pf-m-read-only odh-dashboard__code-editor"]    timeout=10
```
bdattoma, line 129: shouldn't this be placed after `Click Element //button[@aria-label="Pipeline YAML Tab"]` in order to wait for the Editor to be rendered?

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/434__data-science-pipelines-ui.robot
```diff
@@ -105,11 +109,8 @@ Verify Pipeline Metadata Pods Are Not Deployed When Running Pipelines
 Pipelines Suite Setup    # robocop: disable
     [Documentation]    Sets global test variables, create a DS project and a data connection
     Set Library Search Order    SeleniumLibrary
+    # TODO: Install Pipeline only if it does not already installed
     Install Red Hat OpenShift Pipelines
-    ${prj_title}=    Set Variable    ${PRJ_BASE_TITLE}-${TEST_USER_3.USERNAME}
-    ${iris_pipeline_name}=    Set Variable    ${PIPELINE_TEST_BASENAME}-${TEST_USER_3.USERNAME}
-    Set Suite Variable    ${PRJ_TITLE}    ${prj_title}
-    Set Suite Variable    ${PIPELINE_TEST_NAME}    ${iris_pipeline_name}
```
bdattoma, line 112: why removing these lines?

manosnoam: Since these are widely used variables, I moved them under *** Variables *** section:
```
${PRJ_BASE_TITLE}=   dsp
${PRJ_DESCRIPTION}=   ${PRJ_BASE_TITLE} is a test project for validating DS Pipelines feature
```
https://github.com/red-hat-data-services/ods-ci/pull/995/files#diff-0e3e8f6733ca76331b5ae98a47739e91267cb03bd98a02b4d57615fe8ea4480aR16-R17

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
manosnoam: The commands menu should allow to run commands, and this is tested just by a single line of code:
`Press Keys    //div[contains(@class, 'lines-content')]    F1+fold+SPACE+all+ENTER`

If I move this line to a new keyword, we will have to rerun the prerequisites to reach a pipeline yaml in that kw, which will also take more execution time, and duplicating code.

In case we decide to add a complete new test, for example to test all the available command options - if we see it is high priority use case (I don't think it is though) - we can consider to add a new test with more keywords.

For current use case I believe keeping the readonly validation + context check in same keyword is fine.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
manosnoam: Not sure I understand your question... 
The implicit wait I added removes the requirement to add explicit wait for elements.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
manosnoam: Thanks - fixed it.

manosnoam:
Changed to use Menu navigation kw, rebased and tested:

image

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
bdattoma: My concern is that we don't check that the YAML is read-only by trying editing it without using the context-menu.

For example, you could have the pre-requisites in the test. Then you have a kw to test the context menu and you have a kw to try editing the YAML from the Editor (without context menu).

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
bdattoma: hm so why do you need to wait here? is the page or the element reloaded?

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/434__data-science-pipelines-ui.robot
bdattoma: ah, okay :)

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
manosnoam: Here I wait explicitly - To see that the "Read Only" alert becomes visible (it is hidden by default). 
See it in the bottom of screenshot:
https://github.com/red-hat-data-services/ods-ci/pull/995#issuecomment-1784051461

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
manosnoam: In the context menu the "edit" commands do not exists. That's why I need to send keyboard types to check if edit is not allowed.

bdattoma: yeah I'm suggesting to try editing without leveraging on context menu

manosnoam: Yes, that is what the logic does: 
1) Open Context Menu with mouse to verify it is available.
2) Close the menu by clicking the YAML again
3) Open the context menu with F11 and send command to fold all.
4) Try to edit the YAML without the context menu, by sending keyboard types

bdattoma: could you make more clear it in the code? The kw is called "Pipeline Context Menu Should Be Readonly" but actually mixes the context menu with Editor check. I think also for future maintanance it would be better to have clear steps in the test.
You could move the pre-requisites steps outside the kw and run them once. But then with separated keywords you check the editor is read-only in any available scenarios (context menu vs direct typing in the Editor)

manosnoam: Separated to 2 keywords:

image

manosnoam: nope, it's testing one of the available commands ("fold all"), and verifies both mouse reaction and keyboard

diegolovison (approved):

PR #996, opened 2023-10-30
Add support to create DSC without any conditon for rhods24

tarukumar:
From RHODS 2.4 all the channel will have v2 operator. In this PR we are removing the condition to apply DSC and keep it as a default. 

PS: We can also remove the condition for component verification using update channel but for now the above changes should work for RHODS 2.4

### DIFF
--- a/ods_ci/tasks/Resources/Files/dsc_template.yml
+++ b/ods_ci/tasks/Resources/Files/dsc_template.yml
@@ -16,6 +16,8 @@ spec:
       managementState: <modelmeshserving_value>
     ray:
       managementState: <ray_value>
+    trustyai:
+      managementState: <trustyai_value>
     workbenches:
       managementState: <workbenches_value>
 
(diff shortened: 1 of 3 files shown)


jstourac (dismissed):

jstourac:
LGTM, any successful run to prove this is working as expected?

Also, I suppose this needs to wait for merge until 1.34 branch is created, right?

tarukumar:
> LGTM, any successful run to prove this is working as expected?
> 
> Also, I suppose this needs to wait for merge until 1.34 branch is created, right?

yes...and one run has been successful already.

jstourac:
@tarukumar looks like this change needs to be incorporated https://github.com/red-hat-data-services/ods-ci/commit/3dcde980e3f446ff240d3b13a454515f75291c03 - test run here `rhods-smoke/3819`. Please add, thank you.

aloganat (approved):
lgtm

mwaykole (approved):

PR #998, opened 2023-10-31
[Fix] name of the robotframework keyword `Set NVidia GPU Accelerator`

jstourac:
Proper name is `Set GPU Accelerator`.

Without this change the test fails with following error:

```
1046
Verify Number Of Available GPUs Is Correct With Machine Autoscaler... | PASS |
1047
------------------------------------------------------------------------------
1048
Verify Node Autoscales And Then Scales Down :: Tries spawning a se... | FAIL |
1049
No keyword with name 'Set NVidia GPU Accelerator' found. Did you mean:
1050
    JupyterHubSpawner.Set GPU Accelerator
1051
------------------------------------------------------------------------------
1052
Tests.Jupyterhub.Autoscaling-Gpus :: Tests for a scenario in which... | FAIL |
1053
2 tests, 1 passed, 1 failed
```

### DIFF
--- a/ods_ci/tests/Tests/500__jupyterhub/autoscaling-gpus.robot
+++ b/ods_ci/tests/Tests/500__jupyterhub/autoscaling-gpus.robot
@@ -58,7 +58,7 @@ Spawn Notebook And Trigger Autoscale
     ...    of the GPU node.
     Select Notebook Image    ${NOTEBOOK_IMAGE}
     Select Container Size    Small
-    Set NVidia GPU Accelerator
+    Set GPU Accelerator
     Set Number Of Required Accelerators    1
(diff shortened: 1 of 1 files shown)


jstourac:
Drurun failures reduced from 9 to 6.

Ready for review, @jgarciao - it would be nice to merge even for 1.34, I believe.

FedeAlonso (approved):
It has come because an error I reviewed with @bdattoma where I have already updated "Set GPU Accelerator" in order to be generic but on GH still showed "Set NVidia GPU Accelerator". We did not know what was really happening and the tests run OK. 
Sorry for the inconvenience and thanks for fixing it @jstourac .
LGTM

jgarciao (approved):

Issue #953, opened 2023-09-26
Two different `yq` implementations used among our tooling


jstourac:
As a result of my work on #943 I realized that there are two `yq` implementations and both are still alive (I was under the impression that the first one is dead since we download the second one in our codebase):

* https://github.com/kislyuk/yq - Python based
* https://github.com/mikefarah/yq - GoLang based

I'm not sure what were reasons to move to the GoLang based one. From the usage in our tests here we use it either to get particular value from yaml file or to update particular variable in the yaml file (using the  `--inplace` flag).

If I don't miss anything, both features are implemented also by the Python based `yq` implementation, e.g.:

```
yq -y --in-place '.BROWSER.NAME="ahoy"' /tmp/test-variables.yml
```

---

My proposal is to ditch usage of the GoLang based `yq` implementation among this project (and also Jenkins docker image) so that we can rely on Poetry definition completely and these two implementations aren't mixed up in our environments.

bdattoma:
Thanks for the analysis. 

IMO we can do a try migrating to python-based lib if this semplify our dependency mgmt. 
Something to keep in mind is that the goLang-based one seems more continuously committed. It's probably not a big issue for us, unless there are CVEs which arise.

PR #1005, opened 2023-11-02
Fix faulty kserve tests

bdattoma:
KServe (aka LLM or WatsonX) has some failing tests. This PR wants to fix them. 

As part of the test fixing, I had to do some refactoring and remove yq from poetry. 

yq dependency installation should be handled in a different PR to solve https://github.com/red-hat-data-services/ods-ci/issues/953


bdattoma:
**PR validation**
1. `rhods-ci-pr-test/2100` a better run, 2 failures remaining ODS-2377 & ODS-2380

jstourac:
@bdattoma please, there seem to be 2 trailing lines reported by robocop and consider adding the doc to the one keyword robocop complan. :slightly_smiling_face:

bdattoma:
> @bdattoma please, there seem to be multiple trailing whitespace lines reported by robocop and consider adding the doc to the one keyword robocop complan. 🙂

done

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -866,33 +886,41 @@ Model Response Should Match The Expectation
...
 
-Query Models And Check Responses Multiple Times
+Query Model Multiple Times
     [Documentation]    Queries and checks the responses of the given models in a loop
     ...                running ${n_times}. For each loop run it queries all the model in sequence
-    [Arguments]    ${models_names}    ${namespace}    ${endpoint}=${CAIKIT_ALLTOKENS_ENDPOINT}    ${n_times}=10
-    ...            ${streamed_response}=${FALSE}    ${query_idx}=0
+    [Arguments]    ${model_name}    ${namespace}    ${isvc_name}=${model_name}    ${endpoint}=${CAIKIT_ALLTOKENS_ENDPOINT}    ${n_times}=10
+    ...            ${streamed_response}=${FALSE}    ${query_idx}=0    ${validate_response}=${TRUE}    &{args}
+    IF    ${validate_response} == ${FALSE}
```
jstourac, line 894: Do you think it makes sense to follow Robocop advice and simplify this to:
```suggestion
    IF    ${validate_response}
```

(the true/false code block branches need to be switch in such case of course)

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -866,33 +886,41 @@ Model Response Should Match The Expectation
...
-            ...    json_body=${body}    json_header=${header}
-            ...    insecure=${TRUE}    skip_res_json=${streamed_response}
-            Log    ${res}
+        ${res}=    Query Model With GRPCURL   host=${host}    port=443
+        ...    endpoint=${endpoint}
+        ...    json_body=${body}    json_header=${header}
+        ...    insecure=${TRUE}    skip_res_json=${skip_json_load_response}
+        ...    &{args}
+        Log    ${res}
+        IF    ${validate_response} == ${TRUE}
```
jstourac, line 910: Do you think it makes sense to follow the Robocop advice here to change this to:
```suggestion
        IF    ${validate_response}
```
?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -866,33 +886,41 @@ Model Response Should Match The Expectation
...
-Compile And Query LLM model
-    [Arguments]    ${isvc_name}     ${model_storage_uri}    ${model_name}
+Compile Deploy And Query LLM model
+    [Documentation]    Group together the test steps for preparing, deploying
+    ...                and querying a model
+    [Arguments]    ${model_storage_uri}    ${model_name}    ${isvc_name}=${model_name}
     ...            ${canaryTrafficPercent}=${EMPTY}   ${namespace}=${TEST_NS}  ${sa_name}=${DEFAULT_BUCKET_SA_NAME}
-    ...            ${multiple_query}=${EMPTY}
+    ...            ${n_queries}=${1}    ${query_idx}=${0}    ${validate_response}=${TRUE}
+    ${models_names}=    Create List    ${model_name}
```
jstourac, line 923: seems that this robocop alert is relevant? Variable assigned but not used -> should it be removed?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
```diff
@@ -1064,3 +1083,28 @@ Check Query Response Values    # robocop:disable
...
+    [Arguments]    ${exp_percentage}    ${isvc_name}    ${model_name}    ${namespace}
+    ${total}=    Set Variable    ${20}
+    ${hits}=    Set Variable    ${0}
+    ${toleration}=    Set Variable    ${20}
+    FOR    ${counter}    IN RANGE    ${0}    ${total}
+        Log    ${counter}
+        ${status}=    Run Keyword And Return Status
+        ...    Query Model Multiple Times    isvc_name=${isvc_name}    model_name=${model_name}    n_times=1
+        ...    namespace=${namespace}
+        IF    ${status} == ${TRUE}
```
jstourac, line 1101: Do you think it makes sense to follow robocop advise to simplify this?
```suggestion
        IF    ${status}
```

pyproject.toml
```diff
@@ -24,7 +24,6 @@ requests = "^2.28.2"
 escapism = "^1.0.1"
 semver = "^2"
 rpaframework = "^27.2.0"
-yq = "^3.1.0"
```
jstourac, line 27: Out of my curiosity - what was the particular problem here if you can describe a bit?

jstourac (changes requested):
Thank you, put some comments but otherwise LGTM.

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: it doesn't open the details of the alert, however it was originally done with IF var == TRUE, but it doesn't work for all the combinations we may have, because in the ELSE statement I'm not setting the var to a static value, but to another variable.

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: yep, removed

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: nope, I want the variable to accept boolean values only. 
`IF ${status}` is verified when the variable is not None

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: besides `IF ${validate_response}` is verified when the variable is not None

pyproject.toml
bdattoma: it was complaining for `-i (--inplace)`. In the python version of the library it must be `Y --in-place`.

jstourac: So the very same incompatibility we already experienced. Thanks for clarification.

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
jstourac: Just to be sure - I suppose that by `verified` you mean that the expression is resolved as true, right?

In this case I think I'm missing something. Based on the documentation from the `Run Keyword And Return Status` keyword, I understand that the status is always boolean only. So is this just a precaution in case of bad code change in the future?

pyproject.toml
bdattoma: yep

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
bdattoma: >Just to be sure - I suppose that by verified you mean that the expression is resolved as true, right?

yes


> I understand that the status is always boolean only


Ah, sorry I mixed it up with previous comment which is the same but on a different line where the variable is taken as argument. In this particular case we could remove the `== ${TRUE}`. In general i think it's one of my coding habits to explicitly check for the value even if it's True.

jstourac: Okay, I don't insist. Feel free to let it the way it is, I guess. Just wanted to check whether I'm not missing anything :upside_down_face: :slightly_smiling_face:

jstourac (approved):

FedeAlonso (approved):

PR #1015, opened 2023-11-08
[Fix] Tests.Jupyterhub.test-pipChanges-not-permanent

jstourac:
- the `paramiko` is now part of the `science-notebook` so let's use the `minimal-notebook` instead - also, this notebook should be smaller so it could be spinned up a bit faster.
- the default maximum timeout for the check of `import paramiko` failure is too long, let's wait up to 10seconds max.

Relates to: https://github.com/opendatahub-io/notebooks/issues/261

CI: rhods-ci-pr-test/2112

### DIFF
--- a/ods_ci/tests/Tests/500__jupyterhub/test-pipChanges-not-permanent.robot
+++ b/ods_ci/tests/Tests/500__jupyterhub/test-pipChanges-not-permanent.robot
@@ -21,7 +21,7 @@ Verify pip Changes not permanent
     Stop JupyterLab Notebook Server
     Capture Page Screenshot
     Fix Spawner Status
-    Spawn Notebook With Arguments
+    Spawn Notebook With Arguments    image=minimal-notebook
     Verify Package Is Not Installed In JupyterLab  paramiko
     Capture Page Screenshot
 
@@ -32,4 +32,4 @@ Test Setup
     Begin Web Test
     Wait for RHODS Dashboard to Load
     Launch JupyterHub Spawner From Dashboard
-    Spawn Notebook With Arguments
+    Spawn Notebook With Arguments    image=minimal-notebook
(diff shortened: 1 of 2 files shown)


ods_ci/tests/Tests/500__jupyterhub/test-pipChanges-not-permanent.robot
```diff
@@ -21,7 +21,7 @@ Verify pip Changes not permanent
     Stop JupyterLab Notebook Server
     Capture Page Screenshot
     Fix Spawner Status
-    Spawn Notebook With Arguments
+    Spawn Notebook With Arguments    image=s2i-minimal-notebook
```
jgarciao, line 24: Can you change it for `image=minimal-notebook`?  This will make it compatible with ODH too

ods_ci/tests/Tests/500__jupyterhub/test-pipChanges-not-permanent.robot
```diff
@@ -32,4 +32,4 @@ Test Setup
     Begin Web Test
     Wait for RHODS Dashboard to Load
     Launch JupyterHub Spawner From Dashboard
-    Spawn Notebook With Arguments
+    Spawn Notebook With Arguments    image=s2i-minimal-notebook
```
jgarciao, line 35: Can you change it for `image=minimal-notebook`?  This will make it compatible with ODH too

jgarciao (changes requested):

ods_ci/tests/Tests/500__jupyterhub/test-pipChanges-not-permanent.robot
jstourac: updated

bdattoma:
> the paramiko is now part of the science-notebook so let's use the minimal-notebook instead 


what's the logic behind this choice? apart from minimal-nb to be smaller @jstourac

jstourac:
> what's the logic behind this choice? apart from minimal-nb to be smaller @jstourac

@bdattoma  Installation of `paramiko` python package is used to check that after the image is restarted, it's not present on the instance again. Since this package is present by default on the SDS now, we need to either choose a different python package to use for this test or choose some other notebook that don't contain it. I chose the second option.

jgarciao (approved):

PR #1019, opened 2023-11-10
[Fix] of the xpath for the jupyterlab workbench image version check

jstourac:
This is a followup for the previous fix #1016 [1] where I missed that I'm updating two different xpaths actually, which led us into the failing another test.

Relates to: opendatahub-io/notebooks#261

[1] 53ff0d41485485153e95d0433778fee889ff358f

CI check: rhods-ci-pr-test/2131


lugi0 (approved):
tested and all passing, lgtm

FedeAlonso (approved):

PR #1021, opened 2023-11-15
Update Monitoring tests for RHODS 2.4.0 RC2

jgarciao:
Update alerting rules, recording rules and alert groups to reflect monitoring changes in RHODS 2.4.0 RC2

I've tested it in rhods-ci-pr-test/2145/:
-  3 tests failed because one component was disabled and a minor bug that I've addressed after.
-  I will send a follow-up PR adapting the tests to components that can be enabled/disabled


FedeAlonso (approved):
LGTM

asanzgom (approved):

ods_ci/tests/Resources/Page/ODH/Prometheus/Prometheus.robot
```diff
@@ -121,6 +121,9 @@ Alert Should Be Firing    # robocop: disable:too-many-calls-in-keyword
                     END
                 END
             END
+            IF    ${alert_found}
+                Exit For Loop
+            END
```
bdattoma, line 126: why not keep checking for other alerts?

jgarciao: This keyword `Alert Should Be Firing`  is meant to find if a particular alert is firing in an alert_group.  This change is to address a bug when, after having found the  particular alert, in the next loop iteration there is an alert group with the same name and the value of ${alert_found} is set back to False.

PR #1026, opened 2023-11-20
Change the DSC name based on new changes

tarukumar:
Based on the new changes the DSC  name now `default-dsc` . making changes for consistency


manosnoam (approved):
This should solve the failure on trying to delete subscription when it was already deleted.

aloganat (approved):

PR #1027, opened 2023-11-20
Uninstall RHODS: Delete all subscription instead of the first one only

manosnoam:
This should fix:

1) When there's no subscription at all, re-install will fail on: 
`error: resource(s) were provided, but no name was specified`

2) Currently Uninstall delete just the first subscription in redhat-ods-operator, which doesn't make scene


manosnoam:
We want to skip the whole subscription removal, as it was already deleted. See:
https://github.com/red-hat-data-services/ods-ci/pull/1026

PR #1030, opened 2023-11-20
[Fix] custom image tests and one test removal

jstourac:
This fixes custom image tests against 2.4 RC2 release and removes one test that I believe is redundant now.

There are two tests failing due to the product bug, issues tracked here:

* https://github.com/opendatahub-io/odh-dashboard/issues/2185
* https://github.com/opendatahub-io/odh-dashboard/issues/2186

---

CI: rhods-ci-pr-test/2148

Screenshot from 2023-11-22 13-20-44

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
@@ -541,11 +541,12 @@ RHODS Notification Drawer Should Contain
     Close Notification Drawer
 
 Open Notebook Images Page
-    [Documentation]    Opens the RHODS dashboard and navigates to the Notebook Images page
+    [Documentation]    Opens the RHODS dashboard and navigates to the Notebook Image Settings page
     Wait Until Page Contains    Settings
     Page Should Contain    Settings
-    Menu.Navigate To Page    Settings    Notebook images
+    Menu.Navigate To Page    Settings    Notebook image settings
     Wait Until Page Contains    Notebook image settings
+    Wait Until Page Contains    Import new image    # This should assure us that the page content is ready
 
 Import New Custom Image
     [Documentation]    Opens the Custom Image import view and imports an image
@@ -553,7 +554,7 @@ Import New Custom Image
     [Arguments]    ${repo}    ${name}    ${description}    ${software}    ${packages}
     Sleep  1
     Open Custom Image Import Popup
-    Input Text    xpath://input[@id="byon-image-repository-input"]    ${repo}
+    Input Text    xpath://input[@id="byon-image-location-input"]    ${repo}
     Input Text    xpath://input[@id="byon-image-name-input"]    ${name}
     Input Text    xpath://input[@id="byon-image-description-input"]    ${description}
     # No button present anymore?
@@ -563,13 +564,8 @@ Import New Custom Image
 
 Open Custom Image Import Popup
     [Documentation]    Opens the Custom Image import view, using the appropriate button
-    ${first_image} =  Run Keyword And Return Status  Page Should Contain Element  xpath://button[.="Import image"]
-    IF  ${first_image}==True
-        Click Element  xpath://button[.="Import image"]
-    ELSE
-        Click Element  xpath://button[.="Import new image"]
-    END
-    Wait Until Page Contains    Import notebook images
+    Click Element  xpath://button[.="Import new image"]
+    Wait Until Page Contains    Import notebook image
 
 Add Softwares To Custom Image
     [Documentation]    Loops through a dictionary to add software to the custom img metadata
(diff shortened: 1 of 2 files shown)


manosnoam:
@jstourac can you please add screenshot of test execution, and test job number ?

manosnoam:
@jstourac has `Test Bad Image Import` become deprecated, or is it already being covered in `Test Bad Image URL` ?

jstourac:
> @jstourac can you please add screenshot of test execution, and test job number ?

It's part of the description already. Shall we really include also the screenshot here?

> @jstourac has Test Bad Image Import become deprecated, or is it already being covered in Test Bad Image URL ?

Well, based on the descriptions of these two:

```
Test Bad Image URL
    [Documentation]  Test adding an image with a bad repo URL (should fail)
```
and
```
Test Bad Image Import
    [Documentation]  Import a broken image and confirm it is disabled
    ...    in the JH spawner page
```

Now the functionality doesn't allow you to create image in both of these cases. So the latter one doesn't make sense to me anymore.

manosnoam:
> @jstourac can you please add screenshot of test execution, and test job number ?

Looking at the build I see it did not pass:
image

Can you run again and share results ?

jstourac:
@manosnoam as mentioned in the PR description, re-execution won't help in these two cases. Or are you referring to something else?

ods_ci/tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -55,26 +57,25 @@ Verify Custom Image Can Be Added
     #Should Match  ${spawner_packages}  ${IMG_PACKAGES}
 
     Spawn Notebook With Arguments  image=${IMAGESTREAM_NAME}  size=Small
-    [Teardown]  Custom Image Teardown
+    ${CLEANUP}=  Set Variable  True
```
jgarciao, line 60: The reason of using this variable is because you need all the steps in the keyword to be executed before cleaning up? 

If that's the reason, you can use the tags `robot:continue-on-failure` or `robot:recursive-continue-on-failure` to force all steps being executed. There is more info here: https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#continue-on-failure

jstourac: The reason I added it here is to distinguish the behavior of the `Custom Image Teardown`. By default it performs also cleanup operations that are tied with the Jupyter Notebook environment. But this part isn't necessary if the test execution fails before this line. So the default value is set to `false`. Maybe your approach could be used for this too, but I went the way without modifying the current teardown keyword.

manosnoam:
> @manosnoam as mentioned in the PR description, re-execution won't help in these two cases. Or are you referring to something else?

I think we should see that all the smoke tests passed, to indicate that the patch did not break anything.
Especially to see that these tests have passed:

- Test Image From Local registry
- Test Bad Image URL
- Test Bad Image URL

jstourac:
@manosnoam well, how can I make these tests passed since these are product bugs there?

manosnoam (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -563,13 +564,8 @@ Import New Custom Image
...
     [Documentation]    Opens the Custom Image import view, using the appropriate button
-    ${first_image} =  Run Keyword And Return Status  Page Should Contain Element  xpath://button[.="Import image"]
-    IF  ${first_image}==True
-        Click Element  xpath://button[.="Import image"]
-    ELSE
-        Click Element  xpath://button[.="Import new image"]
-    END
-    Wait Until Page Contains    Import notebook images
+    Click Element  xpath://button[.="Import new image"]
+    Wait Until Page Contains    Import notebook image
```
lugi0, line 568: Are you sure in both cases (first image being imported and additional ones) the button says "import new image"?

ods_ci/tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -84,38 +85,30 @@ Test Bad Image URL
...
-    ${OG_URL}=  Set Variable  ${IMG_URL}
-    ${IMG_URL}=  Set Variable  randomstring
-    Set Global Variable  ${IMG_URL}  ${IMG_URL}
-    Create Custom Image
-    RHODS Notification Drawer Should Contain
-    ...  Unable to add notebook image ${IMG_NAME}
+    Wait Until Page Contains    Invalid repository URL: ${IMG_URL}
+    # Since the image cannot be created, we need to cancel the modal window now
+    Click Button    ${GENERIC_CANCEL_BTN_XP}
+    [Teardown]  Bad Image URL Teardown  orig_url=${OG_URL}
```
lugi0, line 91: why did you merge the two tests?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
jstourac: yes, I am sure; and the test is passing too :slightly_smiling_face:

ods_ci/tests/Tests/500__jupyterhub/custom-image.robot
jstourac: It's described in the commit message:

```
Since the behaviour of the custom image creation now is that you cannot
create an image with a wrong image location at all, this test simply
duplicates what is covered in `Test Bad Image URL` already.
```

Truth is that later I was able to create an image with image location like this:
```
image-registry.openshift-image-registry.svc:5000/redhat-ods-applications/s2i-generic-data-science-notebook
```
but such image reported error `"image" not found` and couldn't be enabled and used in JupyterHub anyway.

Thus I think that this scenario doesn't make sense anymore.

jgarciao (approved):

PR #1034, opened 2023-11-21
[Fix] test to check various user group access

jstourac:
The existing test checks that users from different groups have different
access and that this access can be setup via RHOAI User settings.

What this test does:

* creates a `custom-admins-group` with `TEST_USER2` user
* creates a `custom-users-group` with `TEST_USER3` user
* configures first group as admins in RHOAI User settings
* configures second group as users in RHOAI User settings

Once this was applied it tried to perform operations with three
different users:

* TEST_USER - expected no privileges and simple `Access
  permissions needed` page presented
* TEST_USER2 - all privileges
* TEST_USER3 - no Administration page shown

Problem in this approach is that the `TEST_USER` is part of the
`dedicated-admins` group, which has a `cluster-admin` role in our
default cluster settings. Per this [1], such user is an admin by
default.

So my changes incorporates usage of the `TEST_USER4` instead of
`TEST_USER`.

Also, unnecessary check for the product version is removed with this
commit.

[1] https://access.redhat.com/documentation/en-us/red_hat_openshift_data_science/1/html-single/managing_users_and_user_resources/index#defining-openshift-data-science-admin-and-user-groups_user-mgmt
```
By default, users with cluster admin permissions and users in the
dedicated-admins administrator group are OpenShift Data Science
administrators, but all users authenticated in OpenShift can access
OpenShift Data Science. A cluster admin is a superuser that can perform
any action in any project in the OpenShift cluster.
```

---

CI: rhods-ci-pr-test/2159

### DIFF
--- a/ods_ci/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot
+++ b/ods_ci/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot
@@ -103,20 +103,15 @@ Set Custom Access Groups
 Check New Access Configuration Works As Expected
     [Documentation]    Checks if the new access configuration (using two custom groups)
     ...                works as expected in JH
-    Launch Dashboard   ocp_user_name=${TEST_USER.USERNAME}  ocp_user_pw=${TEST_USER.PASSWORD}
-    ...    ocp_user_auth_type=${TEST_USER.AUTH_TYPE}    dashboard_url=${ODH_DASHBOARD_URL}
+    Launch Dashboard   ocp_user_name=${TEST_USER_4.USERNAME}  ocp_user_pw=${TEST_USER_4.PASSWORD}
+    ...    ocp_user_auth_type=${TEST_USER_4.AUTH_TYPE}    dashboard_url=${ODH_DASHBOARD_URL}
     ...    browser=${BROWSER.NAME}  browser_options=${BROWSER.OPTIONS}
     ...    expected_page=${NONE}    wait_for_cards=${FALSE}
-    ${version_check}=    Is RHODS Version Greater Or Equal Than    1.20.0
-    IF    ${version_check} == True
-        ${status}=    Run Keyword And Return Status     Launch Jupyter From RHODS Dashboard Link
-        Run Keyword And Continue On Failure    Should Be Equal    ${status}    ${FALSE}
-        Run Keyword And Continue On Failure    Page Should Contain    Access permissions needed
-        Run Keyword And Continue On Failure    Page Should Contain    ask your administrator to adjust your permissions.
-    ELSE
-        Launch Jupyter From RHODS Dashboard Link
-        Run Keyword And Continue On Failure   Verify Jupyter Access Level   expected_result=none
-    END
+    ${status}=    Run Keyword And Return Status     Launch Jupyter From RHODS Dashboard Link
+    Run Keyword And Continue On Failure    Should Be Equal    ${status}    ${FALSE}
+    Run Keyword And Continue On Failure    Page Should Contain    Access permissions needed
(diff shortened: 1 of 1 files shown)


diegolovison (approved):

ods_ci/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot
```diff
@@ -103,20 +103,15 @@ Set Custom Access Groups
 Check New Access Configuration Works As Expected
     [Documentation]    Checks if the new access configuration (using two custom groups)
     ...                works as expected in JH
-    Launch Dashboard   ocp_user_name=${TEST_USER.USERNAME}  ocp_user_pw=${TEST_USER.PASSWORD}
-    ...    ocp_user_auth_type=${TEST_USER.AUTH_TYPE}    dashboard_url=${ODH_DASHBOARD_URL}
+    Launch Dashboard   ocp_user_name=${TEST_USER4.USERNAME}  ocp_user_pw=${TEST_USER4.PASSWORD}
```
bdattoma, line 106: TEST_USER_4

bdattoma:
@jstourac 
> htpasswd-cluster-admin - expected no privileges and simple Access permissions needed page presented


I think this is not correct, because the test uses `TEST_USER` who is not htpasswd-cluster-admin

bdattoma (dismissed):

ods_ci/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot
jstourac: Uf, you're right, Berto; good catch:

```
TEST_USER:
  AUTH_TYPE: ldap-provider-qe
  USERNAME: ldap-admin1
```

I have it modified in my local properties settings, damn. 

Anyway, I've checked the cluster config for the `ldap-admin1` and the situation is same - it is part of the `dedicated-admins` group, which has a role of `cluster-admin` assigned. I'll update the description.

Not sure for how long this test been failing in which release this got changed.

bdattoma:
> > @jstourac
> > > htpasswd-cluster-admin - expected no privileges and simple Access permissions needed page presented
> > 
> > 
> > I think this is not correct, because the test uses `TEST_USER` who is not htpasswd-cluster-admin
> 
> Uf, you're right, Berto; good catch:
> 
> ```
> TEST_USER:
>   AUTH_TYPE: ldap-provider-qe
>   USERNAME: ldap-admin1
> ```
> 
> I have it modified in my local properties settings, damn.
> 
> Anyway, I've checked the cluster config for the `ldap-admin1` and the situation is same - it is part of the `dedicated-admins` group, which has a role of `cluster-admin` assigned. I'll update the description.
> 
> Not sure for how long this test been failing in which release this got changed.

yeah, this is actually an environement issue rather than test issue, because the script our CI uses to create IDPs assignes wrong permissions to ldap-admin users. We should fix it there, I haven't done it yet because I was working on a new script for IDPs (which is completed and used by interop team).

In the meantime I'm okay with fixing the test. You can have a look at how it's been done in `ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot` where there is a test setup which ensure the users have the right permissions: `Set User Groups For Testing` kw.

ods_ci/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot
jstourac: Oh, sorry, I didn't get it - I though it was related to our discussion that we resolved later; now I see it, will fix. Thanks guys!!!

jgarciao (approved):

lugi0 (approved):

PR #1039, opened 2023-11-24
install Dependent operator for kserve

tarukumar:


ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -117,6 +122,38 @@ Verify RHODS Installation
     ...                   label_selector=app.kubernetes.io/name=data-science-pipelines-operator
     ...                   timeout=400
   END
+  ${kserve} =    Is Component Enabled    kserve    ${DSC_NAME}
```
jgarciao, line 125: The `Verify RHODS Installation` keyword is very long.  Could this code be in a new keyword, like `Install Kserve Dependencies` or similar?

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -117,6 +122,38 @@ Verify RHODS Installation
...
+  IF    "${kserve}" == "true"
+        Install ISV Operator From OperatorHub Via CLI    operator_name=${SERVICEMESH_OP_NAME}
+        ...    subscription_name=${SERVICEMESH_SUB_NAME}
+        ...    catalog_source_name=redhat-operators
+        Wait Until Operator Subscription Last Condition Is
+        ...    type=CatalogSourcesUnhealthy    status=False
+        ...    reason=AllCatalogSourcesHealthy    subcription_name=${SERVICEMESH_SUB_NAME}
```
bdattoma, line 132: probably we could `Wait For Pods To Be Ready    label_selector=name=istio-operator
    ...    namespace=${DEFAULT_OP_NS}`

bdattoma:
lgtm, I would just like to double check about namespace creation, I kind of remember we must create the istio-system NS manually. We could skip doing it in case rhoai operator does it. But we should also check if rhoai operator doesn't try to create the Serverless namespace (which is created by the code in this PR)

bdattoma (approved):

jgarciao (approved):

PR #1047, opened 2023-11-28
Add bucket_name parameter in upload_download sample pipeline

diegolovison:
Please keep the two commits

Compiled pipeline was tested ( there is no automated test for that )
image

### DIFF
--- a/ods_ci/tests/Resources/Files/redhat-openshift-pipelines.yaml
+++ b/ods_ci/tests/Resources/Files/redhat-openshift-pipelines.yaml
@@ -4,7 +4,7 @@ metadata:
   name: openshift-pipelines-operator-rh
   namespace: openshift-operators
 spec:
-  channel: pipelines-1.11
+  channel: pipelines-1.12
   installPlanApproval: Automatic
   name: openshift-pipelines-operator-rh
   source: redhat-operators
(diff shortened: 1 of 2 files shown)


ods_ci/tests/Resources/Files/redhat-openshift-pipelines.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: openshift-pipelines-operator-rh
   namespace: openshift-operators
 spec:
-  channel: pipelines-1.11
+  channel: pipelines-1.12
```
jstourac, line 7: I think this shouldn't be changed?

diegolovison: We store the latest version and update the value during the test by following `OpenShiftPipelines.resource`

jstourac: Yes, I saw that there is a specific script that updates this value. Thus I presumed that the change here doesn't matter and shouldn't be part of the PR. Actually, the value here shouldn't matter, since it's always updated, is it correct?

I don't insist, just trying to understand whether it makes sense to update also this one.

jgarciao (approved):

ods_ci/tests/Resources/Files/redhat-openshift-pipelines.yaml
diegolovison: > shouldn't be part of the PR?

Just curious why.

> the value here shouldn't matter, since it's always updated, is it correct?

In theory, it doesn't matter. But we keep it consistent with the latest version.

diegolovison: > isn't necessary to incorporate unnecessary changes into our history

Didn't understand this point

jstourac: unnecessary changes in the files make it harder to find for particular changes that makes sense in case of debugging or searching for the culprit of an issue... as I wrote at the beginning - since this file is updated automatically in our scripts anyway, I didn't see the reason to update it, that's why :shrug:

PR #1049, opened 2023-11-28
[Fix] for datascience project tests for 2.5

jstourac:
The element for the data science project overview has been moved slightly.

Also, the Window title is renamed now to Red Hat OpenShift AI.

---

CI: TODO

PR #1050, opened 2023-11-28
[Fix] for the window title to accomodate with RHOAI 2.5

jstourac:
This is the very same update for the Window name as was done as a part
of the fixes in #1049 [1].

[1] ef44c42000d1f717b74b3d2a760ea5a980a9bb0d

---

CI: didn't test it, please let me know if you insist


jgarciao (approved):

jiridanek:
> Verify That Idle JupyterLab Servers Are Culled In Sandbox Environment After 24h

Yeah, right!

jiridanek (approved):

PR #1053, opened 2023-11-29
Skip pre-req install by default in the KServe test suite

bdattoma:
To avoid conflicts after https://github.com/red-hat-data-services/ods-ci/pull/1039 got merged, we should skip installing the pre-requirements in the KServe suite. 
The suite can still install them by setting `SKIP_PREREQS_INSTALL` variable to `False` at runtime


aloganat (approved):

jstourac:
Isn't it possible to check whether these prerequisities are installed already? :thinking:

lugi0 (approved):

bdattoma:
> Isn't it possible to check whether these prerequisities are installed already? 🤔

yes it's possible. We could implement a check in future PRs, it would take time we don't have now and the benefit would be minimal

jstourac:
> yes it's possible. We could implement a check in future PRs, it would take time we don't have now and the benefit would be minimal

Maybe I'm missing something but I don't agree that the benefit would be minimal :slightly_smiling_face:  Anyway, I agree that we need this fast.

jstourac (approved):

bdattoma:
> > yes it's possible. We could implement a check in future PRs, it would take time we don't have now and the benefit would be minimal
> 
> Maybe I'm missing something but I don't agree that the benefit would be minimal 🙂 Anyway, I agree that we need this fast.

we can talk offline

PR #1059, opened 2023-12-03
Revert commit #861 that broke creation of kubeconfig

manosnoam:
In #861 the temporary kubeconfig was disabled, and the permanent kubeconfig was used. 
That broke the generation of kubeconfig from the new cluster credentials, which was created first as temporary file (with the `oc login`), and then as permanent file (with the `oc status`).


manosnoam:
@kobihk This should fix the issue you faced on rhods-smoke/4078 where the kubeconfig that was created was empty, and thus the installation the cluster was deprovisioned.

PR #1062, opened 2023-12-06
Add Kserve UI tests

lugi0:
Testing in rhods-ci-pr-test/2178

Adds a set of basic tests for Kserve deployed models using the RHOAI UI

### DIFF
--- a/ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm_UI.robot
+++ b/ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm_UI.robot
@@ -86,14 +86,14 @@ Verify User Can Deploy Multiple Models In Different Namespaces Using The UI
     ...    bloom-560m/${model_one_name}
     Wait For Pods To Be Ready    label_selector=serving.kserve.io/inferenceservice=${model_one_name}
     ...    namespace=singlemodel-multi1
+    Query Model Multiple Times    endpoint=${CAIKIT_ALLTOKENS_ENDPOINT_HTTP}    model_name=${model_one_name}    n_times=2
+    ...    namespace=singlemodel-multi1    protocol=http
     Open Data Science Projects Home Page
     Set Up Project    namespace=singlemodel-multi2    single_prj=${FALSE}    dc_name=kserve-connection-2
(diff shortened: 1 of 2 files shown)


tarukumar:
Generally, it appears good. However, we might consider updating the keyword and refining the inference related to the workbench name, even when we select all projects in the dropdown. This could be enhance we can do in the future

lugi0:
@jstourac no I will update with the correct Polarion IDs ASAP before merging

jstourac (commented):
I have very little knowledge about the topic, but in general this LGTM (even though, there are a handful of robot warnings that could be easily fixed) :upside_down_face: :slightly_smiling_face:

lugi0:
Validated once in rhods-ci-pr-test/2187, rerunning in rhods-ci-pr-test/2188 after last commit to ensure nothing broke

lugi0:
Obviously everything broke :) Retesting for hopefully the last time: rhods-ci-pr-test/2189

lugi0:
verified in rhods-ci-pr-test/2191

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm_UI.robot
```diff
@@ -0,0 +1,572 @@
...
+    Set Suite Variable    ${EXP_RESPONSES}    ${exp_responses}
+
+Set Up Project
+    [Documentation]    Creates the DS Project (if not exists), creates the data connection for the models,
+    ...                creates caikit runtime. This can be used as test setup
+    [Arguments]    ${namespace}    ${single_prj}=${TRUE}    ${enable_metrics}=${FALSE}    ${dc_name}=kserve-connection
+    Open Model Serving Home Page
+    IF    ${single_prj}
+        Try Opening Create Server
+        Wait for RHODS Dashboard to Load    wait_for_cards=${FALSE}    expected_page=Data Science Projects
```
bdattoma, line 295: why do we need to handle this case?

lugi0: `Try Opening Create Server` deletes all existing projects to test with only a single DS Project present - This way we can force multiple projects at the same time by specifying `singel_prj=${FALSE}`

PR #1063, opened 2023-12-07
[Fix] capital letters for the particular menu sections

jstourac:
These changes are based on this fix into ODH-Dashboard component [1,2].

Note: there is still an inconsistency for `Model Serving` menu item and relevant page title `Deployed models` which is gonna be handled by [3].

* [1] https://github.com/opendatahub-io/odh-dashboard/pull/2067
* [2] https://github.com/opendatahub-io/odh-dashboard/issues/1907
* [3] https://github.com/opendatahub-io/odh-dashboard/issues/2105

---

Maybe it would be beneficial to have a separate test for these menu and page title strings where the upper/lower case would be checked and use something like this `...lower-case(@title)...` in the rest of our tests. But truth is that sometimes not only case is changed so probably not worth it.

---

CI: rhods-ci-pr-test/2196

### DIFF
--- a/ods_ci/tests/Resources/ODS.robot
+++ b/ods_ci/tests/Resources/ODS.robot
@@ -62,20 +62,20 @@ Get "Usage Data Collection" Key
 
 Is Usage Data Collection Enabled
     [Documentation]    Returns a boolean with the value of configmap odh-segment-key-config > segmentKeyEnabled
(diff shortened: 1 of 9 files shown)


FedeAlonso (approved):

jstourac:
Thank you guys! Merged :slightly_smiling_face:

PR #1064, opened 2023-12-07
[Fix] the test that checks the User/Group privileges permissions

jstourac:
This fixes the `Verify Unauthorized User Is Not Able To Spawn Jupyter Notebook`
test. It is basically similar change as done in [1,2].

Note: looks like the behavior of the RHOAI is a bit different now - the
user without permissions is shown directly the `Access permissions
needed` page. Nothing else is shown. Based on the test structure, in
previous versions we could see the application but weren't able to start
a JupyterHub tile in the `Applications -> Enabled` menu.

Also, unnecessary check for the product version is removed with this
commit and a small typo fix.

* [1] https://github.com/red-hat-data-services/ods-ci/pull/1034
* [2] https://github.com/jstourac/ods-ci/commit/c719b70d93c88734288ae064fefc71991ccb236a

---

CI: rhods-ci-pr-test/2244 :construction:

### DIFF
--- a/ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
+++ b/ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
@@ -49,28 +49,30 @@ Verify If Unauthorized User Can Not Change The Permission
 Verify Unauthorized User Is Not Able To Spawn Jupyter Notebook
     [Documentation]    Verify unauthorized User Is Not Able To Spawn Jupyter
     ...     Notebook , user should not see a spawner if the user is not in allowed Groups
+    ...     Note: this test configures user access via ODH Dashboard UI in setting appropriate
+    ...     groups in `Settings -> User management` section. There is an another test that changes
+    ...     users/groups via `oc adm groups` command,see: `Verify User Can Set Custom RHODS Groups`
+    ...     in ods_ci/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot
     [Tags]  ODS-1680
     ...     Tier1
     ...     Sanity
     Launch Dashboard And Check User Management Option Is Available For The User   ${TEST_USER.USERNAME}   ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
     Clear User Management Settings
-    Add OpenShift Groups To Data Science Administrators    rhods-users
-    Add OpenShift Groups To Data Science User Groups       rhods-users
+    Add OpenShift Groups To Data Science Administrators    rhods-admins
+    Add OpenShift Groups To Data Science User Groups       rhods-admins
     Save Changes In User Management Setting
-    AdminGroups In OdhDashboardConfig CRD Should Be        rhods-users
-    AllowedGroups In OdhDashboardConfig CRD Should Be      rhods-users
-    Reload Page
-    ${version_check}=    Is RHODS Version Greater Or Equal Than    1.20.0
-    IF    ${version_check} == True
-        ${status}=    Run Keyword And Return Status     Launch Jupyter From RHODS Dashboard Link
-        Run Keyword And Continue On Failure    Should Be Equal    ${status}    ${FALSE}
-        Run Keyword And Continue On Failure    Page Should Contain    Access permissions needed
-        Run Keyword And Continue On Failure    Page Should Contain    ask your administrator to adjust your permissions.
-    ELSE
-        Menu.Navigate To Page    Applications    Enabled
-        Launch Jupyter From RHODS Dashboard Link
-        Run Keyword And Continue On Failure   Verify Jupyter Access Level   expected_result=none
-    END
+    AdminGroups In OdhDashboardConfig CRD Should Be        rhods-admins
+    AllowedGroups In OdhDashboardConfig CRD Should Be      rhods-admins
+    Logout From RHODS Dashboard
+    Login To RHODS Dashboard    ${TEST_USER_4.USERNAME}    ${TEST_USER_4.PASSWORD}    ${TEST_USER_4.AUTH_TYPE}
+    Wait for RHODS Dashboard to Load    expected_page=${NONE}    wait_for_cards=${FALSE}
+    Run Keyword And Continue On Failure    Page Should Contain    Access permissions needed
+    Run Keyword And Continue On Failure    Page Should Contain    ask your administrator to adjust your permissions.
+    # Let's check that we're not allowed to also access the spawner page directly navigating the browser there
+    Go To    ${ODH_DASHBOARD_URL}/notebookController/spawner
+    Wait for RHODS Dashboard to Load    expected_page=${NONE}    wait_for_cards=${FALSE}
+    Run Keyword And Continue On Failure    Page Should Contain    Access permissions needed
+    Run Keyword And Continue On Failure    Page Should Contain    ask your administrator to adjust your permissions.
     [Teardown]  Revert Changes To Access Configuration
 
 Verify Automatically Detects a Group Selected Is Removed and Notify the User
--- a/ods_ci/tests/Resources/ODS.robot
+++ b/ods_ci/tests/Resources/ODS.robot
@@ -102,7 +102,7 @@ Set Access Groups Settings
     [Documentation]    Changes the rhods-groups config map to set the new access configuration
     [Arguments]     ${admins_group}   ${users_group}
     ${return_code}    ${output}    Run And Return Rc And Output    oc patch OdhDashboardConfig odh-dashboard-config -n ${APPLICATIONS_NAMESPACE} --type=merge -p '{"spec": {"groupsConfig": {"adminGroups": "${admins_group}","allowedGroups": "${users_group}"}}}'   #robocop:disable
-    Should Be Equal As Integers	${return_code}	 0    msg=Pathc failed
+    Should Be Equal As Integers	${return_code}	 0    msg=Patch to group settings failed
 
 Set Default Access Groups Settings
     [Documentation]    Restores the default rhods-groups config map
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboardSettings.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboardSettings.resource
@@ -69,7 +69,7 @@ Save Changes In User Management Setting
     [Documentation]  Save User Management Settings In Dashboard
     Press Keys    None    ESC
     Click Button    Save changes
-    Sleep  60s
+    Sleep    120s    reason=Wait for Dashboard to apply the updated configuration...
 
 AdminGroups In OdhDashboardConfig CRD Should Be
     [Documentation]  Verify Expect Changes Are Present In CRD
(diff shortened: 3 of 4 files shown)


lugi0:
The difference in the two tests is that the dashboard one updates the groups via the UI in the admin settings of the dashboard itself, while the test in Jupyter uses `oc` to do it - I'd personally keep them both since they go through two different (but valid) flows. If we had to remove one I'd remove the Jupyter one, since using `oc` is not the documented way to do this.

jstourac:
Thank you Luca! I'll check again and will be back then :slightly_smiling_face:

jstourac:
Okay, I've updated the test instead of deleting it. I'm keeping this as a draft for now, since I want to assure I'm not missing anything. Now the test is written so that it checks that unauthorized user is able to login but `Access permissions denied` page is shown directly to him - no menu with `Applications -> Enabled`, etc. (which the previous test expected probably), see this page that is shown to `TEST_USER4`:

Screenshot from 2023-12-12 21-32-18

Am I missing something WRT the groups that we can configure in our product?

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
```diff
@@ -49,28 +49,29 @@ Verify If Unauthorized User Can Not Change The Permission
...
+    AllowedGroups In OdhDashboardConfig CRD Should Be      rhods-admins
+    Logout From RHODS Dashboard
+    Launch Dashboard    ocp_user_name=${TEST_USER_4.USERNAME}    ocp_user_pw=${TEST_USER_4.PASSWORD}
+    ...    ocp_user_auth_type=${TEST_USER_4.AUTH_TYPE}    dashboard_url=${ODH_DASHBOARD_URL}
+    ...    browser=${BROWSER.NAME}    browser_options=${BROWSER.OPTIONS}
+    ...    expected_page=${NONE}      wait_for_cards=${FALSE}
+    # ${status}=    Run Keyword And Return Status     Launch Jupyter From RHODS Dashboard Link
+    # Run Keyword And Continue On Failure    Should Be Equal    ${status}    ${FALSE}
+    Run Keyword And Continue On Failure    Page Should Contain    Access permissions needed
+    Run Keyword And Continue On Failure    Page Should Contain    ask your administrator to adjust your permissions.
```
lugi0, line 74: You open a new browser but never close it, can you add a final step to close the browser and switch the context to the previous one?

lugi0 (approved):
LGTM but I'd rather not leave an additional browser lying around

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
jstourac: Uff, very well spotted, Luca! Thanks for this. I'll update.

jstourac:
> LGTM but I'd rather not leave an additional browser lying around

@lugi0 Thank you for your review, Luca! Just to be sure - that means you agree also with the slight change of what the test does as I described above? Didn't I miss anything, right?

lugi0 (dismissed):

lugi0:
Yes @jstourac the rest of the changes look good to me

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
```diff
@@ -49,28 +49,25 @@ Verify If Unauthorized User Can Not Change The Permission
 Verify Unauthorized User Is Not Able To Spawn Jupyter Notebook
     [Documentation]    Verify unauthorized User Is Not Able To Spawn Jupyter
     ...     Notebook , user should not see a spawner if the user is not in allowed Groups
+    ...     Note: this test creates users/groups via ODH Dashboard UI in
```
bdattoma, line 52: it doesn't create users nor groups, it configures the user access of the dashboard

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
```diff
@@ -49,28 +49,25 @@ Verify If Unauthorized User Can Not Change The Permission
...
+    ...     in ods_ci/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot
     [Tags]  ODS-1680
     ...     Tier1
     ...     Sanity
     Launch Dashboard And Check User Management Option Is Available For The User   ${TEST_USER.USERNAME}   ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
     Clear User Management Settings
-    Add OpenShift Groups To Data Science Administrators    rhods-users
-    Add OpenShift Groups To Data Science User Groups       rhods-users
+    Add OpenShift Groups To Data Science Administrators    rhods-admins
+    Add OpenShift Groups To Data Science User Groups       rhods-admins
```
bdattoma, line 62: this won't likely solve the test problem, because by default RHODS has system:authenticated group as RHODS Users

bdattoma (changes requested):

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
lugi0: I think you're right - adding `Remove OpenShift Groups From Data Science User Groups    system:authenticated` before/after L62 should fix this

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
```diff
@@ -49,28 +49,25 @@ Verify If Unauthorized User Can Not Change The Permission
...
+    ...     users/groups via `oc adm groups` command,see: `Verify User Can Set Custom RHODS Groups`
+    ...     in ods_ci/tests/Tests/500__jupyterhub/jupyterhub-user-access.robot
     [Tags]  ODS-1680
     ...     Tier1
     ...     Sanity
     Launch Dashboard And Check User Management Option Is Available For The User   ${TEST_USER.USERNAME}   ${TEST_USER.PASSWORD}  ${TEST_USER.AUTH_TYPE}
     Clear User Management Settings
-    Add OpenShift Groups To Data Science Administrators    rhods-users
-    Add OpenShift Groups To Data Science User Groups       rhods-users
+    Add OpenShift Groups To Data Science Administrators    rhods-admins
```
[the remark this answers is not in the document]
jstourac: It is, I can remove it if that is a problem but I didn't see a problem to keep it there.

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
jstourac: ah, yeah - I'll reword that, good catch, thanks!

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
jstourac: The test passed just fine with this :thinking:

bdattoma: @jstourac can you check the user mgmt settings in the Dashboard of your cluster?

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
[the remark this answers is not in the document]
bdattoma: there's no problem I guess, but not needed. Feel free to leave it

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
jstourac: Okay, I know the reason - there is called `Clear User Management Settings` right above this. Didn't remember so I had to look into the code again.

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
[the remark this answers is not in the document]
jstourac: Actually, this line cannot be removed because there is `Clear User Management Settings` called just above this.

jstourac:
Thank you, guys, very much for your reviews! Ready for another review :slightly_smiling_face:

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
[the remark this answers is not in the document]
bdattoma: ah, you're right!

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
bdattoma: sounds good then. Thanks!

ods_ci/tests/Tests/400__ods_dashboard/413__ods_dashboard_user_mgmt.robot
```diff
@@ -49,28 +49,25 @@ Verify If Unauthorized User Can Not Change The Permission
...
-        Launch Jupyter From RHODS Dashboard Link
-        Run Keyword And Continue On Failure   Verify Jupyter Access Level   expected_result=none
-    END
+    AdminGroups In OdhDashboardConfig CRD Should Be        rhods-admins
+    AllowedGroups In OdhDashboardConfig CRD Should Be      rhods-admins
+    Logout From RHODS Dashboard
+    Login To RHODS Dashboard    ${TEST_USER_4.USERNAME}    ${TEST_USER_4.PASSWORD}    ${TEST_USER_4.AUTH_TYPE}
+    Wait for RHODS Dashboard to Load    expected_page=${NONE}    wait_for_cards=${FALSE}
+    Run Keyword And Continue On Failure    Page Should Contain    Access permissions needed
+    Run Keyword And Continue On Failure    Page Should Contain    ask your administrator to adjust your permissions.
```
bdattoma, line 70: given that the test case cares about Jupyter notebooks explicitly, what about adding a check at the end of the test to see if the user is preventing to access `<dashabord_url>/notebookController/spawner`?

jstourac: Well, as I wrote above - the behavior of the product probably changed, because now the `Access permission needed` is shown on whole page without any menu or anything. Or I simply don't know how to mimic original intended behavior and would be glad if you point me how to.

Referring to [this comment](https://github.com/red-hat-data-services/ods-ci/pull/1064#issuecomment-1852765867).

bdattoma: yes, behavior changed indeed. 
And as workaround I'm proposing to try reaching the jupyter notebook launcher via URL instead of navigating the UI

jstourac: Ah, I see now, I'll try.

jstourac: Updated :slightly_smiling_face:

bdattoma (approved):

lugi0 (approved):
LGTM, thanks Jan!

jstourac:
Merged, thank you very much for your reviews and all your comments, guys! You brought up very good points indeed :slightly_smiling_face:

PR #1068, opened 2023-12-08
Update oc_install.robot

tarukumar:
temp fix for gcp failure

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -307,10 +307,6 @@ Install Kserve Dependencies
     ...    operator_group_name=serverless-operators
     ...    operator_group_ns=${SERVERLESS_NS}
     ...    operator_group_target_ns=${NONE}
-    Wait Until Operator Subscription Last Condition Is
-    ...    type=CatalogSourcesUnhealthy    status=False
-    ...    reason=AllCatalogSourcesHealthy    subcription_name=${SERVERLESS_SUB_NAME}
-    ...    namespace=${SERVERLESS_NS}
     Wait For Pods To Be Ready    label_selector=name=knative-openshift
     ...    namespace=${SERVERLESS_NS}
     Wait For Pods To Be Ready    label_selector=name=knative-openshift-ingress


mwaykole (approved):

jstourac (dismissed):
I don't think we should merge this as it is now.

1. This doesn't solve the issue completely as there is another `Wait Until Operator Subscription Last Condition Is` keyword few lines above - and it failed there actually - this one is complete https://github.com/jstourac/ods-ci/commit/fcccada99de6f20f851cf2874cf8c98a1414b45f
2. We should rather check whether the operator dependencies need to be installed before subscribing? To avoid such issue in the future, WDYT?

jstourac:
@tarukumar, okay, wasn't aware that this is just GCP specific; understand. Would be great to check why it behaves differently, though.

Anyway, I still believe that checking whether the dependency operators are installed before actually trying it to install them is still valuable (I suppose it may cause a trouble if the operator is present on the cluster already and we try to install it again - well, at least it was the situation in our case with GCP).

Just to be sure we're on the same page - this PR shouldn't be merged as is, definitely, 'cause it doesn't workaround the issue we saw in the GCP, see my commit I linked in my previous comment.

tarukumar:
> @tarukumar, okay, wasn't aware that this is just GCP specific; understand. Would be great to check why it behaves differently, though.
> 
> Anyway, I still believe that checking whether the dependency operators are installed before actually trying it to install them is still valuable.
> 
> Just to be sure we're on the same page - this PR shouldn't be merged as is, definitely, 'cause it doesn't workaround the issue we saw in the GCP, see my commit I linked in my previous comment.

That was intentional to leave that. Also, as i mentioned earlier yes i agree we need to add dependent operator verification and they will come in future not in this pr

jstourac:
Changes in this PR has been merged together in #1069.

jstourac:
> That was intentional to leave that. Also, as i mentioned earlier yes i agree we need to add dependent operator verification and they will come in future not in this pr
* Intentional to leave that - okay, then I suppose that you were trying to solve a different problem than me which I saw last Friday then.
* Regarding the check for dependencies - all right, I probably didn't understand that fully in your previous comment. Okay then.

PR #1070, opened 2023-12-09
[Fix] to distinguish between PVC in pending and PVC in bound status

jstourac:
The storage size is shown in the UI depending of the PVC status. Before the PVC is bounded, there is shown 'Max XXGi' (or other unit). After the PVC has been bound (even after it has been de-attached from a workbench)
then there is something like: 'XXGi [##     ] YYGi' - so there is shown
both the capacity that is taken and total capacity of the storage; in
the middle of them, there is a graphical element to show how much the
capacity is used.

---

CI: **rhods-ci-pr-test/2198**  - pass of this test should be enough since this method is used only in one other test and that one is tagged with `AutomationBug` so it's irrelevant for now anyway.

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
@@ -47,12 +47,46 @@ Storage Size Should Be
     Run Keyword And Continue On Failure     Wait Until Page Contains Element    ${STORAGE_SECTION_XP}
     Click Element    ${STORAGE_SECTION_XP}//tr[td//*/div[text()="${name}"]]//button[@aria-label="Details"]
     Wait Until Element Is Visible    ${STORAGE_SECTION_XP}//tr[@class="pf-c-table__expandable-row pf-m-expanded"]
-    Wait Until Page Contains Element    ${STORAGE_SECTION_XP}//tr[@class="pf-c-table__expandable-row pf-m-expanded"]/td/div[strong[text()="Size"]]/div    timeout=45s
-    ${displayed_size}=      Get Text    ${STORAGE_SECTION_XP}//tr[@class="pf-c-table__expandable-row pf-m-expanded"]/td/div[strong[text()="Size"]]/div
-    Run Keyword And Continue On Failure     Should Be Equal As Strings    ${displayed_size}    Max ${size}Gi
-    ${rc}  ${oc_object_size}=    Run And Return Rc And Output   oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].spec.resources.requests.storage}'
-    Run Keyword And Continue On Failure     Should Be Equal As Strings    ${oc_object_size}    ${size}Gi
-    ${rc}  ${oc_object_size}=    Run And Return Rc And Output   oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].status.capacity.storage}'
+
+    ${rc}  ${pvc_status_phase}=    Run And Return Rc And Output
+    ...    oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].status.phase}'    # robocop: disable
+    Should Be Equal As Integers    ${rc}    0    An error occurred during the check of PVC ${name} .status.phase value!
+    IF    "${pvc_status_phase}" == "Pending"
+        # PVC hasn't been bound yet - only simple text with max storage should be shown
+        ${storage_size_el}=    Set Variable
+        ...    ${STORAGE_SECTION_XP}//tr[@class="pf-c-table__expandable-row pf-m-expanded"]/td/div[strong[text()="Size"]]/div    # robocop: disable
+        Wait Until Page Contains Element    ${storage_size_el}    timeout=20s
+        ${displayed_size}=    Get Text    ${storage_size_el}
+        Run Keyword And Continue On Failure    Should Be Equal As Strings    ${displayed_size}    Max ${size}Gi
+    ELSE IF    "${pvc_status_phase}" == "Bound"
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
```diff
@@ -47,12 +47,39 @@ Storage Size Should Be
...
+        Run Keyword And Continue On Failure     Should Be Equal As Strings    ${displayed_size}    ${size}Gi
+        ${rc}  ${oc_object_size}=    Run And Return Rc And Output
+        ...    oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].status.capacity.storage}'    # robocop: disable
+        Should Be Equal As Integers    ${rc}    0
+        ...    An error occurred during the check of PVC ${name} '.status.capacity.storage' value!
+        Run Keyword And Continue On Failure     Should Be Equal As Strings    ${oc_object_size}    ${size}Gi
+    END
+    ${rc}  ${oc_object_size}=    Run And Return Rc And Output
+    ...    oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].spec.resources.requests.storage}'    # robocop: disable
+    Should Be Equal As Integers    ${rc}    0
```
lugi0, line 81: What is this check used for?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
```diff
@@ -47,12 +47,39 @@ Storage Size Should Be
...
-    ${displayed_size}=      Get Text    ${STORAGE_SECTION_XP}//tr[@class="pf-c-table__expandable-row pf-m-expanded"]/td/div[strong[text()="Size"]]/div
-    Run Keyword And Continue On Failure     Should Be Equal As Strings    ${displayed_size}    Max ${size}Gi
-    ${rc}  ${oc_object_size}=    Run And Return Rc And Output   oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].spec.resources.requests.storage}'
-    Run Keyword And Continue On Failure     Should Be Equal As Strings    ${oc_object_size}    ${size}Gi
-    ${rc}  ${oc_object_size}=    Run And Return Rc And Output   oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].status.capacity.storage}'
+
+    ${rc}  ${pvc_status_phase}=    Run And Return Rc And Output
+    ...    oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].status.phase}'    # robocop: disable
+    Should Be Equal As Integers    ${rc}    0    An error occurred during the check of PVC ${name} .status.phase value!
+    IF    "${pvc_status_phase}" == "Pending"
```
lugi0, line 54: Are the only options "pending" or a success state? Is there no failed state?
IMHO it'd be safer to do IF==Pending, ELIF==Bound, ELSE to make sure any unexpected state is caught on its own, and we can log it together with an error/warning message.

lugi0 (dismissed):
In general it looks ok to me, but I'd like to understand a bit more what the last check on the PVC does after the if/else block. Perhaps add a line or two of comments to explain the flow?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
jstourac: Hm, I was thinking about it too but then was too lazy to implement :upside_down_face:  . But I guess you're right, I'll amend the code to suite also situation when the state isn't expected one :+1:

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
jstourac: Oh, okay - understand your question now!

So, this check was there even before - there were two `oc get pvc` checks before my changes, but this one is common for both "Pending" and "Bound" PVC statuses, whether the one that is in the ELSE branch now makes sense only for "Bound" case.

I'll add some comment there.

jstourac:
@lugi0 Updated addressing both your comments. Will re-run CI now. Please let me know whether it looks good now :slightly_smiling_face:

lugi0 (dismissed):
thanks @jstourac , LGTM

jstourac:
I just noticed a robocop warning so I rebased and pushed [this change](https://github.com/red-hat-data-services/ods-ci/compare/d337f74de29fcb0c6e80ef05e5bdf3c9b5e4abe1..cd237b3a9c0ef9bb3ba9294ad275824404f92d1f) only.

bdattoma:
@jstourac LGTM! thanks for raising this one.
Only one point/question for you.

In my mind, we should be able to test all the situations (pending, success, failure) explicitly. This fix is great but we are not assuming an expected behavior in the test, while IMO we should if we care of this particular UI aspect.
Of course, if we don't care about it, we can ignore it and let the kw to handle the 3 cases as done in this PR.
The risk is that we don't capture a regression or change in the UI behavior because the kw silently handles the situation.

Wdyt?

(i'm approving the PR because it would let our tests to run smoothly for now, and I'd agree that this particular aspect of the UI doesn't have the highest priority for us)

bdattoma (approved):

jstourac:
@bdattoma thank you for your review! Yeah, I agree that we should probably extend test to assume the PVC state there and extend to check also the `Pending` state at least. Let's do it in a separate PR though. Do you want me to add a Todo note to the test?

bdattoma:
> Do you want me to add a Todo note to the test?


up to you, to me it would also be okay just to have in our Jira backlog

jstourac:
@bdattoma https://issues.redhat.com/browse/RHOAIENG-821

jgarciao (approved):

PR #1075, opened 2023-12-12
Add e2e test for Caikit Nlp Client

bdattoma:
This is a PR for testing https://github.com/opendatahub-io/caikit-nlp-client against models which are deployed using RHOAI single-model serving stack.

The test code was originally written in a python script in my local copy of the caikit-nlp repo for an ease and faster execution. Now I'm moving it to ods-ci.

**UPDATE 21st Dec**
Due to an incompatibility between the python libraries caikit-nlp-client and kfp, we cannot add the former in our poetry.toml. We have tried with engineers to relax the caikit-nlp-client dependency requirements but it breaks its functioning.
It means that I had to remove the CLI tests I previously created and replace it (check below section)
- the dependency conflict is caused by opposite requirement for protobuf library, i.e., former requires >= 4.22 while the latter requires < 4
- there is a recent draft PR upstream which would solve the issue: https://github.com/kubeflow/pipelines/pull/10307. 
When it gets merged, we could think of re-integrate the CLI tests

**_Solution_**
- create a Jupyter notebook to send queries to the models
- create a workbench
- upload the jupyter notebook (and the generated certificates) in the workbench
- execute the workbench
- check there are no errors 

_**Notes**_
- some refactoring is applied to move model serving keywords into common resource files
- python didn't get upgrade anymore as part of this PR
- the original version of the tests in this PR can still be found on one of my branches  [bdattoma:caikit-py-client-pythontests](https://github.com/bdattoma/ods-ci/blob/f/caikit-py-client-pythontests/ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_caikit_py_client.robot)


ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__caikit_py_client.py
```diff
@@ -0,0 +1,67 @@
+from caikit_nlp_client import HttpClient, GrpcClient
```
bdattoma, line 1: temp file: to be used if you need/want to run tests against caikit-nlp-client library

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/gen_client_certs.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
bdattoma, line 1: temp file: to be used if you need/want to run tests against caikit-nlp-client library

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/get_cert.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
```
bdattoma, line 1: temp file: to be used if you need/want to run tests against caikit-nlp-client library

ods_ci/libs/CaikitPythonClient.py
```diff
@@ -0,0 +1,55 @@
...
+
+class CaikitPythonClient:
+    """keywords to test https://github.com/opendatahub-io/caikit-nlp-client"""
+    def __init__(self):
+        self.BuiltIn = BuiltIn()
+        self.self.client = None
+
+    @keyword
+    def get_http_client_without_ssl_validation(self, host, port):
+        self.self.client = HttpClient(f"https://{host}:{port}", verify=False)
```
lugi0, line 15: too many `self`?

bdattoma: yeah I saw yday. i thought I already pushed the commit with the fix, but I didn't xD will do

bdattoma:
this one cannot be tested on CI yet for some pkg incompatibility. Will align with @aloganat

bdattoma:
I will add the missing Documentation tags and PR validation later

lugi0 (dismissed):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -24,24 +25,29 @@ ${MS_TABLE_STATUS_SUCCESS}=    //span[contains(@class,"pf-c-icon__content")][con
...
-    ${loading_error} =    Run Keyword And Return Status    Page Should Contain    Problem loading model serving page
+    ${loading_error}=    Run Keyword And Return Status
+    ...    SeleniumLibrary.Page Should Contain    Problem loading model serving page
     IF    ${loading_error}
         Log    "Page load error encountered"    level=WARN
-        Capture Page Screenshot
+        SeleniumLibrary.Capture Page Screenshot
         Sleep    5s
-        Reload Page
+        SeleniumLibrary.Reload Page
```
FedeAlonso, line 46: Why using "SeleniumLibrary." in some of the keywords and imports ? (not only in this file)
Do we have some SeleniumLibrary keywords duplicated?

FedeAlonso (changes requested):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
bdattoma: there is duplication between SeleniumLibrary and JupyterLibrary. It works fine because we select the order of usage in the Suite setup, but last year it was discussed about it and my understanding is that the final decision was to explicitly call the library name to avoid IDEs to complain.

FedeAlonso (approved):

jgarciao (approved):

PR #1077, opened 2023-12-13
After installation, verify the monitoring stack only for managed

jgarciao:
- Starting at RHOAI 2.5 the modelmesh monitoring stack is no longer deployed in redhat-ods-monitoring. 
- Disabling this check for Self-Managed will save 20 mins in RHOAI Self-Managed installation
- Note: clusters upgrading 2.4 > 2.5 still will have the modelserving monitoring stack until https://issues.redhat.com/browse/RHOAIENG-293 is solved

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
@@ -18,7 +18,7 @@ Installing RHODS Operator ${image_url}
   IF  ${is_codeflare_managed}    Installing CodeFlare Operator
 
 RHODS Operator Should Be installed
-  Verify RHODS Installation
+  Verify RHODS Installation    ${cluster_type}
   ${version} =  Get RHODS Version
   Set Global Variable  ${RHODS_VERSION}  ${version}
   Log  RHODS has been installed  console=yes
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
@@ -18,7 +18,7 @@ Installing RHODS Operator ${image_url}
   IF  ${is_codeflare_managed}    Installing CodeFlare Operator
 
 RHODS Operator Should Be installed
-  Verify RHODS Installation    ${cluster_type}
+  Verify RHODS Installation
   ${version} =  Get RHODS Version
   Set Global Variable  ${RHODS_VERSION}  ${version}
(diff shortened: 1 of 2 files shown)


bdattoma (dismissed):

lugi0 (approved):
lgtm

ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
```diff
@@ -18,7 +18,7 @@ Installing RHODS Operator ${image_url}
   IF  ${is_codeflare_managed}    Installing CodeFlare Operator
 
 RHODS Operator Should Be installed
-  Verify RHODS Installation
+  Verify RHODS Installation    ${cluster_type}
```
tarukumar, line 21: the is not needed cluster type is global variable

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -50,6 +50,7 @@ Install RHODS
   END
 
 Verify RHODS Installation
+  [Arguments]  ${cluster_type}
```
tarukumar, line 53: not needed as mentioned above

bdattoma:
apart from Tarun's comments, lgtm

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
jstourac: Well, it's used also in a keyword above, so let's keep it to make things consistent. Also, it's always better prefer the passing arguments over having too many global, but I understand you.

jstourac (commented):
LGTM, once you provide a test run, I'm all for it :slightly_smiling_face:

lugi0 (approved):

bdattoma (approved):

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
jgarciao: I removed the argument before reading this comment. I'll keep it without the argument

ods_ci/tasks/Resources/RHODS_OLM/install/install.robot
jgarciao: Removed

jgarciao:
Tested successfully here: rhods-smoke/4210

PR #1078, opened 2023-12-13
Improve Hive deployment to fail fast on fatal error, and fix deprovision logic

manosnoam:
- Fail fast on OCP install Fatal error, instead of waiting for 50 minutes timeout.
This is implemented by watching Hive pods completion status.

- Fix deprovision that was running always (in Teardown) if provision failed -  
even when provision failed since the cluster name already existed (legitimate failure),
which caused deprovision of existing cluster without intention.

- 'Wait For Cluster To Be Ready' not only watches for Install to be completed,
but also for Hive cluster deployment to include API endpoint.

### DIFF
--- a/ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
+++ b/ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
@@ -32,7 +32,16 @@ Provision Cluster
     ${clustername_exists} =    Does ClusterName Exists
     ${template} =    Select Provisioner Template    ${provider_type}
     IF    ${clustername_exists}
-    ...    FAIL    Cluster name '${cluster_name}' already exists. Please choose a different name.
+        Log    Cluster name '${cluster_name}' already in use - Checking if it has a valid web-console      console=True
+        ${pool_namespace} =    Get Cluster Pool Namespace    ${pool_name}
+        ${result} =    Run Process 	oc -n ${pool_namespace} get cd ${pool_namespace} -o jsonpath\='{ .status.webConsoleURL }' | grep .    shell=yes
+        IF    ${result.rc} != 0
+            Log    Cluster '${cluster_name}' has previously failed to be provisioned - Cleaning Hive resources    console=True
+            Delete Cluster Configuration
+        ELSE
+            FAIL    Cluster '${cluster_name}' is already in use, please choose a different name.
+        END
+    END
     Log     Configuring cluster ${cluster_name}    console=True
     Create Provider Resources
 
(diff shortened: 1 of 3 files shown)


manosnoam:
This has been tested on PSI, GCP, and AWS. 

Output example of the Provision process exiting right after a fatal error (without waiting for the 50 minutes timeout):

image

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -32,7 +32,16 @@ Provision Cluster
     ${clustername_exists} =    Does ClusterName Exists
     ${template} =    Select Provisioner Template    ${provider_type}
     IF    ${clustername_exists}
-    ...    FAIL    Cluster name '${cluster_name}' already exists. Please choose a different name.
+        Log    Cluster name '${cluster_name}' already exists in Hive pool '${pool_name}' - Checking if it has a valid web-console      console=True
+        ${pool_namespace} =    Get Cluster Pool Namespace    ${pool_name}
+        ${result} =    Run Process 	oc -n ${pool_namespace} get cd ${pool_namespace} -o json | jq -r '.status.webConsoleURL' --exit-status    shell=yes
```
jstourac, line 37: Not sure whether check for webconsole valid url is good enough, but I don't have any better option myself anyway :upside_down_face: so hopefully it is :slightly_smiling_face:

jstourac (approved):
Don't know the code good enough but in general LGTM. Thank you. Hopefully we won't miss any resources to be left over :slightly_smiling_face:

ods_ci/tasks/Tasks/provision_self_managed_cluster.robot
```diff
@@ -28,11 +28,9 @@ Provision Self-Managed Cluster
     Provision Cluster
     Claim Cluster
     Wait For Cluster To Be Ready
-    Verify Cluster Claim
     Save Cluster Credentials
     Login To Cluster
     Set Cluster Storage
-    [Teardown]  Clean Failed Cluster
```
lugi0, line 35: I understand the reasoning behind removing the deprovision step in the case of a legitimate failure (e.g. using the name of a cluster that already exists), but what about the cases in which the provision fails for some other reason? Won't this leave resources pending?

lugi0: I see you have a deprovision step in https://github.com/red-hat-data-services/ods-ci/pull/1078/files#diff-069081f68a5b54944843728104321c2a7f2d596f186734b832ddf7d8265185b7R156, but I'm wondering if that's the only place in which the provisioning can fail

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
manosnoam: Up until now we just checked if the clusterclaim existed, but that is not indication if OCP install completed successfully. We had many times that Install completed, but the API and Web console were not created due to various errors (usually in PSI).

jstourac: Yeah, definitely better than checking the cluster claim.

ods_ci/tasks/Tasks/provision_self_managed_cluster.robot
manosnoam: yes, since we want to automatically trigger deprovision in case Provision has initially failed, but only if it is due to OCP Install error (i.e. Fatal error).
To further justify the deprovision, this patch also watch Hive Cluster Deployment object, where we can see that the deployed cluster has successfully received the endpoints (API and web-console).

lugi0 (approved):

bdattoma (approved):

PR #1079, opened 2023-12-14
Fix 'Operator Subscription Last Condition Should Be'

manosnoam:
'Operator Subscription Last Condition Should Be' can fail if the last condition is not the requested one:
image

For example:
```
oc get subscription servicemeshoperator -n openshift-operators -ojson | jq '.status.conditions'
[
  {
    "lastTransitionTime": "2023-12-13T19:20:40Z",
    "message": "all available catalogsources are healthy",
    "reason": "AllCatalogSourcesHealthy",
    "status": "False",
    "type": "CatalogSourcesUnhealthy"
  },
  {
    "status": "False",
    "type": "BundleUnpacking"
  }
]
```

The current query will return no output:
`oc get subscription servicemeshoperator -n openshift-operators -ojson | jq '.status.conditions | last | select(.type=="CatalogSourcesUnhealthy" and .status=="False" and .reason=="AllCatalogSourcesHealthy")'`

Explanation:
`last` is not required, since k8s condition types do not show twice, but always the latest.

After the fix - the step succeeds:
image

### DIFF
--- a/ods_ci/tests/Resources/Page/Operators/ISVs.resource
+++ b/ods_ci/tests/Resources/Page/Operators/ISVs.resource
@@ -57,7 +57,7 @@ Operator Subscription Last Condition Should Be
     [Arguments]    ${type}    ${status}    ${reason}
     ...    ${subcription_name}    ${namespace}=openshift-operators
     ${rc}    ${out}=    Run And Return Rc And Output
-    ...    oc get subscription ${subcription_name} -n ${namespace} -ojson | jq '.status.conditions | last | select(.type=="${type}" and .status=="${status}" and .reason=="${reason}")'    # robocop: disable
+    ...    oc get subscription ${subcription_name} -n ${namespace} -ojson | jq '.status.conditions[] | select(.type=="${type}" and .status=="${status}" and .reason=="${reason}")'    # robocop: disable
     Should Be Equal As Integers    ${rc}     ${0}
     Should Not Be Empty    ${out}
 


jstourac:
Well, I'm not sure if I understand that correctly. But with the change you provide, you'll simply select the appropriate condition from all the returned conditions. Not only from last (latest) one. So, is it okay for us to check against a condition that was true in the past?

From my understanding, which may be wrong, I don't think this fix is correct. Actually, I think that the code before this fix is correct.

---

This is something that has been seen in #1068 already. When I did my local checks, I was able to easily reproduce the failure you share in the description by putting some 3 second sleep before this check is called. From my observation, at start there is a condition that suites our check, then the last condition is changed to something else and thus the check fails. If I don't understand this in a wrong way (see paragraph above), we're just lucky with checking this condition. I think we should check something else to ensure the operator has been installed successfully.

bdattoma (changes requested):

manosnoam:
@jstourac , @bdattoma  see screenshots in description, before the fix, and after the fix.
Notice to important things about K8S (AFAIK):

1) For K8S objects with conditions, the **same condition type is always the latest**. 
In our case, `CatalogSourcesUnhealthy` cannot appear twice (e.g. one timestamp with status `true`, and another timestamp with status `false`is not possible).
This is why we can simply verify the specific condition status, without worrying if it's the latest or not.

2) Another important point in K8S conditions in the JSON output, that the latest JSON element (on top), is not necessarily the latest condition that was updated. For that you must **compare timestamps**!

jstourac:
@manosnoam okay, objections noted. If that is the way you say, then I suppose, I agree. If I knew this last week, I would have fix for it right away last week instead of commenting these two checks out temporarily for my execution :smile: 

As such, I am okay with the fix now. Question whether this is the proper way how to check the operator is installed successfully and completely is still on the table though, but:

1. it's not point of this PR
2. I may be wrong again :slightly_smiling_face:

jstourac (approved):
LGTM, thank you :slightly_smiling_face:

bdattoma:
> @jstourac , @bdattoma see screenshots in description, before the fix, and after the fix. Notice two important things about K8S (AFAIK):
> 
> 1. For K8S objects with conditions, the **same condition type is always the latest**.
>    In our case, `CatalogSourcesUnhealthy` cannot appear twice (e.g. one timestamp with status `true`, and another timestamp with status `false`is not possible).
>    This is why we can simply verify the specific condition status, without worrying if it's the latest or not.
> 2. Another important point in K8S conditions in the JSON output, that the latest JSON element (on top), is not necessarily the latest condition that was updated. For that you must **compare timestamps**!


Then, because of point 2, we should rename the keyword..up to you. 
Please could you test this PR including the various tests which use this keyword?

bdattoma (approved):

jstourac:
@bdattoma it's used just on that one place (was on two before Tarun's change recently), so I suppose this is checked.

Regarding the renaming - well, it may make sense, truth. In such case, please sync with Tarun because of this one #1080.

bdattoma:
> @bdattoma it's used just on that one place (was on two before Tarun's change recently), so I suppose this is checked.
> 
> Regarding the renaming - well, it may make sense, truth. In such case, please sync with Tarun because of this one #1080.

it's used in:
- ods_ci/tests/Tests/600__ai_apps/680__starburst/682__starburst_enterprise_sm.robot
- ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
- ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot

jstourac:
Okay, right, I was checking just for `Operator Subscription Last Condition Should Be`. WRT `Wait Until Operator Subscription Last Condition Is`, you're right, Berto. Thanks for clearing this :slightly_smiling_face:

manosnoam:
> > @bdattoma it's used just on that one place (was on two before Tarun's change recently), so I suppose this is checked.
> > Regarding the renaming - well, it may make sense, truth. 
> 
> it's used in:
> 
>     * ods_ci/tests/Tests/600__ai_apps/680__starburst/682__starburst_enterprise_sm.robot
> 
>     * ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/422__model_serving_llm.robot
> 
>     * ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot

Besides testing during Install, I also tested with the above test cases (ODS-2247 and ODS-2326):

image

About renaming, it is not really required, since it is the latest like the kw name :) 
As mentioned before, condition types displayed are always with the latest status.

PR #1082, opened 2023-12-15
Remove SMMR clean up from KServe test teardown

bdattoma:
The kserve tests try to delete the test namespace from the ServiceMeshMemberRoll (SMMR) but most of the time it fails, because the model controller does it automatically (it wasn't like this in origin, but then it got fixed).

This PR is removing the SMMR clean up and replace it with a check to ensure that the model controller actually does it automatically.


jstourac (commented):
Some of these Robocop warnings could be easily dismissed. Otherwise LGTM, Berto. Thank you :slightly_smiling_face:  Looking forward to a link to CI.

bdattoma:
> Some of these Robocop warnings could be easily dismissed. Otherwise LGTM, Berto. Thank you 🙂 Looking forward to a link to CI.

robocop alerts fixed (unless new ones appear after the check runs again).

PR validation: `rhods-ci-pr-test/2247/console` running...

jstourac (approved):
Still one easy to remove Robocop warning but approving :smile:

jgarciao (approved):

PR #1086, opened 2023-12-18
RHODS-12973: Consider using our own hosted zones *.rh-ods.com

apodhrad:


apodhrad:
This PR is replaced by https://github.com/red-hat-data-services/ods-ci/pull/1097

PR #1089, opened 2023-12-24
Workaround for old Hive with Openstack

manosnoam:
For Openstack in older Hive version, we need to ignore ClusterClaim that does not show `ClusterRunning` condition as true, while the cluster is actually running, but displaying `Resuming`. For example:
image

This is a workaround to assume Openstack cluster is running, if the ClusterClain condition shows a reason of `Resuming`:
image


kornys (approved):

jstourac:
@manosnoam

Is the `trailing whitespace` robocop warning possible to be removed?

Since you mention the `old Hive` - is this issue present also in the new Hive instance we're about to migrate to?

jgarciao (approved):

PR #1090, opened 2023-12-25
 Add support to validate gpu driver ,demonset etc validator pod verifcation after NFD resource are created

tarukumar:
Noticed that currently, we are only validating the GPU operator pod upon its installation. However, there's a significant risk that users won't be able to access the node because the driver pods aren't deployed. To mitigate this issue, I've implemented code to validate the other essential pods to ensure users don't encounter the aforementioned problem.


tarukumar:
#4269 verified

mwaykole (approved):

aloganat (approved):

PR #1101, opened 2024-01-07
New Test ODS-2223

manosnoam:
Add New Test ODS-2223:
Verify permissions set as part of Project Sharing do not override Dashboard permission.

image

Including:
- New Keyword: Set RHODS Users Group To rhods-users
- New Keyword: Assign Permissions To User in Project
- New Keyword: Verify No Permissions For User
- 'Reload Page If Project Is/Not Listed' Keyword with WHILE and Timeout
- Fix 'Create Group' KW to apply new group only if it doesn't exist
- Fix 'Delete Group' KW to delete the group only if it already exists
- Add Resource ODS.robot to Permissions.resource and set Groups Variables
- Add Users names to Suite Variables

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
@@ -1,14 +1,12 @@
 *** Settings ***
 Documentation    Collection of keywords to interact with Storages
 Resource       ./Projects.resource
-Resource       ../../../../../Resources/ODS.robot
 
 
 *** Variables ***
 ${SAVE_BUTTON}=    xpath://button[@data-id="save-rolebinding-button"]
 ${PERMISSIONS_DROPDOWN}=    xpath://td[@data-label="Permission"]//button[@aria-label="Options menu"]
 ${IS_CLUSTER_ADMIN}=    ${FALSE}
-${CUSTOM_EMPTY_GROUP}=    empty-group
 
 
 *** Keywords ***
@@ -22,6 +20,13 @@ Assign ${permission_type} Permissions To User ${username}
     Select Permission Type    permission_type=${permission_type}
     Save Permission
 
+Assign ${permission_type} Permissions To User ${username} in Project ${project_title}
+    Open Data Science Projects Home Page
+    Wait Until Project Is Listed    project_title=${project_title}
+    Open Data Science Project Details Page    ${project_title}
+    Move To Tab             Permissions
+    Assign ${permission_type} Permissions To User ${username}
+    
 Change ${username} Permissions To ${permission_type}
     [Documentation]    Change the level of permission ${permission_type} for the given user ${username}
     ...    in the currently open DS Project in UI
@@ -104,51 +109,3 @@ Is ${username} In The Permissions Table
     ${present}=    Run Keyword And Return Status
     ...    Page Should Contain Element       xpath=//tr[td[@data-label="Username"]//*[text()="${username}"]]
     RETURN    ${present}
-
-Set RHODS Admins Group Empty Group
-    [Documentation]     Sets the "adminGroups" field in "odh-dashboard-config" ConfigMap
-    ...                 to the given empty group (i.e., with no users)
-    Set Access Groups Settings    admins_group=${CUSTOM_EMPTY_GROUP}
-    ...    users_group=${STANDARD_SYSTEM_GROUP}
-
-Set RHODS Admins Group To system:authenticated    # robocop:disable
-    [Documentation]    Sets the "adminGroups" field in "odh-dashboard-config" ConfigMap
-    ...    to the system:authenticated group
-    Set Access Groups Settings    admins_group=system:authenticated
-    ...    users_group=${STANDARD_SYSTEM_GROUP}
-
-Set RHODS Admins Group To Inexistent Group
-    [Documentation]    Sets the "adminGroups" field in "odh-dashboard-config" ConfigMap
-    ...    to the given inexistent group
-    Set Access Groups Settings    admins_group=${CUSTOM_INEXISTENT_GROUP}
-    ...    users_group=${STANDARD_SYSTEM_GROUP}
-
-Set RHODS Users Group Empty Group
-    [Documentation]    Sets the "allowedGroups" field in "odh-dashboard-config" ConfigMap
-    ...    to the given empty group (i.e., with no users)
-    Set Access Groups Settings    admins_group=${STANDARD_ADMINS_GROUP}
-    ...    users_group=${CUSTOM_EMPTY_GROUP}
-
-Set RHODS Users Group To rhods-users    # robocop:disable
-    [Documentation]    Sets the "allowedGroups" field in "odh-dashboard-config" ConfigMap
-    ...    to the rhods-users group
-    Set Access Groups Settings    admins_group=${STANDARD_ADMINS_GROUP}
-    ...    users_group=${STANDARD_USERS_GROUP}
-
-Set RHODS Users Group To Inexistent Group
-    [Documentation]    Sets the "allowedGroups" field in "odh-dashboard-config" ConfigMap
-    ...    to the given inexistent group
-    Set Access Groups Settings    admins_group=${STANDARD_ADMINS_GROUP}
-    ...    users_group=${CUSTOM_INEXISTENT_GROUP}
-
-Set Default Groups And Check Logs Do Not Change
-    [Documentation]    Teardown for ODS-1408 and ODS-1494. It sets the default configuration of "odh-dashboard-config"
-    ...    ConfigMap and checks if no new lines are generated in the logs after that.
-    [Arguments]    ${delete_group}=${FALSE}
-    ${lengths_dict}=    Get Lengths Of Dashboard Pods Logs
-    Set Access Groups Settings    admins_group=${STANDARD_ADMINS_GROUP}
-    ...    users_group=${STANDARD_SYSTEM_GROUP}
-    Logs Of Dashboard Pods Should Not Contain New Lines  lengths_dict=${lengths_dict}
-    IF  "${delete_group}" == "${TRUE}"
-        Delete Group    group_name=${CUSTOM_EMPTY_GROUP}
-    END
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
@@ -31,7 +31,7 @@ Change ${username} Permissions To ${permission_type}
     Save Permission
 
 Remove ${username} Permissions
-    [Documentation]    Remove the the access to the given user ${username}
+    [Documentation]    Remove the access to the given user ${username}
     ...    from the currently open DS Project in UI
     Permissions.Click Action From Actions Menu    username=${username}
     ...    action=Delete
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
@@ -1,12 +1,14 @@
 *** Settings ***
 Documentation    Collection of keywords to interact with Storages
 Resource       ./Projects.resource
+Resource       ../../../../../Resources/ODS.robot
 
 
 *** Variables ***
 ${SAVE_BUTTON}=    xpath://button[@data-id="save-rolebinding-button"]
 ${PERMISSIONS_DROPDOWN}=    xpath://td[@data-label="Permission"]//button[@aria-label="Options menu"]
 ${IS_CLUSTER_ADMIN}=    ${FALSE}
+${CUSTOM_EMPTY_GROUP}=    empty-group
 
 
 *** Keywords ***
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
@@ -106,37 +106,43 @@ Is ${username} In The Permissions Table
     RETURN    ${present}
 
 Set RHODS Admins Group Empty Group
-    [Documentation]     Sets the "admins_groups" field in "rhods-groups-config" ConfigMap
+    [Documentation]     Sets the "adminGroups" field in "odh-dashboard-config" ConfigMap
     ...                 to the given empty group (i.e., with no users)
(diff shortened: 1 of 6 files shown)


FedeAlonso (approved):
LGTM

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
```diff
@@ -102,3 +104,51 @@ Is ${username} In The Permissions Table
...
+    [Documentation]    Teardown for ODS-1408 and ODS-1494. It sets the default configuration of "odh-dashboard-config"
+    ...    ConfigMap and checks if no new lines are generated in the logs after that.
+    [Arguments]    ${delete_group}=${FALSE}
+    ${lengths_dict}=    Get Lengths Of Dashboard Pods Logs
+    Set Access Groups Settings    admins_group=${STANDARD_ADMINS_GROUP}
+    ...    users_group=${STANDARD_SYSTEM_GROUP}
+    Logs Of Dashboard Pods Should Not Contain New Lines  lengths_dict=${lengths_dict}
+    IF  "${delete_group}" == "${TRUE}"
+        Delete Group    group_name=${CUSTOM_EMPTY_GROUP}
+    END
```
bdattoma, line 154: I think these keywords should go in `ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboardSettings.resource`. 

`ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource` is supposed to contain keyword for handling the Permissions of the DS projects, which is different from Dashboard permissions

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
```diff
@@ -76,13 +82,33 @@ Verify User Can Assign Access Permissions To User Groups
...
+
 
 *** Keywords ***
 Project Permissions Mgmt Suite Setup    # robocop: disable
     [Documentation]    Suite setup steps for testing DS Projects.
     ...                It creates some test variables and runs RHOSi setup
     Set Library Search Order    SeleniumLibrary
     RHOSi Setup
+    Set Standard RHODS Groups Variables
+    Set Default Access Groups Settings
```
bdattoma, line 111: this one should in the teardown

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
```diff
@@ -76,13 +82,33 @@ Verify User Can Assign Access Permissions To User Groups
...
+    ...    ocp_user_auth_type=${OCP_ADMIN_USER.AUTH_TYPE}
+    Assign Admin Permissions To User ${USER_B} in Project ${PRJ_USER_B_TITLE}
+    Assign Edit Permissions To User ${USER_C} in Project ${PRJ_USER_C_TITLE}
+    Remove User From Group    username=${USER_B}    group_name=rhods-users
+    Remove User From Group    username=${USER_B}    group_name=rhods-admins
+    Remove User From Group    username=${USER_C}    group_name=rhods-users
+    Remove User From Group    username=${USER_C}    group_name=rhods-admins
+    Verify No Permissions For User ${USER_B}
+    Verify No Permissions For User ${USER_C}
+    [Teardown]              Run Keywords            Set RHODS Admins Group To system:authenticated
```
bdattoma, line 100: system:authenticated is not the default admin group. Please use `Set Default Access Groups Settings` in the teardown

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
```diff
@@ -263,3 +268,18 @@ ${username} Should Not Have Access To ${project_title}
...
     ...    subject_name=${username}
+
+Assign ${permission_type} Permissions To User ${username} in Project ${project_title}
+    Open Data Science Projects Home Page
+    Wait Until Project Is Listed    project_title=${project_title}
+    Open Data Science Project Details Page    ${project_title}
+    Move To Tab             Permissions
+    Assign ${permission_type} Permissions To User ${username}
+
+Verify No Permissions For User ${username}
```
bdattoma, line 279: since it is checking the user permission to dashboard, what do you think about:
1. renaming the kw to `User ${username} Should Not Be Allowed To Dashboard`
2. moving the keyword to `ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot`

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
```diff
@@ -263,3 +268,18 @@ ${username} Should Not Have Access To ${project_title}
     Project Should Not Be Listed    project_title=${project_title}
     RoleBinding Should Not Exist    project_title=${project_title}
     ...    subject_name=${username}
+
+Assign ${permission_type} Permissions To User ${username} in Project ${project_title}
+    Open Data Science Projects Home Page
+    Wait Until Project Is Listed    project_title=${project_title}
```
bdattoma, line 274: it's missing the `[Argument]   ${project_title}`

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
```diff
@@ -263,3 +268,18 @@ ${username} Should Not Have Access To ${project_title}
     Project Should Not Be Listed    project_title=${project_title}
     RoleBinding Should Not Exist    project_title=${project_title}
     ...    subject_name=${username}
+
+Assign ${permission_type} Permissions To User ${username} in Project ${project_title}
```
bdattoma, line 272: I think this one could be moved to `ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource` as well

bdattoma (changes requested):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
manosnoam: I moved them from `ods_ci/tests/Tests/400__ods_dashboard/400__ods_dashboard.robot` :
https://github.com/red-hat-data-services/ods-ci/pull/1101/files#diff-be67337b3f53b14968b105e32d5a8a2769771595618bb421553a393adb1373f6

Into `ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource`
(And just added few more similar keywords), because they all uses `Set Access Groups Settings`, and it seems more relevant.

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
manosnoam: I had to add it to the Setup too, since we need to make sure the required variables and Users are set correctly before starting the tests. It is safer this way, why do you there an advantage not calling it ?

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
manosnoam: It doesn't miss [Argument] section, as all arguments already embedded in the keyword name.
See also:
https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#embedding-arguments-into-keyword-name

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
manosnoam: I will change the name, but it's not a good idea to move this keyword to another resource since:
1. It is only used in this test suite
2. It is using `Switch To User` kw, which is unique to this suite, as it is actually switching to a new browser that was already defined for this user (user is actually browser ID, not user of RHODS)

bdattoma: okay, thx

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
bdattoma: ops, sorry I misread it

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
bdattoma: yes but to me it makes sense to restore the default access group in the teardown as well.

manosnoam: You right - I run 'Set Default Access Groups Settings' in the new TEST TEARDOWN.
This is to prevent other tests that runs afterwards - but before SUITE TEARDOWN - to fail due to missing default permissions. I will add it to SUITE TEARDOWN as well.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Permissions.resource
manosnoam: As you suggested, I moved these Keywords into `ODHDashboardSettings.resource`

manosnoam:
Thanks @bdattoma , @FedeAlonso  I resolved the requested issues, and retested:
image

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
```diff
@@ -76,13 +82,31 @@ Verify User Can Assign Access Permissions To User Groups
...
+    Assign Admin Permissions To User ${USER_B} in Project ${PRJ_USER_B_TITLE}
+    Assign Edit Permissions To User ${USER_C} in Project ${PRJ_USER_C_TITLE}
+    Remove User From Group    username=${USER_B}    group_name=rhods-users
+    Remove User From Group    username=${USER_B}    group_name=rhods-admins
+    Remove User From Group    username=${USER_C}    group_name=rhods-users
+    Remove User From Group    username=${USER_C}    group_name=rhods-admins
+    User ${USER_B} Should Not Be Allowed To Dashboard
+    User ${USER_C} Should Not Be Allowed To Dashboard
+    [Teardown]              Run Keywords            Set Default Access Groups Settings
+    ...                     AND                     Set User Groups For Testing
```
bdattoma, line 99: wait, why running this in the test teardown? I think it's not needed. You don't see issues because there is the Suite Teardown which restores the default settings.

I believe you can remove the `[Teardown]` here and let the suite teardown to restore default user settings

bdattoma:
LGTM, apart for my last comment, I didn't notice earlier sorry. @manosnoam

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_permissions_mgmt.robot
manosnoam: I think it is safer to run `Set User Groups For Testing` in test teardown, in case the test is **not the last one** in suite, so if another test is called afterwards, then the expected Users are already set properly.

bdattoma: hmm I think you could actually avoid lines 92 and 94 because those are already executed in the Suite Setup. If you want to remove the users from the rhods-admin group as well, you can also add it to the Suite Setup (and change the Suite Teardown accordingly)

Alternatively, instead of using `Set RHODS Users Group To rhods-users` you can do `Set RHODS Users Group To ${USER_GROUP_1}` and `Set RHODS Users Group To ${USER_GROUP_1}`. This should remove lines from 91 to 95 + 98-99

bdattoma: I understand now why you did it as you did, it works of course. However it think it could be cleaner as I wrote in the above comment.

If you disagree, I will approve anyways

manosnoam: @bdattoma the Test case steps to automate say:

> assign "Admin" permissions to a user B who does not belong to any RHODS user group
> assign "Edit" permissions to a user C who does not belong to any RHODS user group

The existing KW  `Set RHODS Users Group To` doesn't do that exactly, as it does not specifically make the user not part of a group. It modifies `oc patch OdhDashboardConfig odh-dashboard-config`, which is a slightly different scenario. 
This is why I use KW `Remove User From Group` rhods-users and rhods-admin, which does:
`oc adm groups remove-users ${group_name} ${username}`

I think this is good enough for the scenario, we can later extend this scenario if needed.

bdattoma: I know, but the suite setup already removes users from group. Besides, I'm suggesting to use `${USER_GROUP_1}`  for both admin and basic user group because the USER_B and USER_C you're using are not part of that group.

manosnoam:
I cannot count on the suite setup, if previous test could alter the users,
that's why I make sure the users are ready for the test.

I don't see any advantage in changing this logic if it works good, and if
we change the logic, it will require further tests and rebase. I think only
if you see an actual problem in this logic, we should fix it before merging.

bdattoma (approved):

PR #1102, opened 2024-01-09
[RHOAIENG-223] Test VSCode - add checks for VS Code image

jstourac:
This introduces tests for the recently added VS Code IDE image. For more information see [1,2].

[1] https://issues.redhat.com/browse/RHOAIENG-223
[2] https://issues.redhat.com/browse/RHOAIENG-225

---

CI: rhods-ci-pr-test/2349

### DIFF
--- a/ods_ci/tests/Resources/Common.robot
+++ b/ods_ci/tests/Resources/Common.robot
@@ -46,6 +46,16 @@ End Web Test
     END
     Close Browser
 
+End Non JupyterLab Web Test
+    [Documentation]  Stops running workbench that was started by logged-in user via the
+    ...              JupyterHub launcher space.
+    Go To  ${ODH_DASHBOARD_URL}
+    Wait for RHODS Dashboard to Load
+    Launch Jupyter From RHODS Dashboard Link
+    Handle Control Panel
+    Capture Page Screenshot
+    Close Browser
+
 Load Json File
     [Arguments]   ${file_path}
     ${j_file}=    Get File    ${file_path}
--- a/ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -269,6 +269,42 @@ Has Spawn Failed
     ${spawn_status} =  Run Keyword And Return Status  Page Should Contain  Spawn failed
     RETURN  ${spawn_status}
 
+Notebook Expected Ide
+    [Documentation]  Returns type of expected IDE for the given image name.
+    ...              At the moment, there are two IDE types available in the notebook images:
+    ...              JupyterLab and VSCode.
+    ...              Returns type of expected IDE for the given image name as a string.
+    [Arguments]  ${image_name}
+    IF  "${image_name}"=="${EMPTY}"
+        Log  level=ERROR  message=No image name has been provided!
+        RETURN  ${EMPTY}
+    END
+    IF  "${image_name}"=="code-server"
+        RETURN  VSCode
+    ELSE
+        RETURN  JupyterLab
+    END
+
+Wait Notebook To Be Loaded
+    [Documentation]  Waits for the notebook IDE environment to be loaded completely and performs
+    ...              a simple check with the Menu action in the loaded IDE.
+    [Arguments]  ${image_name}  ${version}
+    ${ide}=  Notebook Expected Ide  ${image_name}
+
+    IF  "${ide}"=="VSCode"
+        Wait Until Page Contains Element  xpath://div[@class="menubar-menu-button"]  timeout=60s
+        Wait Until Page Contains Element  xpath://div[@class="monaco-dialog-box"]  timeout=60s
+        Wait Until Page Contains  Do you trust the authors of the files in this folder?
+    ELSE IF  "${ide}"=="JupyterLab"
+        Wait Until Page Contains Element  xpath://div[@id="jp-top-panel"]  timeout=60s
+        Sleep    2s    reason=Wait for a possible popup
+        Maybe Close Popup
+        Open New Notebook In Jupyterlab Menu
+        Spawned Image Check    ${image}    ${version}
+    ELSE
+        Fail    msg=Unknown IDE typ has been resolved: '${ide}'. Please check and fix or implement.
+    END
+
 Spawn Notebook With Arguments  # robocop: disable
     [Documentation]  Selects required settings and spawns a notebook pod. If it fails due to timeout or other issue
     ...              It will try again ${retries} times (Default: 1) after ${retries_delay} delay (Default: 0 seconds).
@@ -318,11 +354,7 @@ Spawn Notebook With Arguments  # robocop: disable
             Run Keyword And Warn On Failure   Login To Openshift  ${username}  ${password}  ${auth_type}
             ${authorization_required} =  Is Service Account Authorization Required
             IF  ${authorization_required}  Authorize jupyterhub service account
-            Wait Until Page Contains Element  xpath://div[@id="jp-top-panel"]  timeout=60s
-            Sleep    2s    reason=Wait for a possible popup
-            Maybe Close Popup
-            Open New Notebook In Jupyterlab Menu
-            Spawned Image Check    ${image}    ${version}
+            Wait Notebook To Be Loaded  ${image}    ${version}
             ${spawn_fail} =  Has Spawn Failed
             Exit For Loop If  ${spawn_fail} == False
             Reload Page
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
@@ -268,13 +268,14 @@ Launch And Access Workbench From Projects Home Page
     [Documentation]    Launches a running workbench from DS Projects home page.
     [Arguments]     ${workbench_title}    ${project_title}    ${username}=${TEST_USER_3.USERNAME}
     ...             ${password}=${TEST_USER_3.PASSWORD}    ${auth_type}=${TEST_USER_3.AUTH_TYPE}
+    ...             ${expected_ide}=JupyterLab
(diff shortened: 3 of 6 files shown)


jiridanek (approved):

jiridanek:
+1 I think it makes sense merging just the spawning test, after a little cleanup.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
```diff
@@ -258,13 +258,14 @@ Launch And Access Workbench From Projects Home Page
     [Documentation]    Launches a running workbench from DS Projects home page.
     [Arguments]     ${workbench_title}    ${project_title}    ${username}=${TEST_USER_3.USERNAME}
     ...             ${password}=${TEST_USER_3.PASSWORD}    ${auth_type}=${TEST_USER_3.AUTH_TYPE}
+    ...             ${expected_ide}=JupyterHub
```
bdattoma, line 261: pls change it to Jupyter or JupyterLab

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
```diff
@@ -321,15 +321,23 @@ Start Workbench
...
+    IF  "${expected_ide}"=="VSCode"
+        Wait Until Page Contains Element  xpath://div[@class="menubar-menu-button"]  timeout=60s
+        Wait Until Page Contains Element  xpath://div[@class="monaco-dialog-box"]  timeout=60s
+        Wait Until Page Contains  Do you trust the authors of the files in this folder?
+    ELSE IF  "${expected_ide}"=="JupyterHub"
+        Wait Until Page Contains Element  xpath://div[@id="jp-top-panel"]  timeout=60s
+        Maybe Close Popup
+    ELSE
+        Fail    msg=Unknown IDE typy given: '${expected_ide}'. Please check and fix or implement.
+    END
```
bdattoma, line 340: why not re-using `Wait Notebook To Be Loaded`?

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_notebook_images.robot
```diff
@@ -41,6 +41,28 @@ Verify User Can Create A Workbench Using Intel AiKit Image
...
+    ...        pv_description=${PV_DESCRIPTION}  pv_size=1
+    ...        press_cancel=${FALSE}    envs=${NONE}
+    Wait Until Workbench Is Started     workbench_title=${WORKBENCH_TITLE}
+    Open Data Science Projects Home Page
+    Wait Until Project Is Listed    project_title=${PRJ_TITLE}
+    Launch And Access Workbench From Projects Home Page    workbench_title=${WORKBENCH_TITLE}
+    ...    project_title=${PRJ_TITLE}    username=${TEST_USER_3.USERNAME}
+    ...    password=${TEST_USER_3.PASSWORD}    auth_type=${TEST_USER_3.AUTH_TYPE}
+    ...    expected_ide=VSCode
+
```
bdattoma, line 65: what about ensuring VS Code can be actually used? like running some code inside

ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -268,6 +268,42 @@ Has Spawn Failed
     ${spawn_status} =  Run Keyword And Return Status  Page Should Contain  Spawn failed
     RETURN  ${spawn_status}
 
+Notebook Expected Ide
+    [Documentation]  Returns type of expected IDE for the given image name.
+    ...              At the moment, there are two IDE types available in the notebook images:
+    ...              JupyterHub and VSCode.
```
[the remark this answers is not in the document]
jstourac: Of course, updated! :slightly_smiling_face:

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
jstourac: Of course, updated! :slightly_smiling_face:

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
jstourac: Yeah, I was thinking about it. Unfortunately, I would have to know the ID of the image here. Since there is a different approach how the workbench is spawned via JupyterHub vs Data Science project, I don't know the image ID that was spawned. If I decide to propagate image ID instead of the expected IDE variable, it means more changes on more places. With this approach I just put `JupyterLab` as a default, since it was only one other value before.

Another thing is that in JupyterHub launcher, there are different IDs for images used than what is used to determine in DS projects - but (even though not nice) that would be easy to fix with amending the `Notebook Expected Ide` so this argument doesn't apply fully.

---

Let me think about it more, maybe I'll amend.

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_notebook_images.robot
jstourac: Yeah, I'll need to understand how to do it properly first. The UI written for VS Code is more JS event based compared to our RHOAI dashboard or JupyterLab. I was thinking about having at least this basic one in place and then try to do more.

Ideal situation would be to be able to actually perform such check in the upstream repository instead here, but we're not there yet...

bdattoma: >Ideal situation would be to be able to actually perform such check in the upstream repository instead here, but we're not there yet...


wouldn't it still be part of an E2E test to validate it works fine?

jstourac: It depends on the actual check - if we would like to check an actual integration with the product that wasn't done in upstream repo, then yes. But if we would like to just check e.g. that we can create a file with some simple code and are able to compile it, then we don't need to do it here IMHO. We can just check the image tag that it matches what we expected/tested upstream... but we're far from this anyway...

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects_notebook_images.robot
```diff
@@ -41,6 +41,28 @@ Verify User Can Create A Workbench Using Intel AiKit Image
...
+Verify User Can Create A Workbench Using Code Server Image
+    [Documentation]    Verifies that a workbench can be created using Code Server image
+    [Tags]    Sanity    Tier1
+    Set Test Variable   ${IMG_NAME}    Code Server
+    Set Test Variable   ${WORKBENCH_TITLE}    vscode
+    Set Test Variable   ${PV_NAME}    vscodepv
+    Set Test Variable   ${PV_DESCRIPTION}    PV for vscode
+    Go To RHODS Dashboard
+    Open Data Science Project Details Page       project_title=${PRJ_TITLE}
+    Create Workbench    workbench_title=${WORKBENCH_TITLE}  workbench_description=workbench for testing
```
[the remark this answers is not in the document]
jstourac: hm, the workbench name is different for each test. Do I miss something? :thinking:

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
jstourac: I think I would keep this as is for now not mixing the approaches for JupyterLab and DS projects at the moment. Let me know in case you're strongly against it. Thanks!

ods_ci/tests/Tests/500__jupyterhub/minimal-vscode-test.robot
```diff
@@ -0,0 +1,39 @@
...
+Verify VSCode Image Suite Setup
+    [Documentation]    Suite Setup, spawns vscode image
+    Begin Web Test
+    Launch JupyterHub Spawner From Dashboard
+    Spawn Notebook With Arguments  image=${NOTEBOOK_IMAGE}  size=Small
+
+Close Previous Server
+    [Documentation]  Closes previous server and goes back to Spawner
+    Clean Up Server
+    Stop JupyterLab Notebook Server
```
bdattoma, line 37: does it work for VS Code as well?

bdattoma (approved):

ods_ci/tests/Tests/500__jupyterhub/minimal-vscode-test.robot
jstourac: yeah, no... it isn't being called at all, actually :slightly_frowning_face: let me check

jstourac: Updated - I removed this part completely and changed the teardown method at the beginning of the file that stops the created workbench at the end of the test execution.

ods_ci/tests/Resources/Common.robot
```diff
@@ -46,6 +46,16 @@ End Web Test
     END
     Close Browser
 
+End Non JupyterLab Web Test
```
bdattoma, line 49: maybe this could fit more at `ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot `

jstourac: I can move it, although I copied and amended method above named `End Web Test`, that's why I kept it here. Let me know if you still prefer me to move it.

bdattoma (approved):

jiridanek (approved):

PR #1104, opened 2024-01-10
ODS-1975 - Verify user can cancel workbench start from event log [AUTOMATION]

jiridanek:
CI: rhods-ci-pr-test/2301
Tracking issue: RHOAIENG-1167

### DIFF
--- a/ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
+++ b/ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
@@ -221,7 +221,7 @@ Verify User Can Create A S3 Data Connection And Connect It To Workbenches
     ...    status=${WORKBENCH_STATUS_STARTING}
     Run Keyword And Continue On Failure    Wait Until Workbench Is Started     workbench_title=${WORKBENCH_TITLE}
     Workbench Status Should Be      workbench_title=${WORKBENCH_2_TITLE}      status=${WORKBENCH_STATUS_STOPPED}
-    [Teardown]    Run Keywords    
+    [Teardown]    Run Keywords
     ...    Clean Project From Workbench Resources    workbench_title=${WORKBENCH_2_TITLE}    project_title=${PRJ_TITLE}
     ...    AND
     ...    Clean Project From Workbench Resources    workbench_title=${WORKBENCH_TITLE}    project_title=${PRJ_TITLE}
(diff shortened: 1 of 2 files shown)


jstourac (approved):

jstourac:
Thank you, LGTM. Could you please also provide a link to the relevant tracking issue? Thanks!

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
```diff
@@ -400,6 +400,32 @@ Verify Event Log Is Accessible While Starting A Workbench
...
+    ...        press_cancel=${FALSE}    envs=${NONE}
+
+    Workbench Status Should Be    workbench_title=${WORKBENCH_TITLE}  status=${WORKBENCH_STATUS_STARTING}
+
+    Open Notebook Event Log    workbench_title=${WORKBENCH_TITLE}
+    Page Should Contain Event Log
+    Cancel Workbench Startup From Event Log
+    Wait Until Workbench Is Stopped    workbench_title=${WORKBENCH_TITLE}
+
+    Workbench Status Should Be    workbench_title=${WORKBENCH_TITLE}  status=${WORKBENCH_STATUS_STOPPED}
```
bdattoma, line 421: it's redundant with the previous line. `Wait Until Workbench Is Stopped` internally uses `Workbench Status Should Be`

jiridanek: `Wait Until Workbench Is Stopped` does not have `Should` in it and it does not feel like i am making an assertion. And if I wanted to run `Workbench Status Should Be` repeatedly, then I'd have to set a check frequency and timeout which feels like extraneous detail when reading the test.

I'll use `Wait Until Workbench Is Stopped` only and remove `Workbench Status Should Be`.

bdattoma: > does not have Should in it and it does not feel like i am making an assertion. 


yep, makes sense but it is technically doing the same check, it would be a line of code we could spare. But if you think the benifits of keeping it are important enough, feel free to do it :) 


> [...] which feels like extraneous detail when reading the test.


Maybe this logic can be applied to the following lines as well:

```
Wait Until Keyword Succeeds  5 min  5s
    ...    Run Keyword And Expect Error  EQUALS:ResourceOperationFailed: Get failed\nReason: Not Found
    ...    Get Workbench Pod    project_title=${PRJ_TITLE}  workbench_title=${WORKBENCH_TITLE}
```

jiridanek: > Maybe this logic can be applied to the following lines as well:

I guess so. I did not have any bright ideas about what to do instead. And I still don't.

bdattoma: You could create a separate keyword

PR #1106, opened 2024-01-10
Fix Dashboard Cards xpath to fix failure in 2.6 for Patternfly

nehachopra27:


jstourac:
Hi @nehachopra27, thank you for this. May I ask why only some subset of the `pf-*` class attributes were updated? At first sight I though this was a bulk update of all of these. But maybe you did a more thorough check instead?

ods_ci/tests/Resources/Page/HybridCloudConsole/HCCLogin.robot
```diff
@@ -9,8 +9,8 @@ Is SSO Login Page Visible
   RETURN  ${login_prompt_visible}
 
 Wait For HCC Splash Page
-   Wait Until Page Contains Element    xpath://span[contains(@class, 'pf-c-spinner')]   timeout=15
-   Wait Until Page Does Not Contain Element    xpath://span[contains(@class, 'pf-c-spinner')]   timeout=20
+   Wait Until Page Contains Element    xpath://span[contains(@class, 'pf-v5-c-spinner')]   timeout=15
+   Wait Until Page Does Not Contain Element    xpath://span[contains(@class, 'pf-v5-c-spinner')]   timeout=20
```
bdattoma, line 13: hm probably we could use `-c-spinner` only. My concern is that `-v5-` is supposed to change over versions from now on

nehachopra27: This one was a quick fix, shouldn't we invest more time on it when we have data-testid field ready and replace xpaths then to be not dependant on it. Because this -c-spinner approach will only work on 'contains' xpaths not all.

bdattoma: >when we have data-testid field ready


is it ready? if yes, then we should replace as you're suggesting, sooner than later. If we leave "v5", I can see these tests broken some time in the future.

Ok with this PR if it's intended to be temporary

jgarciao (approved):

jiridanek (approved):
CI: rhoai/job/2.6/job/selfmanaged/job/cli/job/psi/job/rhods-smoke/5

This will get us to a state that the tests don't all pass with patternfly4 (2.5) nor with patternfly5 (2.6). But, since we do need to fix all tests for 2.6, it makes sense to merge this because it fixes many tests (and keep working at making the rest green as well).

PR #1118, opened 2024-01-12
Pin Python version to 3.11

jiridanek:
### Don't we pin Python already?

We don't, or not strictly enough. Currently we have

https://github.com/red-hat-data-services/ods-ci/blob/afa1de64ba5e632e794d70eebb4197956270e419/pyproject.toml#L9-L10

which means "use any Python `>=3.8.1, <4.0.0`".

### Why pin Python version?

Because we _can_ put everybody on the same version.
This way,
we make it less likely different people will hit different infra issues.

### Why use Python 3.11?

It is well established,
binary wheels exist for it in PyPI,
it is available on UBI 8 and 9, and on Fedora starting with Fedora 37.

### Why say `python = ">=3.11.5, <3.12.0"`?

The syntax uses basic math instead of tildes and carets,
which not everybody might've learned what they mean.
That makes the requirement easier to read and understand.

### Will this break CI infra?

~~Most likely, which is why this PR exists, to try things out.~~

All things in order, everything is very good, 🌈 🦄 🌇

### DIFF
--- a/ods_ci/run_robot_test.sh
+++ b/ods_ci/run_robot_test.sh
@@ -319,6 +319,7 @@ if command -v yq &> /dev/null
 fi
 
 if [[ ${SKIP_INSTALL} -eq 0 ]]; then
+  # look for pre-created poetry .venv
   virtenv="${HOME}/.local/ods-ci/.venv"
   if [[ -d "${virtenv}" ]]; then
     echo "Using a pre-created virtual environment in '${virtenv}' for poetry to save time."
@@ -326,6 +327,23 @@ if [[ ${SKIP_INSTALL} -eq 0 ]]; then
     ln --symbolic "${virtenv}" "${basepath}/../.venv"
   else
     echo "Pre-created virtual environment has not been found in '${virtenv}'. All dependencies will be installed from scratch."
+  fi
+  # ensure python 3.11
+  python=$(poetry env info --executable)
+  if [[ -n "${python}" ]] && ${python} -c 'import sys; sys.exit(0 if sys.version_info[0:2] == (3, 11) else 1)'; then
+    echo "Python '${python}' will be used"
+  else
+    echo "Python '${python}' is not of the correct version"
+    python311=$(which python3.11)
+    if [[ -n "${python311}" ]]; then
+      echo "Configuring poetry to use Python ${python311}"
+      poetry env use "${python311}"
+    else
+      echo "[ERROR] Python 3.11 was not found!"
+      echo "Install Python 3.11 on your machine. On Fedora, do 'sudo dnf install -y python3.11-devel'"
+      echo "then run 'poetry env use /path/to/python3.11' and then try running robot again"
+      exit 1
+    fi
   fi
 
   poetry --no-interaction install --sync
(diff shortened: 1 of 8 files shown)


jiridanek:
And, it fails CI as expected, rhods-ci-pr-test/2308

```
21:19:50  The currently activated Python version 3.9.17 is not supported by the project (>=3.11.5, <3.12.0).
21:19:50  Trying to find and use a compatible version. 
21:19:50  
21:19:50  Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
21:19:51  ods_ci/run_robot_test.sh: line 320: /bin/activate: No such file or directory
```

So, I'll have to update Python in CI. Then, I'll announce that contributors should do `dnf install python3.11; poetry env use /usr/bin/python3.11`, and after some time it should be safe to merge this change.

jiridanek:
Here is a CI image update that needs to be merged first, red-hat-data-services/ods-ci/pull/1118

jiridanek:
Also need Python 3.11 in the Dockerfile here, https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/build/Dockerfile

ods_ci/run_robot_test.sh
```diff
@@ -314,6 +314,7 @@ if command -v yq &> /dev/null
 fi
 
 if [[ ${SKIP_INSTALL} -eq 0 ]]; then
+  poetry env use /usr/bin/python3.11
```
jiridanek, line 317: this is not good on macos

jstourac: we should stop to support these :smile:

jiridanek:
CI: rhods-ci-pr-test/2323 (smokes do run)

jgarciao:
I've added the "do not merge" label until we have an agreement on slack about merging this

jiridanek:
This should be all the needed changes. Now onto rechecking it in CI.

ods_ci/run_robot_test.sh
jiridanek: we can't, redhatters like their macs

jiridanek:
There are two todo's that I'll do when this is merged
* [ ] go to gitlab.cee and in ods/jenkins change the pr job so that it does not archive .venv directory
* [ ] go to gitlab.cee and in ods/rhods-qe-tools make python3.11 be the default python for poetry to save CI time

ods_ci/run_robot_test.sh
jiridanek: done mac-friendly solution

ods_ci/utils/scripts/Sender/EmailSender.py
```diff
@@ -24,16 +24,16 @@ def __init__(self):
         self._message = MIMEMultipart()
 
     def prepare_payload(
-        self, text: str = "", attachments: Optional[List[Any]] = None
+        self, text: str = "", attachments: list[Any] | None = None
```
bdattoma, line 27: why?

jiridanek: linter wants this when you have a py3.11 project

at some point they made list[...] be an alias for typing.List[...], so that's a cleanup it wants

not sure about why it dislikes Optional, that should be fine imo, and | None is not a clear win in my eyes

bdattoma (approved):

jstourac (approved):

jiridanek:
job/rhods/job/rhods-ci-pr-test/2487, still looks good

PR #1119, opened 2024-01-13
Improve the code quality ruff action

jiridanek:
* clone the bare minimum
* cache installed poetry
* cache poetry virtualenv
* run poetry check --lock

### DIFF
--- a/.github/workflows/code_quality.yaml
+++ b/.github/workflows/code_quality.yaml
@@ -63,8 +63,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
 
       - name: Get changed files
         id: changed-files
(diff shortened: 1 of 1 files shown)


jstourac (approved):

Frawless (approved):

.github/workflows/code_quality.yaml
```diff
@@ -84,15 +82,33 @@ jobs:
...
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
+      - name: Cache poetry in ~/.local
+        uses: actions/cache@v3
+        id: cached-home-local
+        with:
+          path: ~/.local
+          key: ${{ runner.os }}-local-${{ env.poetry_version }}
```
bdattoma, line 97: what is the benefit of this step?

.github/workflows/code_quality.yaml
```diff
@@ -84,15 +82,33 @@ jobs:
...
+        with:
+          cache: 'poetry'
+
+      - run: poetry check --lock
+
       - name: Install ruff
-        run: |
-          pip install poetry
-          poetry install --only dev
+        run: poetry install --only=dev --sync
```
bdattoma, line 112: why installing once again?

.github/workflows/code_quality.yaml
jiridanek: It allows skipping pip install poetry (the next step) if restoring cache succeeds. Restoring poetry from cache this way takes a few seconds less than installing with pip every time.

So, it saves 3 to 5 seconds off of the job runtime.

.github/workflows/code_quality.yaml
jiridanek: The previous install only got us the poetry executable itself. Now I use poetry to install the ods-ci dependencies

bdattoma: oh right, stupid me, sorry.

PR #1135, opened 2024-01-19
Add support for self-signed certificates

apodhrad:
First, I have updated only one test (ODS-1932) which should not cause any issue.
After the review I will update other tests. We need to add a namespace as we update the underlying secrets.

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
@@ -23,15 +23,40 @@ ${DC_WORKBENCH_SELECTOR_XP}=         xpath=//div[contains(@class,"modal")]//div[
 *** Keywords ***
 Create S3 Data Connection
     [Documentation]    Creates a S3 Data Connection from DS Project details page
+    ...    If 'aws_s3_ca_bundle' or 'S3.AWS_CA_BUNDLE' is defined then the underlying secret is patched
     [Arguments]    ${project_title}    ${dc_name}    ${aws_access_key}    ${aws_secret_access}
     ...            ${aws_s3_endpoint}=${S3.AWS_DEFAULT_ENDPOINT}    ${aws_region}=${S3.AWS_DEFAULT_REGION}
     ...            ${connected_workbench}=${NONE}    ${press_cancel}=${FALSE}    ${aws_bucket_name}=${NONE}
+    ...            ${aws_s3_ca_bundle}=${NONE}
     Click Button    ${DC_ADD_BTN_1_XP}
     Fill Data Connection Form    ${project_title}    ${dc_name}    ${aws_access_key}    ${aws_secret_access}
     ...            ${S3_DC_ADD_BTN}    aws_s3_endpoint=${aws_s3_endpoint}    aws_region=${aws_region}
     ...            connected_workbench=${connected_workbench}    press_cancel=${press_cancel}
     ...            aws_bucket_name=${aws_bucket_name}
     Wait Until Project Is Open    project_title=${project_title}
+    # Check if there is any CA Bundle
+    # I did not find a better way to check if multi-line string is empty or "NONE"
+    ${length}=    Get Length    "${aws_s3_ca_bundle}"
+    IF    ${length} <= 6
+        ${aws_s3_ca_bundle}=    Get Variable Value    ${S3.AWS_CA_BUNDLE}    ${NONE}
+        ${length}=    Get Length    "${aws_s3_ca_bundle}"
+    END
+    IF    ${length} > 6
+        Patch S3 Data Connection With CA Bundle
+        ...    project_title=${project_title}    dc_name=${dc_name}    aws_s3_ca_bundle=${aws_s3_ca_bundle}
+    END
+
+Patch S3 Data Connection With CA Bundle
+    [Documentation]    Patch the underlying secret of S3 Data Connection With CA Bundle
+    [Arguments]    ${project_title}    ${dc_name}    ${aws_s3_ca_bundle}
+    ${dc_namespace}=    Get Openshift Namespace From Data Science Project    ${project_title}
+    ${rc}    ${cert}=    Run And Return Rc And Output
+    ...    echo -n "${aws_s3_ca_bundle}" | base64 -w0    # robocop: disable
+    Should Be Equal As Strings    ${rc}    0
+    ${rc}    ${out}=    Run And Return Rc And Output
+    ...    oc patch secret aws-connection-${dc_name} -n ${dc_namespace} -p '[{"op": "replace", "path": "/data/AWS_CA_BUNDLE", "value" : "${cert}"}]' --type json    # robocop: disable
+    Should Be Equal As Strings    ${rc}    0
+    Log To Console    "${out}"
 
 Edit S3 Data Connection
     [Documentation]    Edits a S3 Data Connection from DS Project details page
--- a/ods_ci/utils/scripts/testconfig/generateTestConfigFile.py
+++ b/ods_ci/utils/scripts/testconfig/generateTestConfigFile.py
@@ -176,6 +176,8 @@ def generate_test_config_file(
     data["S3"]["AWS_SECRET_ACCESS_KEY"] = config_data["S3"]["AWS_SECRET_ACCESS_KEY"]
     data["S3"]["AWS_DEFAULT_ENDPOINT"] = config_data["S3"]["AWS_DEFAULT_ENDPOINT"]
     data["S3"]["AWS_DEFAULT_REGION"] = config_data["S3"]["AWS_DEFAULT_REGION"]
+    if config_data["S3"].get("AWS_CA_BUNDLE"):
+        data["S3"]["AWS_CA_BUNDLE"] = config_data["S3"]["AWS_CA_BUNDLE"]
     data["S3"]["BUCKET_1"]["NAME"] = config_data["S3"]["BUCKET_1"]["NAME"]
     data["S3"]["BUCKET_1"]["REGION"] = config_data["S3"]["BUCKET_1"]["REGION"]
     data["S3"]["BUCKET_1"]["ENDPOINT"] = config_data["S3"]["BUCKET_1"]["ENDPOINT"]


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
```diff
@@ -26,12 +26,39 @@
...
     Wait Until Project Is Open    project_title=${project_title}
+    Patch S3 Data Connection With CA Bundle    dc_name=${dc_name}    dc_namespace=${dc_namespace}    aws_s3_ca_bundle=${aws_s3_ca_bundle}
+
+Patch S3 Data Connection With CA Bundle
+    [Documentation]    Patch S3 Data Connection With CA Bundle
+    ...    If 'dc_name' or 'dc_namespace' is missing then this keyword is skipped
+    ...    If 'ca_bundle' is missing then it is taken from global variable S3.AWS_CA_BUNDLE
+    ...    If 'ca_bundle' is still missing then this keyword is skipped
+    [Arguments]    ${dc_name}    ${dc_namespace}=${NONE}    ${aws_s3_ca_bundle}=${NONE}
+    Return From Keyword If
```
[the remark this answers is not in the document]
apodhrad: I do not know what to put here... I tried simple `IF` with `Return    0` but it did not work :(

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
```diff
@@ -26,12 +26,40 @@ Create S3 Data Connection
...
+    ...    '${dc_name}' == '' or '${dc_name}' == '${NONE}' or '${dc_namespace}' == '' or '${dc_namespace}' == '${NONE}'
+    ...    0
+    # I did not find a better way to check if multi-line string is empty or "NONE"
```
tarukumar, line 48: you can use OR keyword, something like below and the check the status using IF block:

`${status}=  Run Keyword And Return Status    Should Be Empty    ${m_string} OR Should Not Contain    ${m_string}    NONE`

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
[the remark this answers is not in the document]
tarukumar: Have you tried like this :

`    IF   '${dc_name}' == '' or '${dc_name}' == '${NONE}' or '${dc_namespace}' == '' or '${dc_namespace}' == '${NONE}'
         RETURN    0
    END`

apodhrad:
The problem with simple comparing multi-line strings (certs are multi-line) like `'{aws_s3_ca_bundle}' == ''` is
```
Invalid IF condition: Evaluating expression ''-----BEGIN CERTIFICATE-----
MIIFYTCCA0mgAwIBAgIURaJI92kpgDRwDdVeLN9DkaVSo+YwDQYJKoZIhvcNAQEL
...
80dd2sLkoUcy2tcoIs7+N9g9jykfSVNeAGH2DSHO6BNLeMZHSQ==
-----END CERTIFICATE-----
' == ''' failed: SyntaxError: EOL while scanning string literal (<string>, line 1)

Variables in the original expression ''${aws_s3_ca_bundle}' == ''' were resolved before the expression was evaluated. Try using ''$aws_s3_ca_bundle' == ''' syntax to avoid that. See Evaluating Expressions appendix in Robot Framework User Guide for more details.
```

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
apodhrad: For now, I think it is ok to compare the length and checking if length is > 6 is also safe

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
```diff
@@ -26,12 +26,41 @@ Create S3 Data Connection
...
+    IF    ${length} <= 6
+        ${aws_s3_ca_bundle}=    Get Variable Value    ${S3.AWS_CA_BUNDLE}    ${NONE}
+    END
+    ${length}=    Get Length    "${aws_s3_ca_bundle}"
+    IF    ${length} > 6
+        ${rc}    ${cert}=    Run And Return Rc And Output
+        ...    echo -n "${aws_s3_ca_bundle}" | base64 -w0 # robocop: disable
+        Should Be Equal As Strings    ${rc}    0
+        ${rc}    ${out}=    Run And Return Rc And Output
+        ...    oc patch secret aws-connection-${dc_name} -n ${dc_namespace} -p '[{"op": "replace", "path": "/data/AWS_CA_BUNDLE", "value" : "${cert}"}]' --type json # robocop: disable
```
jstourac, line 60: Could you assure that the number of spaces is enough here?

```suggestion
        ...    oc patch secret aws-connection-${dc_name} -n ${dc_namespace} -p '[{"op": "replace", "path": "/data/AWS_CA_BUNDLE", "value" : "${cert}"}]' --type json    # robocop: disable
```

Also, just to be sure - `/data/AWS_CA_BUNDLE` is a hardcoded path ton the target, we don't want to use any common variable for that at the moment, right?

jstourac (changes requested):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
```diff
@@ -26,12 +26,41 @@ Create S3 Data Connection
     [Arguments]    ${project_title}    ${dc_name}    ${aws_access_key}    ${aws_secret_access}
     ...            ${aws_s3_endpoint}=${S3.AWS_DEFAULT_ENDPOINT}    ${aws_region}=${S3.AWS_DEFAULT_REGION}
     ...            ${connected_workbench}=${NONE}    ${press_cancel}=${FALSE}    ${aws_bucket_name}=${NONE}
+    ...            ${dc_namespace}=${NONE}
```
bdattoma, line 29: you can use `Get Openshift Namespace From Data Science Project` keyword instead of passing the NS as argument

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
```diff
@@ -26,12 +26,41 @@ Create S3 Data Connection
...
     ...            connected_workbench=${connected_workbench}    press_cancel=${press_cancel}
     ...            aws_bucket_name=${aws_bucket_name}
     Wait Until Project Is Open    project_title=${project_title}
+    Patch S3 Data Connection With CA Bundle
+    ...    dc_name=${dc_name}    dc_namespace=${dc_namespace}    aws_s3_ca_bundle=${aws_s3_ca_bundle}
+
+Patch S3 Data Connection With CA Bundle
+    [Documentation]    Patch S3 Data Connection With CA Bundle
+    ...    If 'dc_name' or 'dc_namespace' is missing then this keyword is skipped
+    ...    If 'ca_bundle' is missing then it is taken from global variable S3.AWS_CA_BUNDLE
```
bdattoma, line 43: we don't have this global variable in the test-variable file. You need another MR for adding it (apart from changing the `ods_ci/utils/scripts/testconfig/generateTestConfigFile.py`

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
bdattoma: why 6?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
```diff
@@ -26,12 +26,41 @@ Create S3 Data Connection
...
+    IF    '${dc_name}' == '' or '${dc_name}' == '${NONE}' or '${dc_namespace}' == '' or '${dc_namespace}' == '${NONE}'
+        RETURN    0
+    END
+    # I did not find a better way to check if multi-line string is empty or "NONE"
+    ${length}=    Get Length    "${aws_s3_ca_bundle}"
+    IF    ${length} <= 6
+        ${aws_s3_ca_bundle}=    Get Variable Value    ${S3.AWS_CA_BUNDLE}    ${NONE}
+    END
+    ${length}=    Get Length    "${aws_s3_ca_bundle}"
+    IF    ${length} > 6
```
bdattoma, line 55: ELSE?
should it fail the test?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
apodhrad: thx, I will try it

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
apodhrad: That global variable is option and is added in this PR at generateTestConfigFile.py

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
apodhrad: Good point, I should add checking for the CA Bundle in the parent keyword `Create S3 Data Connection`

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
apodhrad: We need to set exactly this `AWS_CA_BUNDLE` variable. Then, the data connection will start working with self-signed certificates.

jstourac: Thx, good. Just please check that the number of spaces is enough, I believe there should be at least 2 spaces before the comment otherwise it is passed to the `oc` command  (which doesn't break anything in this case since it's a comment but the robocop disablement won't work this way IIUC).

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/DataConnections.resource
apodhrad: Hi @bdattoma , because `Get Length    "${NONE}"` return `6`

jstourac (approved):

bdattoma (approved):

PR #1142, opened 2024-01-22
Add javascrit to click the radio button

tarukumar:


tarukumar:
job passed: rhods/job/rhods-smoke/4389/

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -330,7 +330,10 @@ Check Card Badges And Return Titles
 
 Open Get Started Sidebar And Return Status
     [Arguments]  ${card_locator}
-    Click Element  xpath:${card_locator}/${CARD_BUTTON_XP}
+    Wait Until Element Is Visible    xpath:${card_locator}/${CARD_BUTTON_XP}
+    Wait Until Element Is Enabled     xpath:${card_locator}/${CARD_BUTTON_XP}    timeout=20s     error=Element is not clickbale  #robocop : disable
+    ${element}=    Get WebElement    xpath:${card_locator}/${CARD_BUTTON_XP}
+    Execute Javascript    arguments[0].click();     ARGUMENTS    ${element}
```
bdattoma, line 336: why do we need to click on the radio button? cards should be clickable now

tarukumar: Manually it works(click on on card) but selenium click element doesn't work with RC1 which we received in the morning. I didn't change any thing i just replace the selenium with java script. You can look at the detail in the auto-triggered job

mwaykole (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
jstourac: So, just want to be sure we really don't want to put back the original behavior to click on the whole tile/card here?

Having this `xpath` called works just fine `//div[@data-testid="pachyderm"]`.

If above is true, I'm okay to merge this as is. Otherwise I vote to click on the whole tile as this is preferred scenario for customer and we wouldn't notice in case of regression again on this.

bdattoma: >Otherwise I vote to click on the whole tile as this is preferred scenario for customer and we wouldn't notice in case of regression again on this.


I agree with this ^^

My understanding was that we were going to remove the radio button at all, instead now we have both radio button + clickable cards. Technically now it keeps working if we click on radio button, my personal preference matches Jan's, but cannot say what this PR does is wrong for sure

bdattoma (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
jstourac: I can see this is merged. Is there any conclusion of this discussion?

PR #1149, opened 2024-01-23
Add test tier description into the README

sutaakar:
Added test tier description to have a source of truth for identifying what tier should be any test assigned to.

This is so far a proposal, feel free to suggest changes.


README.md
```diff
@@ -55,6 +55,24 @@ and its upstream project, [Open Data Hub](https://opendatahub.io/).
...
+    - Tests should cover only critical functionality, verifying that build is usable and can be accepted for testing
+    - Tests should be ideally 100% stable and reliable
+    - Test execution time should be less than 1 minute
+  - Sanity
+    - Base test coverage making sure that core functionality works as expected
+    - Tests should be ideally 100% stable and reliable
+    - Test execution time should be less than 5 minutes
+  - Tier1
+    - Test coverage making sure that overall functionality works as expected
+    - Test execution time should be less than 10 minutes
```
jiridanek, line 70: Can we (later) also document a decision what to do if the time budget for tier1 is exhausted but there are still tier1 tests to be written?

jiridanek (approved):

README.md
jstourac: I understand it that each test has to be less than 10 minutes but the execution time of the whole set can be longer. Because otherwise we don't match at all currently, right? :upside_down_face:

jiridanek: We certainly don't, that's why I was worried these targets might be a bit too ambitious.

Anyways, we should then have timebounds for each test and for suite as a whole; both would provide useful guidance.

sutaakar: Yes, the time limit should be related to single test, will adjust the text.

jstourac: Yeah, I agree.

Also, these facts are reason I don't want to approve this myself, since I don't have the historic knowledge and what we should write in here. I agree that being able to follow what is written here would be ideal. But waiting for guys with more history on this project to express themselves :slightly_smiling_face:

sutaakar: Adjusted test time description, I hope it is more understandable :)

README.md
```diff
@@ -55,6 +55,24 @@ and its upstream project, [Open Data Hub](https://opendatahub.io/).
    * The wrapper script has several arguments and you can find details in the dedicated document file. See [run_args.md](ods_ci/docs/RUN_ARGUMENTS.md)
    * As alternative, you can run any of the test cases by creating the python virual environment, install the packages in [poetry.lock](poetry.lock) and running the `robot` command directly
 
+# Test tiers
+  - Smoke
+    - Tests executed as part of productization pipeline to verify basic RHOAI functionality
+    - Tests should cover only critical functionality, verifying that build is usable and can be accepted for testing
+    - Tests should be ideally 100% stable and reliable
+    - Execution time of every single test should be less than 1 minute
```
jiripetrlik, line 63: I'm not sure about 1 minute timeout. It is often problem to create a pod because of downloading images. I would suggest:
smoke - 3 minutes
sanity - 10 minutes
tier1 - 20 minutes

jiridanek: Pod downloading could be put out of the test, and have one "setup test" that gets some image set prefetched, that the subsequent tests can then use... Limiting the size of this image set (not extending it without a good reason) would also help to make things faster.

Maybe specify the limits for first run and for subsequent runs of the test. I do like the idea of "1 minute test" and would like to preserve it in some form. I was often running a test multiple times recently, when debugging fails.

README.md
jiridanek: The 2 mandatory approval votes here are ofc. insufficient, this is something that needs to be "socialized to the whole team", to use this phrase I recently saw for the first time. I meant my approval vote more like an approval of the effort which I think is helpful guidance, not as a "merge it" stamp.

jgarciao: This PR has been updated just to have a link to the style guide doc. We'll discuss more about the max time per test in the Automatin Meeting

README.md
```diff
@@ -55,6 +55,8 @@ and its upstream project, [Open Data Hub](https://opendatahub.io/).
    * The wrapper script has several arguments and you can find details in the dedicated document file. See [run_args.md](ods_ci/docs/RUN_ARGUMENTS.md)
    * As alternative, you can run any of the test cases by creating the python virual environment, install the packages in [poetry.lock](poetry.lock) and running the `robot` command directly
 
+# Test tiers
+Test tier description is located in [ODS-CI RobotFramework Style Guide doc](https://docs.google.com/document/d/11ZJOPI1uq-0Wl6a2V8fkAv_TQhfzp9t_IjXAheaJxmQ/edit#heading=h.s819p3c5ud7p)
```
jstourac, line 59: Just go to this PR after some time. This has changed quite a lot - what is the meaning to provide a link to the RH internal gDoc in the upstream repository? Either we should provide this information here (if it's not internal) or we shouldn't point this doc here at all.

sutaakar: Will close the PR until this gets clarified.

PR #1150, opened 2024-01-23
Adjust tests for Kuberay 1.0

jiripetrlik:
Tests on master branch are for Kuberay 1.1.

### DIFF
--- a/ods_ci/tests/Tests/650__distributed_workloads/test-run-kuberay-e2e.robot
+++ b/ods_ci/tests/Tests/650__distributed_workloads/test-run-kuberay-e2e.robot
@@ -8,7 +8,7 @@ Resource          ../../../tasks/Resources/RHODS_OLM/install/oc_install.robot
 
 *** Variables ***
 ${KUBERAY_DIR}            kuberay
-${KUBERAY_REPO_URL}       %{KUBERAY_REPO_URL=https://github.com/ray-project/kuberay.git}
+${KUBERAY_REPO_URL}       %{KUBERAY_REPO_URL=https://github.com/red-hat-data-services/kuberay}
 ${KUBERAY_REPO_BRANCH}    %{KUBERAY_REPO_BRANCH=master}
 
 *** Test Cases ***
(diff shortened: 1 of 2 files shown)


ods_ci/tests/Tests/650__distributed_workloads/test-run-kuberay-e2e.robot
```diff
@@ -9,7 +9,7 @@ Resource          ../../../tasks/Resources/RHODS_OLM/install/oc_install.robot
 *** Variables ***
 ${KUBERAY_DIR}            kuberay
 ${KUBERAY_REPO_URL}       %{KUBERAY_REPO_URL=https://github.com/ray-project/kuberay.git}
```
sutaakar, line 11: Can you rather use https://github.com/red-hat-data-services/kuberay as this is the source code for product?

jgarciao (dismissed):

sutaakar:
Liine number 49 is referencing wrong variables:
`FAIL    Unable to clone Kuberay repo ${DW_REPO_URL}:${DW_REPO_BRANCH}:${DW_DIR}`

sutaakar:
Can you add a waiting command after enabling Ray component, waiting for Ray to be up and running (or maybe ready in DSC status)?
Locally when I ran the test it fails, seems like the test already expects Ray to be running, while it takes time for RHOAI to spin up Ray after t is enabled.

sutaakar:
Can you also pass custom Ray image as env variable for test? See https://github.com/red-hat-data-services/kuberay/blob/master/ray-operator/test/support/environment.go#L15 .
Default image refer to dockerhub, in our tests we can reference i.e. https://github.com/project-codeflare/codeflare-common/blob/main/support/defaults.go#L10

jiripetrlik:
@sutaakar Thank you for your comments. It should be all fixed now. I was able to successfully run it also on Jenkins.

sutaakar (approved):
Looks good to me, thanks

jgarciao (approved):

PR #1154, opened 2024-01-24
Fix timing issue when logging into OpenShift. Remove deprecated jupyter test

jgarciao:
- Attempts to fix timing issue when logging into OpenShift Console and page is still loading
- Introduces keyword  `Wait Until Page Contains A String In List`
- Deletes deprecated test related to CUDA builds

Tested successfully in rhods-ci-pr-test/2375 (just a few tests) and partially successfully in rhods-ci-pr-test/2374 (all smoke)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -67,7 +67,11 @@ Authorize rhods-dashboard service account
 
 Login To RHODS Dashboard
    [Arguments]  ${ocp_user_name}  ${ocp_user_pw}  ${ocp_user_auth_type}
-   #Wait Until Page Contains  Log in with
+
+   # Wait until we are in the OpenShift auth page or already in Dashboard
+   ${expected_text_list} =    Create List    Log in with    Data Science Projects
```
bdattoma, line 72: why `Data Science Projects`? we don't use this keyword only to login directly in DS Project page. besides, there is `Wait for RHODS Dashboard to Load` keyword which waits for the dashboard to be loaded no?

jgarciao: I wanted to use a text that always appears in Dashboard, and "Data Science Projects" is always shown in the menu on the left

>Wait for RHODS Dashboard to Load keyword which waits for the dashboard to be loaded no?

This code is at the very beginning of `Login To RHODS Dashboard`. We cannot use `RHODS Dashboard to Load` yet because it is possible we'll be redirected to the login page for Dashboard

bdattoma: okay but why do we need to wait at the very beginning of `Login To RHODS Dashboard` for `Data Science Projects`?

it's probably harmless, but I wonder if this wait wouldn't be covered already by `Wait for RHODS Dashboard to Load` which gets usually called after the Login keyword

bdattoma (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
jgarciao: The problem I found when running the tests in local is that `Login To RHODS Dashboard` sometimes has false in `${oauth_prompt_visible} =  Is OpenShift OAuth Login Prompt Visible` because the page is still loading. 

I added this wait to avoid this error. As we could already have a session with dashboard, the wait should be checking if login is required or if we are already logged into dashboard. That's why it is waiting for "Log in with" or "Data Science Projects"

bdattoma: understood, thanks

ods_ci/tests/Resources/Common.robot
```diff
@@ -93,6 +93,21 @@ CSS Property Value Should Be
...
+    [Arguments]  ${text_list}
+    FOR    ${text}    IN    @{text_list}
+        ${text_found}=    Run Keyword And Return Status    Page Should Contain   ${text}
+        IF  ${text_found}    RETURN
+    END
+    Fail    Current page doesn't contain any of the strings in: @{text_list}
+
+Wait Until Page Contains A String In List
+    [Documentation]    Waits until page contains at least one of the strings in text_list
+    [Arguments]    ${text_list}    ${retry}=12x    ${retry_interval}=5s
```
diegolovison, line 107: I prefer increasing the retry and decreasing the interval to 1 second. This operation won't harm the cpu

diegolovison (approved):

ods_ci/tests/Resources/Common.robot
jgarciao: Now is retrying every 5s. I think it's fine like that. We don't want to stress the cluster :-)

bdattoma (approved):

diegolovison (approved):

PR #1155, opened 2024-01-25
Trivial logging update attempt and increase of timeout

jstourac:
- logging update to make it slightly clearer
- increase of the timeout for the provision and claim check from 5 to 10
  minutes

I don't think this one needs a CI run.

Please let me know whether you think it can be a bit clearer to somebody who's not familiar with the code itself.


manosnoam (approved):
Thanks, looks good

jgarciao (approved):

PR #1158, opened 2024-01-25
Fix multiple Sanity failures in 2.6

lugi0:


lugi0:
Testing in rhods-ci-pr-test/2387

ods_ci/tests/Tests/500__jupyterhub/custom-image.robot
```diff
@@ -72,7 +72,7 @@ Test Duplicate Image
     Import New Custom Image    ${IMG_URL}    ${IMG_NAME}    ${IMG_DESCRIPTION}
     ...    software=${IMG_SOFTWARE}
     ...    packages=${IMG_PACKAGES}
-    Wait Until Page Contains    Unable to add notebook image: ${IMG_NAME}
+    Wait Until Page Contains    Unable to add notebook image: HTTP request failed
```
jstourac, line 75: This test is marked as `ProductBug` now due to https://issues.redhat.com/browse/RHOAIENG-1192 (https://github.com/opendatahub-io/odh-dashboard/issues/2186).

Thus if we really want to do this change here, we should mark it as a workaround and I vote to keep the previous line here. WDYT?

lugi0: >I vote to keep the previous line here. WDYT?
I'm not sure which line you're referring to in this case? The previous error message was already a "workaround" for the ProductBug you've linked, but it appears that in 2.6 the error message has changed. We can certainly highlight that this is a workaround, but which other line would you keep? because if it's `Unable to add notebook image: ${IMG_NAME}` I don't think it makes much sense

jstourac: I'm referring to the original line:

```
Wait Until Page Contains    Unable to add notebook image: ${IMG_NAME}
```

TBH, I don't recall anymore if I put it there as a workaround or that I simply estimated how the message will look after the fix. Based on the description of https://issues.redhat.com/browse/RHOAIENG-1192, the `HTTP request failed` was in the product in version 2.4 already. Since this is marked as a `ProductBug`, I suppose that these shouldn't be run in our suite and thus this couldn't pass until the fix is provided.

If we want to make this pass, we should remove the `ProductBug` and log a warning referring to this issue maybe so we're aware we should update this sometime?

lugi0: IMHO we can go with the warning and note about the bug, but given that the error message format has changed I'd still leave the previous version removed

jiridanek:
I clicked Update branch to get in this change to make GitHub Actions workflows green

* https://github.com/red-hat-data-services/ods-ci/pull/1166

jstourac (approved):

bdattoma (approved):

PR #1172, opened 2024-01-31
Enhancements for Trusty Tests

nehachopra27:
1) Adds wait between the trusty service install and verifying the trusty route is created or not.
2) Adding wait betwen verifying the. metrics created and prometheus scraping data time.
3) Renaming the bias metrics file


diegolovison (approved):

jgarciao (approved):
LGTM. 

Consider adding in a follow up PR a new keyword in ods_ci/tests/Resources/OCP.resource `Wait Until Route Is Ready`, that could use `oc wait --for=condition=... --timeout=XXXs`

PR #1177, opened 2024-02-06
[Fix] wait for the RHOAI/ODH operator CSV to be in state condition Su…

jstourac:
…cceeded

The installation of ODH often failed due to the fact that we tried to apply the DSC on the cluster even before such CRD resource is known to the cluster (that means that operator itself hasn't been installed completely yet). This change should wait for the installation to be prepared for the DSC application. Timeout is set to 2 minutes.

---

CI: `odh/job/autotrigger-smoke/11` for ODH :white_check_mark:
CI: `rhoai/job/autotrigger-smoke/11` for RHOAI :white_check_mark:

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -13,17 +13,21 @@ ${SERVERLESS_SUB_NAME}=    serverless-operator
 ${SERVERLESS_NS}=    openshift-serverless
 ${SERVICEMESH_OP_NAME}=     servicemeshoperator
 ${SERVICEMESH_SUB_NAME}=    servicemeshoperator
+${RHODS_OPERATOR_NAME}=    rhods-operator
+${ODH_OPERATOR_NAME}=    opendatahub-operator
 
 *** Keywords ***
 Install RHODS
   [Arguments]  ${cluster_type}     ${image_url}
   Install Kserve Dependencies
   Clone OLM Install Repo
+  ${operator_csv_name} =    Set Variable    ${RHODS_OPERATOR_NAME}
   IF  "${cluster_type}" == "selfmanaged"
       IF  "${TEST_ENV}" in "${SUPPORTED_TEST_ENV}" and "${INSTALL_TYPE}" == "CLi"
             IF  "${UPDATE_CHANNEL}" != "odh-nightlies"
                  Install RHODS In Self Managed Cluster Using CLI  ${cluster_type}     ${image_url}
             ELSE
+                 ${operator_csv_name} =    Set Variable    ${ODH_OPERATOR_NAME}
                  Create Catalog Source For Operator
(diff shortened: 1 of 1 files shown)


lugi0 (approved):

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -41,13 +45,31 @@ Install RHODS
...
+  ...    Csv Is Ready    ${csv_name}
+  Log    Operator '${csv_name}' CSV is in 'Succeeded' status condition now, let's continue    console=yes
+
+Csv Is Ready
+  [Documentation]   Check whether given CSV to be in Succeeded status condition
+  [Arguments]    ${csv_name}
+  ${rc}    ${output} =    Run And Return Rc And Output
+  ...    oc get csv --namespace openshift-operators --output=json | jq --raw-output --exit-status '.items[] | select(.metadata.name | test("${csv_name}")).status.conditions[] | select(.phase == "Succeeded").phase'    # robocop: disable:line-too-long
+  Should Be Equal As Integers    ${rc}    0
+  Should Be Equal As Strings    Succeeded    ${output}
```
bdattoma, line 72: we have kw which check the subscription and looks for `AllCatalogSourcesHealthy` condition. Isn't it equivalent?

```
    Wait Until Operator Subscription Last Condition Is
    ...    type=CatalogSourcesUnhealthy    status=False
    ...    reason=AllCatalogSourcesHealthy    subcription_name=${SUBSCRIPTION_NAME}
    ...    namespace=${NAMESPACE}
```

jstourac: Yeah, TBH, I'm not really sure what is the proper way here. I saw that the subscription marks `CatalogSourcesUnhealthy = false` earlier than the CSV sets `Succeeded = InstallSucceeded` with message `install strategy completed with no errors`. Thus, I though that better way how to check for this is actually checking CSV status. But maybe I'm wrong here, please point me to some source of truth :shrug:

tarukumar (approved):
/lgtm

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
bdattoma: hmm I imagine it's a matter of microseconds/seconds? anyways, okay.

Will you have another PR to actually use the keyword?

bdattoma (approved):

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
jstourac: Well, the difference is definitely at least seconds. In case of ODH, it looks like the installation take a bit more time than for RHOAI for some reason, where we didn't hit this issue :shrug:

I don't plan to use the other keyword here or this one on other places unless we face an issues, I guess :slightly_smiling_face:

PR #1178, opened 2024-02-06
Decrease TTL for our DNS records for the cluster creation.

jstourac:
At the moment we have 2days TTL for the DNS records for the cluster IPs. Since this causes us troubles when we want to recreate the cluster with the very same name (the records is probably still in the DNS cache of the system where the new cluster is being created), let's reduce this TTL value to 5minutes only.

The overhead with the DNS shouldn't be hopefully that bad.

---

CI: autotrigger-smoke/6 :white_check_mark: - cluster creation failed due to different reason, although we can clearly see that the DNS records have expected TTL set now.


apodhrad (dismissed):
I recommend 300 but 600 is ok as well

jstourac:
> I recommend 300 but 600 is ok as well

I updated to 300 (5mins) in the end :slightly_smiling_face: 

Looks like this may have a chance to get a support. So I'll try some run with this change and remove WIP once verified.

apodhrad (approved):

PR #1181, opened 2024-02-06
Add Kueue tests to Tier2

jiripetrlik:
It seems we have 2.7 pre-release build and therefore we should probably start running these tests. @ChughShilpa @sutaakar What do you think?


ChughShilpa:
One more change i would like to request @jiripetrlik @sutaakar 
Current kueue tests are using 'main' branch and i think we should change it to 'dev' branch for testing against RC builds

This change is not related to this PR, and we can have different PR as well if we don't want to incorporate in this PR

sutaakar (approved):

ChughShilpa (approved):

sutaakar (approved):

PR #1179, opened 2024-02-06
Add log gather when pod doesn't get ready by timeout

bdattoma:
When `Wait For Pods To Be Ready` hits the timeout, we can't tell why it happened. So, trying to get clear info by capturing pod logs and namespace events (on failure only)

PR #1184, opened 2024-02-07
Add event and pod logs fetch in Model Mesh tests

bdattoma:
Following what was implemented in https://github.com/red-hat-data-services/ods-ci/pull/1179, i'm adding the capture of events and logs in MM tests when deployment and query checks fail.


jstourac (dismissed):
lgtm, once tested successfully :slightly_smiling_face:

bdattoma:
PR Validation:

1. Model Mesh Sanity: `rhods-ci-pr-test/2423`
2. Regression test on LLM tests (https://github.com/red-hat-data-services/ods-ci/pull/1179): _coming soon_

FedeAlonso (approved):

manosnoam (approved):
Thanks @bdattoma 
lgtm.
FYI, soon I'll add the new test and logic into the same 420 suite, which revealed the recent model deployment bugs (RHOAIENG-2759,  RHOAIENG-2724)

ods_ci/tests/Resources/OCP.resource
```diff
@@ -95,6 +95,17 @@ Fetch CA Certificate If RHODS Is Self-Managed
...
 
+Get Events And Pod Logs
+    [Documentation]    Fetches the namespace events and the pod logs (all containers).
+    ...                It prints and return both items
+    [Arguments]    ${namespace}    ${label_selector}
+    ${rc}    ${events}=    Run And Return Rc And Output    oc get events -n ${namespace}
+    Log    ${events}
+    ${rc}    ${podlogs}=    Run And Return Rc And Output
+    ...    oc logs -l ${label_selector} --all-containers -n ${namespace}
+    Log    ${podlogs}
```
lugi0, line 106: I don't know if it's a good idea to print all these logs, as they might be very long. IMHO it would be better to write them to a file.

ods_ci/tests/Tests/100__deploy/120__upgrades/120__pre_upgrades.robot
```diff
@@ -96,16 +99,17 @@ Verify Model Can Be Deployed Via UI For Upgrade
...
     Run Keyword And Continue On Failure  Wait Until Keyword Succeeds  5 min  10 sec  Verify Serving Service
     Verify Model Status    ${MODEL_NAME}    success
     Set Suite Variable    ${MODEL_CREATED}    True
     Run Keyword And Continue On Failure    Verify Model Inference    ${MODEL_NAME}    ${INFERENCE_INPUT_OPENVINO}    ${EXPECTED_INFERENCE_OUTPUT_OPENVINO}    token_auth=${FALSE}
     Remove File    openshift_ca.crt
-    [Teardown]   Dashboard Test Teardown
+    [Teardown]   Run Keywords    Dashboard Test Teardown
+    ...    AND
+    ...    Run Keyword If Test Failed    Get Events And Pod Logs    namespace=${PRJ_TITLE}
+    ...    label_selector=name=modelmesh-serving-${runtime_pod_name}
```
lugi0, line 112: This might not work if the test fails before entering execution, i.e. during the Setup. Perhaps it would be a good idea to initialize runtime_pod_name at a suite level and run a check against it before running this keyword?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -51,13 +54,13 @@ Verify Model Can Be Deployed Via UI
...
     ...    data_connection_name=model-serving-connection    model_path=mnist-8.onnx
-    ${runtime_pod_name} =    Replace String Using Regexp    string=${RUNTIME_NAME}    pattern=\\s    replace_with=-
-    ${runtime_pod_name} =    Convert To Lower Case    ${runtime_pod_name}
     Run Keyword And Continue On Failure  Wait Until Keyword Succeeds
     ...  5 min  10 sec  Verify Openvino Deployment    runtime_name=${runtime_pod_name}
     Run Keyword And Continue On Failure  Wait Until Keyword Succeeds  5 min  10 sec  Verify Serving Service
     Verify Model Status    ${MODEL_NAME}    success
     Set Suite Variable    ${MODEL_CREATED}    True
+    [Teardown]    Run Keyword If Test Failed    Get Events And Pod Logs    namespace=${PRJ_TITLE}
+    ...    label_selector=name=modelmesh-serving-${runtime_pod_name}
```
lugi0, line 63: Same comment on the runtime pod name

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -87,13 +92,13 @@ Verify Openvino_IR Model Via UI
...
     ...    data_connection_name=model-serving-connection    model_path=openvino-example-model
-    ${runtime_pod_name} =    Replace String Using Regexp    string=${RUNTIME_NAME}    pattern=\\s    replace_with=-
-    ${runtime_pod_name} =    Convert To Lower Case    ${runtime_pod_name}
     Run Keyword And Continue On Failure  Wait Until Keyword Succeeds
     ...  5 min  10 sec  Verify Openvino Deployment    runtime_name=${runtime_pod_name}
     Run Keyword And Continue On Failure  Wait Until Keyword Succeeds  5 min  10 sec  Verify Serving Service
     Verify Model Status    ${MODEL_NAME}    success
     Set Suite Variable    ${MODEL_CREATED}    True
+    [Teardown]    Run Keyword If Test Failed    Get Events And Pod Logs    namespace=${PRJ_TITLE}
+    ...    label_selector=name=modelmesh-serving-${runtime_pod_name}
```
lugi0, line 101: Same comment on the runtime pod name

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -127,7 +132,8 @@ Verify Tensorflow Model Via UI
     ${status_code}    ${response_text} =    Send Random Inference Request     endpoint=${url}    name=input
     ...    shape={"B": 1, "H": 299, "W": 299, "C": 3}    no_requests=1
     Should Be Equal As Strings    ${status_code}    200
-
+    [Teardown]    Run Keyword If Test Failed    Get Events And Pod Logs    namespace=${PRJ_TITLE}
+    ...    label_selector=name=modelmesh-serving-${runtime_pod_name}
```
lugi0, line 136: Same comment on the runtime pod name

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/421__model_serving_gpu.robot
```diff
@@ -50,6 +51,9 @@ Verify GPU Model Deployment Via UI
     Should Be Equal As Strings    ${type}    "g4dn.xlarge"
     Verify Model Status    ${MODEL_NAME}    success
     Set Suite Variable    ${MODEL_CREATED}    True
+    [Teardown]    Run Keyword If Test Failed    Get Events And Pod Logs    namespace=${PRJ_TITLE}
+    ...    label_selector=name=modelmesh-serving-${runtime_pod_name}
+
```
lugi0, line 56: Same comment on the runtime pod name, also too many empty lines between the two tests?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/423__model_serving_customruntimes.robot
```diff
@@ -89,6 +91,8 @@ Verify RHODS Users Can Deploy A Model Using A Custom Serving Runtime
     Verify Model Inference With Retries    ${model_name}    ${inference_input}    ${exp_inference_output}
     ...    token_auth=${TRUE}
     ...    project_title=${PRJ_TITLE}
+    [Teardown]    Run Keyword If Test Failed    Get Events And Pod Logs    namespace=${PRJ_TITLE}
+    ...    label_selector=name=modelmesh-serving-${runtime_pod_name}
```
lugi0, line 95: Same comment on the runtime pod name

ods_ci/tests/Tests/100__deploy/120__upgrades/120__pre_upgrades.robot
bdattoma: if the test fails at suite setup leve there won't be nothing to fetch logs from

ods_ci/tests/Resources/OCP.resource
bdattoma: it is done in this way to we can see them soon instead of going finding files, besides we should handle the filenames as well to avoid overwriting.
I believe it's worth to try having the logs here. If it really helps and we find suboptimal to have all the logs printed here we can give another round. Too many changes in this PR for now IMO

ods_ci/tests/Tests/100__deploy/120__upgrades/120__pre_upgrades.robot
lugi0: Yes, that's the problem, but it will also fail trying to access a variable that does not exist (IIRC the Teardown is always executed) - if we initialize the var with a known name (e.g. "no-pod-yet") we can have a small check within the teardown to make sure that the variable exists and it doesn't have that name, thus avoiding potential problems of no logs to fetch.

bdattoma: what I mean to say is that if the test fails at setup, there won't be no pod to check

ods_ci/tests/Resources/OCP.resource
bdattoma: >  the logs overall less readable.


the log don't get visible unless you open the `Log` statement in the HTML


> For me it's ok to try this approach ad re-evaluate down the line though


okay, thx!

ods_ci/tests/Tests/100__deploy/120__upgrades/120__pre_upgrades.robot
bdattoma: done!

lugi0: That wasn't what I had in mind, you could check against the value of the variable and decide to not fetch logs from the pod(s) but rather only the events of the namespace. 
I'll approve for now but imho this could be improved.

bdattoma: you're right sorry, I mixed it up a bit. Anyways, I don't think there are many benefits from that logic checking the `no-pod-yet` because at the end of the the test would be failed in any case if the pod didn't come up. no?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
bdattoma: discussed in https://github.com/red-hat-data-services/ods-ci/pull/1184#discussion_r1481752983

jstourac (approved):
lgtm in general; don't forget to squash the commits before merge :slightly_smiling_face:

lugi0 (approved):

PR #1195, opened 2024-02-13
Another fix to `Wait Until Csv Is Ready` keyword

manosnoam:
Fixing the `oc wait` command that ended too soon for RHODS operator CSV verification in previous PR #1191 .

This is the output after the fix:
image


jstourac (approved):
Approving from mobile just to unblock us. I would rather welcome a simple revert and do a proper check tomorrow on PC.

Anyway, if this is tested properly, please go ahead.

jiridanek (approved):

PR #1203, opened 2024-02-15
Fix 'Clean All Models Of Current User' KW

manosnoam:
In case the current user does not have permissions to view other DS projects, then the Model Serving page does not include "select a project" link, which causes the 'Clean All Models Of Current User' KW to fail.

The fix includes:
- Skip cleanup if no Data Sciense projects exist
- 'Set Up Project' SETUP KW continues if cleanup failed
- 'Set Up Project' SETUP KW creates DS project only if doesn't exist yet

image

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
@@ -527,18 +527,23 @@ Clean All Models Of Current User
     ...    controls how many retries are made.
     [Arguments]    ${retries}=5
     Open Model Serving Home Page
-    FOR    ${try}    IN RANGE    0    ${retries}
-        ${status}=    Run Keyword And Return Status    Page Should Contain    Select a project
-        IF    ${status}
-            Capture Page Screenshot
-            Click Button    Select a project
-            RETURN
-        ELSE
-            Clean Up Model Serving Page
-            Clean Up DSP Page
-            Open Model Serving Home Page
-            Reload Page
-            Sleep  5s
+    ${projects_empty}=    Run Keyword And Return Status    Page Should Contain    No data science projects
+    IF    ${projects_empty}
+        Log    No Data Sciense projects exist, skipping cleanup    console=True
+    ELSE
+        FOR    ${try}    IN RANGE    0    ${retries}
+            ${status}=    Run Keyword And Return Status    Page Should Contain    Select a project
+            IF    ${status}
+                Capture Page Screenshot
+                Click Button    Select a project
+                RETURN
+            ELSE
+                Clean Up Model Serving Page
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -547,11 +552,10 @@ Set Up Project
     ...                creates caikit runtime. This can be used as test setup
     [Arguments]    ${namespace}    ${single_prj}=${TRUE}    ${enable_metrics}=${FALSE}    ${dc_name}=kserve-connection
     IF    ${single_prj}
-        Clean All Models Of Current User
+        Run Keyword And Continue On Failure    Clean All Models Of Current User
     END
     Open Data Science Projects Home Page
-    Wait For RHODS Dashboard To Load    wait_for_cards=${FALSE}    expected_page=Data Science Projects
```
bdattoma, line 553: why?

manosnoam: Because that's what the KW docstring says:

> Set Up Project
>     [Documentation]    Creates the DS Project (if not exists)

FedeAlonso (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
bdattoma: > Because Open Data Science Projects Home Page already does it :)

Ok

I wasn't questioning the create DS project part :)

manosnoam:
@bdattoma I've also verify the fix by running another suite that uses 'Clean All Models Of Current User' KW at suite's Setup and Teardown, and it works as expected:
image

The failed test (Verify Secure Model Can Be Deployed In Same Project) is due to a recent product bug: https://issues.redhat.com/browse/RHOAIENG-2759

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -566,11 +571,10 @@ Set Up Project
...
-    Create Data Science Project    title=${namespace}    description=kserve test project
+    Create Data Science Project    title=${namespace}    description=kserve test project    existing_project=${TRUE}
```
lugi0, line 577: This is the only line I'm not sure about, specifically `existing_project=${TRUE}`.
Could you test the change against `422__model_serving_caikit_py_client.robot`, `422__model_serving_llm_other_runtimes_UI.robot` and `422__model_serving_llm_UI.robot`?

lugi0 (approved):
LGTM but please test against the kserve UI suites before merging

PR #1211, opened 2024-02-16
Fix modelmesh suite, add test for productbug RHOAIENG-2869

lugi0:


ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -129,10 +133,47 @@ Test Inference With Token Authentication
...
+    ...    num_projects=4
+    [Teardown]    Run Keyword If Test Failed    Get Modelmesh Events And Logs
+    ...    server_name=${RUNTIME_NAME}    project_title=${PRJ_TITLE}
+
+Verify Editing Existing Model Deployment
+    [Documentation]    Tries editing an existing model deployment to see if the underlying deployment is updated
+    [Tags]    Sanity    Tier1
+    ...       ProductBug    RHOAIENG-2869
+    Open Data Science Projects Home Page
+    Create Data Science Project    title=${PRJ_TITLE}    description=${PRJ_DESCRIPTION}
```
manosnoam, line 148: To regress bug RHOAIENG-2869 we need to update existing model, which is tied to an existing DS project. 

i.e. we need to call this KW with `existing_project=${TRUE}` (and only after project `${PRJ_TITLE}` already exists, with same model name `${MODEL_NAME}`.

```suggestion
    Create Data Science Project    title=${PRJ_TITLE}    description=${PRJ_DESCRIPTION}    existing_project=${TRUE}
```

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -129,10 +133,47 @@
...
+    Create Model Server    token=${FALSE}    server_name=${RUNTIME_NAME}    existing_server=${TRUE}
+    Serve Model    project_name=${PRJ_TITLE}    model_name=${MODEL_NAME}    framework=tensorflow
+    ...    existing_data_connection=${TRUE}    data_connection_name=model-serving-connection
+    ...    model_path=inception_resnet_v2.pb
+    ${runtime_pod_name}=    Replace String Using Regexp    string=${RUNTIME_NAME}    pattern=\\s    replace_with=-
+    ${runtime_pod_name}=    Convert To Lower Case    ${runtime_pod_name}
+    Wait Until Keyword Succeeds    5 min  10 sec  Verify Openvino Deployment    runtime_name=${RUNTIME_POD_NAME}
+    Wait Until Keyword Succeeds    5 min  10 sec  Verify Serving Service
+    Verify Model Status    ${MODEL_NAME}    success
+    Set Suite Variable    ${MODEL_CREATED}    ${TRUE}
```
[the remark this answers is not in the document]
manosnoam: Nice tip, starting from Robot Framework 7.0 we can change to:
 ```suggestion
    VAR    ${MODEL_CREATED}    ${TRUE}
```

File: ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot [hunk 3]
```diff
@@ -129,10 +133,47 @@
...
+    ${url}    ${kserve}=    Get Model Route via UI    ${MODEL_NAME}
```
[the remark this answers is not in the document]
manosnoam: `Get Model Route Via UI` - In case you want to change everywhere

manosnoam (changes requested):
One comment how we should regress bug https://issues.redhat.com/browse/RHOAIENG-2869

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -129,10 +133,47 @@ Test Inference With Token Authentication
...
+    ${status_code}    ${response_text}=    Send Random Inference Request     endpoint=${url}    name=input
+    ...    shape={"B": 1, "H": 299, "W": 299, "C": 3}    no_requests=1
+    Should Be Equal As Strings    ${status_code}    200
+    Serve Model    project_name=${PRJ_TITLE}    model_name=${MODEL_NAME}    framework=openvino_ir
+    ...    existing_data_connection=${TRUE}    data_connection_name=model-serving-connection
+    ...    model_path=openvino-example-model    existing_model=${TRUE}
```
bdattoma, line 168: I wonder if it would be more clear to have a keyword `Edit Served Model`

bdattoma (approved):

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: This is done in lines 166 to 168, where we try editing the existing model and querying it; I've done it this way in order to decouple this test from other ones, otherwise we'd always need to expect a DS Project to already exist with a given name and deployed model

manosnoam: Although in line 168 you use `existing_model=${TRUE}`, it will still be a different instance, even if the model name is the same. 
This is because we can have same model name across different DS projects, so it's different model instance.
To edit existing model, you must edit **_existing model in an existing DS project_**.

BTW,  see the new test `Verify Multiple Projects With Same Model` where we use the same model name, but in different project names, so we get multiple models with same name across different projects.

manosnoam: Oh I think I missed line 153, where you already created that model name in the project...
In that case, we're all good :)

manosnoam (approved):
minor linter comments if you want to change, not critical

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
bdattoma: I find it less readable and predictable

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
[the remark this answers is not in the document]
lugi0: Doing it in #1212

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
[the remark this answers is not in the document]
lugi0: We have yet to upgrade to RF7

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
lugi0: Sticking with this format for now, we can discuss tomorrow if we want to change it everywhere

PR #1212, opened 2024-02-16
Add OVMS on Kserve GPU test

lugi0:


lugi0:
Verified in rhods-ci-pr-test/2512 but need to figure out why the nvidia prometheus query is returning 0

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
```diff
@@ -95,6 +100,49 @@ Verify Multiple Projects With Same Model (Kserve)
...
+    IF  ${kserve}
+        Fetch Knative CA Certificate    filename=openshift_ca_istio_knative.crt
+    END
+    Send Random Inference Request     endpoint=${url}    no_requests=100
+    # Verify metric DCGM_FI_PROF_GR_ENGINE_ACTIVE goes over 0
+    ${prometheus_route}=    Get OpenShift Prometheus Route
+    ${sa_token}=    Get OpenShift Prometheus Service Account Token
+    ${expression}=    Set Variable    DCGM_FI_PROF_GR_ENGINE_ACTIVE
+    ${resp}=    Prometheus.Run Query    ${prometheus_route}    ${sa_token}    ${expression}
+    Log    DCGM_FI_PROF_GR_ENGINE_ACTIVE: ${resp.json()["data"]["result"][0]["value"][-1]}
```
jstourac, line 139: Does it make sense to log more than just the particular result? In case of any issue, we can see whole response, WDYT?

jstourac (commented):
I put one comment, otherwise LGTM, but I don't have much knowledge behind this :slightly_smiling_face:

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
lugi0: The full query is already logged in the `Run Query` keyword, but you have to drill into the logs to see it. The "top level" log here is only showing the value we care about for quick validation - imho it makes sense as is, but let me know if you feel otherwise
image

jstourac: oh, okay - if it is in the log already, I think it's okay as it is then

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
```diff
@@ -95,6 +100,49 @@ Verify Multiple Projects With Same Model (Kserve)
...
+    ...            aws_bucket_name=ods-ci-s3
+    Deploy Kserve Model Via UI    model_name=${MODEL_NAME_GPU}    serving_runtime=OpenVINO Model Server
+    ...    data_connection=model-serving-connection    path=vehicle-detection-kserve    model_framework=openvino_ir
+    ...    no_gpus=1
+    Wait For Pods To Be Ready    label_selector=serving.kserve.io/inferenceservice=${MODEL_NAME_GPU}
+    ...    namespace=${PRJ_TITLE_GPU}
+    Verify Displayed GPU Count In Single Model Serving    model_name=${MODEL_NAME_GPU}    no_gpus=1
+    ${requests}=    Get Container Requests    namespace=${PRJ_TITLE_GPU}
+    ...    label=serving.kserve.io/inferenceservice=${MODEL_NAME_GPU}    container_name=kserve-container
+    Should Contain    ${requests}    "nvidia.com/gpu": "1"
```
bdattoma, line 122: you could re-use 
```suggestion
Container Hardware Resources Should Match Expected    container_name=kserve-container
    ...    pod_label_selector=serving.kserve.io/inferenceservice=${model_name}
    ...    namespace=${test_namespace}    exp_requests=${requests}    exp_limits=${limits}
```

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
```diff
@@ -95,6 +100,49 @@ Verify Multiple Projects With Same Model (Kserve)
...
+    ${node}=    Get Node Pod Is Running On    namespace=${PRJ_TITLE_GPU}
+    ...    label=serving.kserve.io/inferenceservice=${MODEL_NAME_GPU}
+    ${type}=    Get Instance Type Of Node    ${node}
+    Should Be Equal As Strings    ${type}    "g4dn.xlarge"
```
bdattoma, line 126: is there a way to check the node has gpu without pinning to the node title? We may run on different GPU types

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
```diff
@@ -95,6 +100,49 @@ Verify Multiple Projects With Same Model (Kserve)
...
+    Verify Model Status    ${MODEL_NAME_GPU}    success
+    Set Suite Variable    ${MODEL_CREATED}    True
+    ${url}    ${kserve}=    Get Model Route via UI    ${MODEL_NAME_GPU}
+    IF  ${kserve}
+        Fetch Knative CA Certificate    filename=openshift_ca_istio_knative.crt
+    END
```
bdattoma, line 132: could you remove the IF/ELSE from the test case? You could move this logic in the Suite/test setup. I see that in `OVMS On Kserve Suite Setup` there is already `Fetch CA Certificate If RHODS Is Self-Managed` which works for MM only. You could replace it with the one for KNative

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
```diff
@@ -148,3 +196,15 @@ OVMS On Kserve Suite Teardown
...
+    [Documentation]    Fetches the route for the Prometheus instance of openshift-monitoring
+    ${host}=    Run    oc get route prometheus-k8s -n openshift-monitoring -o json | jq '.status.ingress[].host'
+    ${host}=    Strip String    ${host}    characters="
+    ${route}=    Catenate    SEPARATOR=    https://    ${host}
+    RETURN    ${route}
+
+Get OpenShift Prometheus Service Account Token
+    [Documentation]    Returns a token for a service account to be used with Prometheus
+    ${token}=    Run    oc create token prometheus-k8s -n openshift-monitoring --duration 10m
+    RETURN    ${token}
```
bdattoma, line 210: could you move these 2 keywords to a common file like `Resources/Page/ODH/Monitoring/Monitoring.resource`?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
lugi0: Done

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
lugi0: Done

lugi0:
Validated again with rhods-ci-pr-test/2514, the TC fails but it's tagged with it's own product bug ID

jstourac (approved):

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
```diff
@@ -95,6 +98,48 @@ Verify Multiple Projects With Same Model (Kserve)
...
+    Wait For Pods To Be Ready    label_selector=serving.kserve.io/inferenceservice=${MODEL_NAME_GPU}
+    ...    namespace=${PRJ_TITLE_GPU}
+    Verify Displayed GPU Count In Single Model Serving    model_name=${MODEL_NAME_GPU}    no_gpus=1
+    Container Hardware Resources Should Match Expected    container_name=kserve-container
+    ...    pod_label_selector=serving.kserve.io/inferenceservice=${MODEL_NAME_GPU}
+    ...    namespace=${PRJ_TITLE_GPU}    exp_requests=${requests}    exp_limits=${limits}
+    ${node}=    Get Node Pod Is Running On    namespace=${PRJ_TITLE_GPU}
+    ...    label=serving.kserve.io/inferenceservice=${MODEL_NAME_GPU}
+    ${type}=    Get Instance Type Of Node    ${node}
+    Should Be Equal As Strings    ${type}    "g4dn.xlarge"
```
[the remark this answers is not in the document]
lugi0: Yes, as I said in another comment above the only other option I can think of is looking at the labels, but that is dependent on the nfd operator / nvidia label which wouldn't work for other accelerator types

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
bdattoma: we'll need to make the test compatible with different GPU node types. I think it can be done in another PR once we understand how

bdattoma (approved):

PR #1215, opened 2024-02-19
Update Dockerfile to use minimal CentOS and microdnf

jiridanek:
The Dockerfile has been updated to use a minimal version of CentOS. Furthermore, `dnf` commands have been replaced with `microdnf` to save space and enhance performance. The dependency installation steps were also adjusted accordingly.


bdattoma:
ensure to test an entire Smoke run at least

jstourac:
If it works, I have no objections :slightly_smiling_face: just please, be aware that we don't use it in our jobs at the moment, so to test it one should do a local run, otherwise we break workflow for the perf guys.

jiridanek:
My docker is broken?

```
76.40 Downloading packages...
104.7 error: Downloading from https://mirror.karneval.cz/pub/linux/centos-stream/9-stream/AppStream/x86_64/os/Packages/poppler-data-0.4.9-9.el9.noarch.rpmwas successful but error encountered while checksumming: fsync failed: Input/output error
```

PR #1216, opened 2024-02-19
Update ODS-CI image with Python version arg = 3.11

manosnoam:


ods_ci/build/Dockerfile
```diff
@@ -60,9 +60,10 @@ RUN  chmod +x ods_ci/build/run.sh &&\
 COPY pyproject.toml .
 COPY poetry.lock .
 COPY README.md .
+RUN alternatives --install /usr/local/bin/python3 python3 /usr/bin/python${PYTHON_VERSION} 1
```
jiridanek, line 63: This line should be the only functional change in the PR. It looks ok to me in principle. It can break things, for example we will not have `python3-distro` installed on python3.11, so if there's something that needs that package, it will break. But overall, feels reasonably safe.

jiridanek (approved):

ods_ci/build/Dockerfile
bdattoma: ensure to run an entire Smoke job before merging it.

About distro, https://github.com/red-hat-data-services/ods-ci/pull/817

jiridanek: > closing this one because it's not high in the priority queue, besides this is not causing issues.

hopefully this will remain true

jstourac (approved):

manosnoam:
Built and verified image by running: 
`podman run --rm -v $PWD/.setup/test-variables.yml:/tmp/ods-ci/ods_ci/test-variables.yml:Z -v $PWD/ods_ci/test-output:/tmp/ods-ci/ods_ci/test-output:Z -e RUN_SCRIPT_ARGS='--skip-oclogin false --includ
e Smoke' ods-ci:latest`

All tests (except bias tests that are not supported on 2.7) passed.

jstourac:
For the record - this is a follow-up of #1213.

PR #1225, opened 2024-02-22
Add error message and continue on failure of bug RHOAIENG-404

manosnoam:
Due to bug RHOAIENG-404 there are arbitrary failures during execution
of `Wait Until Cards Are Loaded` KW in Dashboard tests.
For example:

image

With this PR it continues if the error above occurs, and shows an error message about potential bug RHOAIENG-404, in order not to break arbitrary tests (or tests setup).

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
@@ -256,8 +256,8 @@ Load Expected Data Of RHODS Explore Section
 
 Wait Until Cards Are Loaded
     [Documentation]    Waits until the Application cards are displayed in the page
-    Wait Until Page Contains Element    xpath:${CARDS_XP}
-    ...    timeout=15s
+    Run Keyword And Continue On Failure    Wait Until Page Contains Element
+    ...    xpath:${CARDS_XP}    timeout=15s    error="This might be caused by bug RHOAIENG-404"
 
 Get App ID From Card
     [Arguments]  ${card_locator}


FedeAlonso (dismissed):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -109,6 +109,7 @@ Wait for RHODS Dashboard to Load
     IF    ${wait_for_cards} == ${TRUE}
         Wait Until Cards Are Loaded
     END
+    [Teardown]   Run Keyword If Test Failed    Capture Page Screenshot
```
FedeAlonso, line 112: why are we adding this in the Teardowns? Shouldn't be something that has to be done by default if the test fails?

manosnoam: Yes, you're right. But it doesn't capture the screenshot in test setup...
I think it's a bug in RF Selenium Library, so I opened an issue for them: https://github.com/robotframework/SeleniumLibrary/issues/1887

diegolovison (approved):

bdattoma:
This may not solve the issue, discussing offline

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
manosnoam: I've reverted this change, as it was caused by protected kw that hides screenshots.

bdattoma (approved):

FedeAlonso (approved):

PR #1228, opened 2024-02-26
[Fix] jupyter notebook spawn checks

jstourac:
This is a fix for the recent change [1,2]

* [1] #1102
* [2] 60fffb510a5e724e4a7ee3e400eb166843c8e6d8

---

CI:

Screenshot from 2024-02-26 10-59-28

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
+++ b/ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
@@ -274,12 +274,12 @@ Notebook Expected Ide
     ...              At the moment, there are two IDE types available in the notebook images:
     ...              JupyterLab and VSCode.
     ...              Returns type of expected IDE for the given image name as a string.
-    [Arguments]  ${image_name}
-    IF  "${image_name}"=="${EMPTY}"
+    [Arguments]  ${image}
+    IF  "${image}"=="${EMPTY}"
         Log  level=ERROR  message=No image name has been provided!
         RETURN  ${EMPTY}
     END
-    IF  "${image_name}"=="code-server"
+    IF  "${image}"=="code-server"
         RETURN  VSCode
     ELSE
         RETURN  JupyterLab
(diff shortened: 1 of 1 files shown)


jiridanek (approved):
I'd expect variable `image` to hold something like `"quay.io/modh/notebooks/...@sha256-..."`, so maybe image_name is a slightly better name. But up to you, I guess there are reasons I'm not seeing.

ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -288,8 +288,8 @@ Notebook Expected Ide
 Wait Notebook To Be Loaded
     [Documentation]  Waits for the notebook IDE environment to be loaded completely and performs
     ...              a simple check with the Menu action in the loaded IDE.
-    [Arguments]  ${image_name}  ${version}
-    ${ide}=  Notebook Expected Ide  ${image_name}
+    [Arguments]  ${image}  ${version}
+    ${ide}=  Notebook Expected Ide  ${image}
```
bdattoma, line 292: you're just changing variable name, what is this PR fixing?

jstourac:
> I'd expect variable `image` to hold something like `"quay.io/modh/notebooks/...@sha256-..."`, so maybe image_name is a slightly better name. But up to you, I guess there are reasons I'm not seeing.

I agree, but everywhere else there is used just `image`, thus I'm making this cosistent.

ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
jstourac: Few lines below, there was just `image` used in the method, which didn't work with the current state. My first intention was to rename that variable to `image_name` also, but then I realised that everywhere else we use just `image`...

bdattoma: Ah, it's hidden bcos unchanged. Ok thanks

bdattoma (approved):

PR #1056, opened 2023-11-30
Always describe clusters with ID

apodhrad:

PR #1229, opened 2024-02-26
fix to create user or idp by cluster id

kobihk:
when we tried to create idp or user we got the following error:

11:34:16  ods-ci-logger: INFO     CMD: ocm --v=0 create idp -c mods-master -t htpasswd -n htpasswd-cluster-admin --username htpasswd-cluster-admin-user --password ****
11:34:16  Error: Failed to get cluster 'mods-master': There are 2 subscriptions with cluster identifier or name 'mods-master'
11:34:16  
11:34:16  ods-ci-logger: INFO     CMD: ocm --v=0 create user htpasswd-cluster-admin-user --cluster mods-master --group=cluster-admins
11:34:16  Error: Failed to get cluster 'mods-master': There are 2 subscriptions with cluster identifier or name 'mods-master'


nehachopra27:
Thanks @kobihk for the PR. Could you do a test run on Jenkins with these changes before merging. LGTM otherwise

jiridanek (dismissed):

jstourac:
LGTM, just for the reference - note that this is technically a followup of #1056.

There are still some linter errors, probably from the time we didn't enforce this. Could you please update too? In a separate commit would be the best so that we don't mix these, but depends on you.

kobihk:
> LGTM, just for the reference - note that this is technically a followup of #1056.
> 
> There are still some linter errors, probably from the time we didn't enforce this. Could you please update too? In a separate commit would be the best so that we don't mix these, but depends on you.

I fixed it already how do I retrigger the linter test?

ods_ci/utils/scripts/ocm/ocm.py
```diff
@@ -1331,9 +1323,7 @@ def get_all_osd_versions(self):
             ).split("-")[0]
             if version:
                 osd_versions_dict[".".join(version.split(".")[:2])] = version
-                latest_osd_versions_data[
-                    str(self.osd_major_version)
-                ] = osd_versions_dict
+                latest_osd_versions_data[str(self.osd_major_version)] = osd_versions_dict
```
jiridanek, line 1326:
```suggestion
                latest_osd_versions_data[str(self.osd_major_version)] = (
                    osd_versions_dict
                )
```

this is what the linter wants to see

jstourac:
> I fixed it already how do I retrigger the linter test?

It is re-triggerred automatically once you push the changes. Problem is what Jiri shared above.

ods_ci/utils/scripts/ocm/ocm.py
jiridanek: it's a stupid change; we should reconfigure black to not do things like ^^^; allow longer lines?

jstourac: we possibly can; not sure what is the default right now, my guess would be 80;  ... 120 characters seem like a normal length these days; not sure whether we want more...

manosnoam:
@kobihk Can you please elaborate what is the fix ?
I can see mainly changes in the format of strings, but I don't see what is the flag/value for ocm cli that was changed ?

jstourac (approved):

manosnoam (changes requested):
Just a request to update PR description with the required fix (see my comment), and also mention it's a followup of https://github.com/red-hat-data-services/ods-ci/pull/1056 as Jan suggested.

jiridanek (approved):

manosnoam (approved):

PR #1235, opened 2024-02-29
Stabilize Dashboard Test Suite (415)

manosnoam:
Fix known Automation bugs and general stability of Dashboard Suite: 
415__ods_dashboard_projects.robot.

It also includes fixes to the Storages, Workbenches and Projects resources that are being used by this suite:
- `Create PersistentVolume Storage` option to update existing PVC
- Fix supported mount name to lower case and dashes only
- `Workbench Should Be Listed` with timeout option
- Fix `Select Existing Data Connection` in Workbench
- `Wait for RHODS Dashboard to Load` with retry logic to reload cards
- `Wait Until Project Is Open` first waits for spinner ball to disappear
- `Open Workbench` waits for workbench link, before clicking it
- Increase timeout waiting to switch to/from JupyterLab console
- New Storage KW: `Delete All PVC In Project From CLI`
- New Common KW: `Get All Text Under Element`
- New Common KW: `Get All Strings That Contain`
- Removed All Tags of Product/Automation Bugs (All Suite Tests Passed)
- Append to Suite 415: `Test Tags   Dashboard`

All 23 Test cases (including Smoke, Sanity and Tier1) passed:

image

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
@@ -32,6 +32,7 @@ ${IMAGE_XP_OLD}=  ${HEADER_XP}/img[contains(@class, 'odh-card__header-brand')]
 ${APPS_DICT_PATH_LATEST}=   ods_ci/tests/Resources/Files/AppsInfoDictionary_latest.json
 ${SIDEBAR_TEXT_CONTAINER_XP}=  //div[contains(@class,'odh-markdown-view')]
 ${SUCCESS_MSG_XP}=  //div[@class='pf-v5-c-alert pf-m-success']
+${PAGE_TITLE_XP}=  //*[@data-testid="app-page-title" and text()="Cluster settings"]
 ${USAGE_DATA_COLLECTION_XP}=    //*[@id="usage-data-checkbox"]
 ${CUSTOM_IMAGE_SOFTWARE_TABLE}=  //caption[contains(., "the advertised software")]/../tbody
 ${CUSTOM_IMAGE_PACKAGE_TABLE}=  //caption[contains(., "the advertised packages")]/../tbody
@@ -256,7 +257,7 @@ Load Expected Data Of RHODS Explore Section
 
 Wait Until Cards Are Loaded
     [Documentation]    Waits until the Application cards are displayed in the page
-    Run Keyword And Continue On Failure    Wait Until Page Contains Element
+    Run Keyword And Ignore Error    Wait Until Page Contains Element
     ...    xpath:${CARDS_XP}    timeout=15s    error="This might be caused by bug RHOAIENG-404"
 
 Get App ID From Card
@@ -508,7 +509,7 @@ Verify Cluster Settings Is Available
     Page Should Contain    Settings
     Menu.Navigate To Page    Settings    Cluster settings
     Capture Page Screenshot
-    Wait Until Page Contains    Update global settings for all users    timeout=30
+    Wait Until Page Contains Element    ${PAGE_TITLE_XP}    timeout=30
     Wait Until Page Contains Element    ${USAGE_DATA_COLLECTION_XP}    timeout=30
 
 Verify Cluster Settings Is Not Available
@@ -755,9 +756,9 @@ Get Dashboard Pods Names
 Get Dashboard Pod Logs
     [Documentation]     Fetches the logs from one dashboard pod
     [Arguments]     ${pod_name}
-    ${pod_logs}=            Oc Get Pod Logs  name=${pod_name}  namespace=${APPLICATIONS_NAMESPACE}  container=rhods-dashboard
-    ${pod_logs_lines}=      Split String    string=${pod_logs}  separator=\n
-    ${n_lines}=     Get Length    ${pod_logs_lines}
+    ${pod_logs}=    Oc Get Pod Logs  name=${pod_name}  namespace=${APPLICATIONS_NAMESPACE}  container=rhods-dashboard
+    ${pod_logs_lines}=    Split String    string=${pod_logs}  separator=\n
+    ${n_lines}=    Get Length    ${pod_logs_lines}
     Log     ${pod_logs_lines}[${n_lines-3}:]
     IF   "${pod_logs_lines}[${n_lines-1}]" == "${EMPTY}"
         Remove From List    ${pod_logs_lines}   ${n_lines-1}
@@ -768,8 +769,10 @@ Get Dashboard Pod Logs
 Get ConfigMaps For RHODS Groups Configuration
     [Documentation]     Returns a dictionary containing "rhods-group-config" and "groups-config"
     ...                 ConfigMaps
-    ${rgc_status}   ${rgc_yaml}=     Run Keyword And Ignore Error     OpenShiftLibrary.Oc Get    kind=ConfigMap  name=${RHODS_GROUPS_CONFIG_CM}   namespace=${APPLICATIONS_NAMESPACE}
-    ${gc_status}   ${gc_yaml}=      Run Keyword And Ignore Error     OpenShiftLibrary.Oc Get    kind=ConfigMap  name=${GROUPS_CONFIG_CM}   namespace=${APPLICATIONS_NAMESPACE}
+    ${rgc_status}   ${rgc_yaml}=     Run Keyword And Ignore Error     OpenShiftLibrary.Oc Get
+    ...    kind=ConfigMap  name=${RHODS_GROUPS_CONFIG_CM}   namespace=${APPLICATIONS_NAMESPACE}
+    ${gc_status}   ${gc_yaml}=      Run Keyword And Ignore Error     OpenShiftLibrary.Oc Get
+    ...    kind=ConfigMap  name=${GROUPS_CONFIG_CM}   namespace=${APPLICATIONS_NAMESPACE}
     IF   $rgc_status == 'FAIL'
         ${rgc_yaml}=    Create List   ${EMPTY}
     END
@@ -783,7 +786,7 @@ Get ConfigMaps For RHODS Groups Configuration
 Get Links From Switcher
     [Documentation]    Returns the OpenShift Console and OpenShift Cluster Manager Link
     ${list_of_links} =    Create List
-    ${link_elements}=    Get WebElements    //a[@class="pf-m-external pf-v5-c-app-launcher__menu-item" and not(starts-with(@href, '#'))]
+    ${link_elements} =    Get WebElements    //a[@class="pf-m-external pf-v5-c-app-launcher__menu-item" and not(starts-with(@href, '#'))]
     FOR    ${ext_link}    IN    @{link_elements}
         ${href}=    Get Element Attribute    ${ext_link}    href
         Append To List    ${list_of_links}    ${href}
@@ -798,16 +801,15 @@ Maybe Wait For Dashboard Loading Spinner Page
     [Documentation]     Detecs the loading symbol (spinner) and wait for it to disappear.
     ...                 If the spinner does not appear, the keyword ignores the error.
     [Arguments]    ${timeout-pre}=3s    ${timeout}=5s
-
     ${do not wait for spinner}=    Get Variable Value    ${ODH_DASHBOARD_DO_NOT_WAIT_FOR_SPINNER_PAGE}  # defaults to None if undefined
     IF   ${do not wait for spinner} == ${true}
       RETURN
     END
-
+    ${spinner_ball} =   Set Variable    xpath=//span[@class="pf-v5-c-spinner__tail-ball"]
     Run Keyword And Ignore Error    Run Keywords
-    ...    Wait Until Page Contains Element    xpath=//span[@class="pf-v5-c-spinner__tail-ball"]    timeout=${timeout-pre}
+    ...    Wait Until Page Contains Element    ${spinner_ball}    timeout=${timeout-pre}
     ...    AND
-    ...    Wait Until Page Does Not Contain Element    xpath=//span[@class="pf-v5-c-spinner__tail-ball"]    timeout=${timeout}
+    ...    Wait Until Page Does Not Contain Element    ${spinner_ball}    timeout=${timeout}
 
 Reload RHODS Dashboard Page
     [Documentation]    Reload the web page and wait for RHODS Dashboard
@@ -822,7 +824,7 @@ Handle Deletion Confirmation Modal
     [Arguments]     ${item_title}    ${item_type}   ${press_cancel}=${FALSE}    ${additional_msg}=${NONE}
     # Once fixed https://issues.redhat.com/browse/RHODS-9730 change the button xpath to
     # xpath=//button[text()="Delete ${item_type}"]
-    ${delete_btn_xp}    Set Variable    xpath=//button[contains(text(), 'Delete')]
+    ${delete_btn_xp} =    Set Variable    xpath=//button[contains(text(), 'Delete')]
     Wait Until Generic Modal Appears
     Run Keyword And Warn On Failure    Page Should Contain    Delete ${item_type}?
     Run Keyword And Warn On Failure    Page Should Contain    This action cannot be undone.
@@ -845,7 +847,7 @@ Click Action From Actions Menu
     [Arguments]    ${item_title}    ${action}    ${item_type}=${NONE}
     Click Element       xpath=//tr[td[@data-label="Name"]//*[text()="${item_title}"]]/td[contains(@class,"-table__action")]//button[@aria-label="Kebab toggle"]    # robocop: disable
     IF    "${item_type}" != "${NONE}"
-        ${action}=    Catenate    ${action}    ${item_type}
+        ${action} =    Catenate    ${action}    ${item_type}
     END
     Wait Until Page Contains Element       xpath=//tr[td[@data-label="Name"]//*[text()="${item_title}"]]//td//li//*[text()="${action}"]    # robocop: disable
     Click Element       xpath=//tr[td[@data-label="Name"]//*[text()="${item_title}"]]//td//li//*[text()="${action}"]
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
@@ -73,11 +73,11 @@ Login To RHODS Dashboard
    ${expected_text_list}=    Create List    Log in with    Data Science Projects
    Wait Until Page Contains A String In List    ${expected_text_list}
 
-   ${oauth_prompt_visible} =  Is OpenShift OAuth Login Prompt Visible
+   ${oauth_prompt_visible}=  Is OpenShift OAuth Login Prompt Visible
    IF  ${oauth_prompt_visible}  Click Button  Log in with OpenShift
-   ${login-required} =  Is OpenShift Login Visible
+   ${login-required}=  Is OpenShift Login Visible
    IF  ${login-required}  Login To Openshift  ${ocp_user_name}  ${ocp_user_pw}  ${ocp_user_auth_type}
-   ${authorize_service_account} =  Is rhods-dashboard Service Account Authorization Required
+   ${authorize_service_account}=  Is rhods-dashboard Service Account Authorization Required
    IF  ${authorize_service_account}  Authorize rhods-dashboard service account
 
 Logout From RHODS Dashboard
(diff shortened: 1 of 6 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -256,7 +257,7 @@ Load Expected Data Of RHODS Explore Section
 
 Wait Until Cards Are Loaded
     [Documentation]    Waits until the Application cards are displayed in the page
-    Run Keyword And Continue On Failure    Wait Until Page Contains Element
+    Run Keyword And Ignore Error    Wait Until Page Contains Element
```
bdattoma, line 260: If we ignore the Error we might not see issues with the Dashboard page, cards are supposed to be loaded.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
```diff
@@ -424,4 +418,4 @@ Clean Project From Workbench Resources
     ...    ${pvc_title}=${workbench_title}
     Delete Workbench From CLI    workbench_title=${workbench_title}
     ...    project_title=${project_title}
-    Delete PVC From CLI    pvc_title=${pvc_title}    project_title=${project_title}
+    Delete PVC In Project From CLI    pvc_title=${pvc_title}    project_title=${project_title}
```
bdattoma, line 421: was this addition in the kw name really necessary?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
```diff
@@ -54,10 +54,11 @@ Storage Size Should Be
     IF    "${pvc_status_phase}" == "Pending"
         # PVC hasn't been bound yet - only simple text with max storage should be shown
         ${storage_size_el}=    Set Variable
-        ...    ${STORAGE_SECTION_XP}//tr[@class="pf-c-table__expandable-row pf-m-expanded"]/td/div[strong[text()="Size"]]/div    # robocop: disable
+        ...    ${STORAGE_SECTION_XP}//td[contains(@data-label,"Size")]//div[strong[text()="Size"]]/div    # robocop: disable
         Wait Until Page Contains Element    ${storage_size_el}    timeout=20s
-        ${displayed_size}=    Get Text    ${storage_size_el}
-        Run Keyword And Continue On Failure    Should Be Equal As Strings    ${displayed_size}    Max ${size}Gi
+        ${displayed_sizes}=    Get All Text Under Element    ${storage_size_el}
+        ${matched_sizes}=    Get All Strings That Contain    ${displayed_sizes}    ${size}Gi
```
bdattoma, line 60: why are you removing "Max"? when the PVC is not bounded to a workload it shows Max ${size}Gi, at least it used to do so last time I checked

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
```diff
@@ -66,53 +67,65 @@ Storage Size Should Be
...
         ...    scenario. We expect either 'Pending' or 'Bound' states here. Please, either complete the
         ...    implementation or investigate the test failure.
         Fail    ${error_msg}
         RETURN    1
     END
-    # This check is common for both "Pending" and "Bound" states of PVC.
+    # This check is common for both "Pending" and "Bound" states of PVC, and it verifies storage size directly via OC CLI
     ${rc}  ${oc_object_size}=    Run And Return Rc And Output
-    ...    oc get pvc -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${name}")].spec.resources.requests.storage}'    # robocop: disable
+    ...    oc get pvc -n ${namespace} -o jsonpath="{.items[?(@.metadata.annotations['openshift\\.io/display-name']=='${name}')].spec.resources.requests.storage}"    # robocop: disable
```
bdattoma, line 89: any particular reason for this change?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
```diff
@@ -66,53 +67,65 @@ Storage Size Should Be
...
     Should Be Equal As Integers    ${rc}    0
     ...    An error occurred during the check of PVC ${name} '.spec.resources.requests.storage' value!
     Run Keyword And Continue On Failure     Should Be Equal As Strings    ${oc_object_size}    ${size}Gi
 
 Create PersistentVolume Storage
     [Documentation]    Create a PersistenVolume storage in DS Project details page
-    [Arguments]    ${project_title}    ${name}    ${description}    ${size}    ${connected_workbench}=${NONE}    ${press_cancel}=${FALSE}
-    Click Button    ${STORAGE_ADD_BTN_1_XP}
+    [Arguments]    ${project_title}    ${name}    ${description}    ${size}    ${connected_workbench}=${NONE}
+    ...    ${press_cancel}=${FALSE}    ${existing_storage}=${FALSE}
```
bdattoma, line 97: what's the benefit of re-using a pvc when calling the kw to create one?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
```diff
@@ -151,13 +166,35 @@ Delete Storage
...
+
+Delete All PVC In Project From CLI
+    [Documentation]    Deletes All PVCs in a project from CLI
+    [Arguments]    ${project_title}
+    ${ns_name}=    Get Openshift Namespace From Data Science Project   project_title=${project_title}
+    ${result}=    Run Process
+    ...    oc delete pvc -n ${ns_name} --all    shell=yes
+    Should Be True    ${result.rc} == 0    msg=${result.stderr}
+
+Genereta Mount Name
```
bdattoma, line 195: typo "Generate"

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
```diff
@@ -356,8 +354,11 @@ Launch And Access Workbench
 Open Workbench
     [Documentation]    Clicks on "open" link for a given workbench
     [Arguments]    ${workbench_title}
-    Click Link       ${WORKBENCH_SECTION_XP}//tr[td[@data-label="Name"]/*[div[text()="${workbench_title}"]]]/td//a[text()="Open"]
-    Switch Window   NEW
+    ${open_workbench_link}=    Set Variable
+    ...    ${WORKBENCH_SECTION_XP}//tr[td[@data-label="Name"]/*[div[text()="${workbench_title}"]]]/td//a[text()="Open"]
+    Wait Until Page Contains Element    ${open_workbench_link}    timeout=30s
```
bdattoma, line 359: this may hide some prod issue like workbench Running but not reporting the link to open. This was something captured in PSAP test runs

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
manosnoam: But apparently `Run Keyword And Continue On Failure` blocks Test Setup (and everything afterwards).
Shouldn't we ignore it (RHOAIENG-404) and just warn.
It will fail later if there's anything other than  RHOAIENG-404. 
We can also refresh if you think that is a better approach ?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
```diff
@@ -422,19 +423,21 @@ Check Launched Workbench Is The Correct One
 Get Openshift Notebook CR From Workbench
     [Documentation]    Retrieves name of Notebook CR corresponding to a workbench
     [Arguments]     ${workbench_title}      ${namespace}
-    ${rc}  ${cr_name}=    Run And Return Rc And Output   oc get notebook -n ${namespace} -o jsonpath='{.items[?(@.metadata.annotations.openshift\\.io/display-name=="${workbench_title}")].metadata.name}'
-    RETURN    ${rc}    ${cr_name}
+    ${result}=    Run Process
```
bdattoma, line 426: why? was it causing issues?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
```diff
@@ -422,19 +423,21 @@ Check Launched Workbench Is The Correct One
...
+    ...    oc get notebook -n ${namespace} -o jsonpath\='{.items[?(@.metadata.annotations.openshift\\.io/display-name\=\="${workbench_title}")].metadata.name}'
+    ...    shell=yes
+    RETURN    ${result.rc}    ${result.stdout}
 
 Start Workbench Should Fail
     [Documentation]    Checks the workbench fails to start
-    [Arguments]    ${workbench_title}
+    [Arguments]    ${prj_title}    ${workbench_title}
+    Wait Until Project Is Open    project_title=${prj_title}
```
bdattoma, line 434: why? this should be already open when this kw is supposed to be used

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
```diff
@@ -121,8 +118,8 @@ Verify User Can Create And Start A Workbench With Existent PV Storage
     [Documentation]    Verifies users can create a workbench and connect an existent PersistenVolume
     ${pv_name}=    Set Variable    ${PV_BASENAME}-existent
     Open Data Science Project Details Page       project_title=${PRJ_TITLE}
-    Create PersistentVolume Storage    name=${pv_name}    description=${PV_DESCRIPTION}
-    ...                               size=${PV_SIZE}    connected_workbench=${NONE}    project_title=${PRJ_TITLE}
+    Create PersistentVolume Storage    name=${pv_name}    description=${PV_DESCRIPTION}    project_title=${PRJ_TITLE}
+    ...                               size=${PV_SIZE}    connected_workbench=${NONE}   existing_storage=${TRUE}
```
bdattoma, line 122: why would we have a PVC with the same name already created in the project? 
I think we should keep the test the most predictable we can.

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
```diff
@@ -136,23 +133,30 @@ Verify User Can Create And Start A Workbench With Existent PV Storage
...
     ${pv_name}=    Set Variable    ${PV_BASENAME}-A
     ${ns_name}=    Get Openshift Namespace From Data Science Project   project_title=${PRJ_TITLE}
     Open Data Science Project Details Page       project_title=${PRJ_TITLE}
-    ${workbenches}=    Create Dictionary    ${WORKBENCH_TITLE}=mount-data
-    Create PersistentVolume Storage    name=${pv_name}    description=${PV_DESCRIPTION}
-    ...                               size=${PV_SIZE}    connected_workbench=${NONE}     press_cancel=${TRUE}
-    ...                               project_title=${PRJ_TITLE}
-    Create PersistentVolume Storage    name=${pv_name}    description=${PV_DESCRIPTION}
-    ...                               size=${PV_SIZE}    connected_workbench=${workbenches}   project_title=${PRJ_TITLE}
+    Create Workbench    workbench_title=${WORKBENCH_TITLE}  workbench_description=${WORKBENCH_DESCRIPTION}
```
bdattoma, line 142: what it the benefit of creating the workbench here? 
Unless you check the workbench gets restarted when the PVC gets created and attached, it is not needed

bdattoma, line 142: could you add the check to ensure the workbench gets restarted after the PVC is connected?

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
```diff
@@ -219,7 +223,7 @@ Verify User Can Create A S3 Data Connection And Connect It To Workbenches
     ...                          aws_s3_endpoint=${DC_S3_ENDPOINT}    aws_region=${DC_S3_REGION}
     ...                          connected_workbench=${workbenches}
     Data Connection Should Be Listed    name=${DC_2_S3_NAME}    type=${DC_S3_TYPE}    connected_workbench=${workbenches}
-    Run Keyword And Continue On Failure    Workbench Status Should Be     workbench_title=${WORKBENCH_TITLE}
+    Run Keyword And Ignore Error    Wait Until Workbench Is Started     workbench_title=${WORKBENCH_TITLE}
```
bdattoma, line 226: the old version was checking that the workbench is in "Starting" status. The new check is useless especially cos it's duplicated with the very next check

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
bdattoma: if it's "Continue on Failure" it should not block test setup and everything afterwards, unless the reason for the subsequent failures is the same which caused the first failure. no?
If you think the waiting is not needed in a particular test case or step, you could set `wait_for_cards=${FALSE}` in the `Wait for RHODS Dashboard to Load` kw where `Wait Until Cards Are Loaded` is used.

I personally would not ignore the failure. The warn can easily get unseen in this case since the kw is mainly used nested in others

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
manosnoam: I will remove the whole line, since as you said we're waiting for the workbench to start anyway on line 226.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
bdattoma: ah ok that's normal, true. I thought it didn't continue other test setup keywords, sorry.

Anyways, instead of ignoring always, I would rather skip the wait with the existing parameter if needed. Ignoring this error doesn't sound good to me.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
bdattoma: that should be handled by the IF/ELSE already present. iirc @jstourac worked on it. 

Max appears when the PVC is not bounded

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
bdattoma: hm okay

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
bdattoma: imo it makes the tests less predictable and may hide other issues (maybe test env issues). Unless we explicitly want the PVC to be already present for so efficiency reasons of the tests maybe. In this particular case, why the PVC would be already present before?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
bdattoma: I was referring to issues like https://issues.redhat.com/browse/RHODS-5819 where the link wasn't ready after the status of workbench was Running. So i'm wondering if adding this additional ait may hide similar issues

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
bdattoma: you could check if ${rc} is different from 0 then it fails. I don't see an actual need of changing it, but ok if you have a stronger opinion.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
bdattoma: >It caused instability in the (only single) test that used it


meaning that the project wasn't open? 

> verifying that project within the KW


it's redudant, in order to run this kw you already ran the kw to open the project which already waits for the project to be opened.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
manosnoam: It's an improvement of this Keyword, like we have with other keywords. 
 This opens an option for more test logic where we can edit existing PVC, if it exists (and create if it is not).
Search in code for `${existing_` and you'll other keywords that uses that logic.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
manosnoam: > meaning that the project wasn't open?

Yes.
But I can move the logic outside the KW, which means that we will need to verify the project of the workbench in each test, before using that KW.

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
manosnoam: Yes it caused issues, since the projects could have existed before.
I also added a cleanup step on Test Suite Setup, for an additional safety.
However, `existing_storage` is an optional test feature, which allows us:
1) To test UI in more ways 
2) Without facing Automation bugs if test logic did not cleanup previous PVC
3) While not not introducing any risk of "hidden bugs"

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
manosnoam: Yes, because sometimes more time passed between the workbench creation until this line, where the status already changed to running, and thus it failed (false failure).

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
manosnoam: It happened few times when that kw was called, for example look at `Delete Workbench From CLI`:
```
${_}  ${cr_name}=    Get Openshift Notebook CR From Workbench
    ...    workbench_title=${workbench_title}  namespace=${ns_name}
IF    "${cr_name}" == "${EMPTY}"    BREAK
```
In the old way, when `Get Openshift Notebook CR From Workbench` failed, then `${cr_name}` was not empty, but included the stderr, which then did not lead to the `BREAK` of the WHILE. 
Using `Run Process` in `Get Openshift Notebook CR From Workbench`, and any other KW out there would not lead to such failure, since we return the stdout only, and in our case empty value, which leads to the BREAK as expected.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
manosnoam: I've implemented option (3) so it retries to reload page up to 3 times before giving up

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Storages.resource
manosnoam: I restored the Max check (it required xpath fix)

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Workbenches.resource
manosnoam: I restored to the original logic, and removed `Wait Until Project Is Open`

manosnoam:
@bdattoma  I fixed/answered all of your suggestions, and rerun the whole suite (including sanity and tier1) + smoke of rhods-ci-pr-test/2537 - All tests passed!

image

jstourac (approved):
Thank you, Noam. I had a very brief look and LGTM.

jiridanek (approved):

FedeAlonso (approved):

PR #1238, opened 2024-02-29
increase the delete cluster timeout

kobihk:
A few Jenkins jobs failed because of the short timeout of the delete cluster with the following error:

20:06:28  ods-ci-logger: INFO     ocm cluster with name xxxx exists!
20:07:24  ods-ci-logger: ERROR    xxxx not deleted even after an hour. EXITING

increasing the timeout from 60 minutes to 90 minutes

related to jira ticket: RHOAIENG-3917

### DIFF
--- a/ods_ci/utils/scripts/ocm/ocm.py
+++ b/ods_ci/utils/scripts/ocm/ocm.py
@@ -937,7 +937,7 @@ class OpenshiftClusterManager:
             sys.exit(1)
         self.wait_for_osd_cluster_to_get_deleted()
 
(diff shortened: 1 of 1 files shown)


jstourac:
Are we sure that the timeout extension helps? Does the cluster deletion really can take that much time? :thinking:

kobihk:
> Are we sure that the timeout extension helps? Does the cluster deletion really can take that much time? 🤔

Please look at the Jira ticket 
you can see links to 3 jenkins jobs that failed with the same timeout error
when I list the clusters they are no longer there.
btw it was on production(not staging)

nehachopra27 (approved):
LGTM

PR #1242, opened 2024-02-29
add support for latest rosa version

kobihk:
add the support to create a new ROSA cluster with OPENSHIFT_VERSION=x.y-latest
in this case, it will retrieve the latest ROSA version and use it.
 
related to jira ticket: RHOAIENG-2683

### DIFF
--- a/ods_ci/utils/scripts/rosa/rosa.py
+++ b/ods_ci/utils/scripts/rosa/rosa.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import re
 import sys
 
 dir_path = os.path.dirname(os.path.abspath(__file__))
@@ -7,6 +8,7 @@ sys.path.append(dir_path + "/../")
 from awsOps import aws_configure
 from logger import log
 from rosaOps import create_account_roles, rosa_create_cluster, rosa_whoami, wait_for_osd_cluster_to_be_ready
+from util import execute_command
 
 
 class RosaClusterManager:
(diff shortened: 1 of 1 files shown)


ods_ci/utils/scripts/rosa/rosa.py
```diff
@@ -21,6 +23,22 @@ def __init__(self, args={}):
...
 
+    def set_rosa_version(self):
+        version_match = re.match(r"(\d+\.\d+)\-latest", self.rosa_version)
+        if version_match is None:
+            log.info(f"Using the rosa version given by user: {self.rosa_version}")
+            return
+        version = version_match.group(1)
+        channel_group = ""
+        if self.channel_name == "candidate":
+            channel_group = f"--channel-group {self.channel_name}"
```
jstourac, line 34: I don't fully understand this, TBH. Disclaimer: I have no knowledge about rosa CLI so I may be wrong.

TBH, I would rather include the `--channel-group` based on whether a `channel_name` is defined or not rather than including based on the type of the channel name here. Does that make sense? Because, with this implementation, it means that only in case that `candidate` channel name is used, this argument will be given to the command. What if this channel will be renamed in the future? What is the reason not to add this argument to the command when the channel name is different?

ods_ci/utils/scripts/rosa/rosa.py
```diff
@@ -21,6 +23,22 @@ def __init__(self, args={}):
         self.rosa_version = args.get("rosa_version")
         self.channel_name = args.get("channel_name")
 
+    def set_rosa_version(self):
+        version_match = re.match(r"(\d+\.\d+)\-latest", self.rosa_version)
+        if version_match is None:
+            log.info(f"Using the rosa version given by user: {self.rosa_version}")
+            return
```
jstourac, line 30: Maybe we can include some info message also for the `else` branch? Something like:
```
User provided {self.rosa_version} OpenShift version, trying to determine appropriate latest version for the ROSA cluster creation.
```

ods_ci/utils/scripts/rosa/rosa.py
kobihk: you are right, we can add --channel-group  in any case 
Done!!

jstourac (approved):
lgtm, with my limited knowledge!

ods_ci/utils/scripts/rosa/rosa.py
```diff
@@ -21,6 +23,21 @@ def __init__(self, args={}):
...
+        log.info(f"User provided {self.rosa_version}, trying to determine the appropriate latest version for ROSA")
+        version = version_match.group(1)
+        latest_version_cmd = (
+            f"rosa list versions --channel-group {self.channel_name} | "
+            f"awk '{{print $1}}' | grep -w '^{re.escape(version)}*' | head -n1"
+        )
```
jiridanek, line 36: I'd run `rosa list versions --channel-group {self.channel_name}` as subprocess and then I'd process the output in Python, not with awk, grep, and head

jiridanek (approved):

ods_ci/utils/scripts/rosa/rosa.py
kobihk: In case I need the command output for more usages I can agree  
but in this case, I like it in "one line" of code.
Thanks @jiridanek

PR #1243, opened 2024-03-01
Improve `utils.execute_command`, add a Pytest selftest for it, and run it in GHA

### LINKED ISSUE #1241: Various problems with `util.execute_command`
I have to admit that I know almost nothing about Python, but from what I tired, the while loop isn't called after the command is completed but during the command is running. Try to play e.g. with something like this:

```
import subprocess
import sys

cmd = "while true; do echo 'ahoj'; sleep 2; done"
#cmd = "echo ahoj; sleep 2"
output = ""

print(f"CMD: {cmd}")

with subprocess.Popen(
    cmd,
    shell=True,
    stdout=subprocess.PIPE,
    stderr=subprocess.STDOUT,
    universal_newlines=True,
    encoding="utf-8",
    errors="replace",
) as p:
    while True:
        line = p.stdout.readline()
        if line != "":
            output += line + "\n"
           # print(line)
        elif p.poll() is not None:
            break
    sys.stdout.flush()
    print(f"OUTPUT: {output}")
```

Thus as I said - we don't have to face any issue for a short running calls here, but we may eventually run into issue of missing a long-running command execution that has been terminated before the command ended and could print out the final `output`. Also, with this change we don't have any idea what is happening in the meantime for these long-running commands.

_Originally posted by @jstourac in https://github.com/red-hat-data-services/ods-ci/pull/1234#discussion_r1507777908_

jiridanek:
Pytest is configured according to https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-as-part-of-application-code

I am using the same `poetry` command for both the ruff and pytest jobs, so that they can share cache.

Fixes #1241


ods_ci/utils/scripts/util.py
```diff
@@ -8,7 +8,7 @@
 
 import jinja2
 import yaml
-from logger import log
+from utils.scripts.logger import log
```
jiridanek, line 11: hmmm....

ods_ci/utils/scripts/util.py
```diff
@@ -53,13 +53,12 @@ def read_yaml(filename):
         return None
 
 
-def execute_command(cmd):
+def execute_command(cmd, print_stdout=True):
```
jiridanek, line 56: this is to have an option to quiet down the output in selftests

ods_ci/utils/scripts/util.py
jiridanek: yes, that was a mistake, fixed now

jiridanek:
Run this twice, there was only a test fail in "Verify that the must-gather image provides RHODS logs and info", which is known open issue, https://redhat-internal.slack.com/archives/C03V3J222D7/p1707985201274959?thread_ts=1707974794.266549&cid=C03V3J222D7

CI: job/rhods/job/rhods-ci-pr-test/2544/

ods_ci/selftests/utils/scripts/test_util.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_stderr(self):
+        assert util.execute_command("echo stderr >&2") == "stderr\n"
+
+    def test_string_cmd(self):
+        assert util.execute_command("echo hello world", print_stdout=False) == "hello world\n"
+
+    def test_list_cmd(self):
+        # this is surprising, but it's what subprocess.Popen does
+        assert util.execute_command(["echo", "hello", "world"], print_stdout=False) == "\n"
```
jstourac, line 78:
```suggestion
    def test_list_cmd(self):
        assert util.execute_command(["echo hello world"], print_stdout=False) == "hello world\n"

    def test_list_cmd(self):
        # this is surprising, but it's what subprocess.Popen does
        assert util.execute_command(["echo", "hello", "world"], print_stdout=False) == "\n"
```

ods_ci/utils/scripts/util.py
```diff
@@ -69,17 +68,16 @@ def execute_command(cmd):
...
-        return None
+            output = []
+            for line in p.stdout:
+                output.append(line)
+                if print_stdout:
+                    print(">:", line, end="")
+                    sys.stdout.flush()
+            return "".join(output)
+    except Exception as e:
+        log.exception("Starting the subprocess failed", exc_info=e)
```
jstourac, line 79: Tentative (optional):

```suggestion
        log.exception(f"Starting the subprocess '{cmd}' failed", exc_info=e)
```

Reasoning: I know that we print the command right before this try block. I just think it's nice to have the command also as part of the exception message to be completely clear on what command was about to be executed as a subprocess. E.g. in case that there is some other output in the log in the meantime for any reason.

jstourac (approved):
With my limited knowledge, LGTM. I put just two optional changes.

It's really funny that all I wanted was to revert 2 lines of code, and here you've come with this PR :smile:

Thank you.

ods_ci/utils/scripts/util.py
jiridanek: If I recall correctly, the command that was executed is also logged in the exception data, but I'm not certain this is the case. No harm printing it twice, in the worst case.

jstourac (approved):

ods_ci/utils/scripts/util.py
```diff
@@ -69,17 +68,16 @@ def execute_command(cmd):
...
-            return output
-    except:
-        return None
+            output = []
+            for line in p.stdout:
+                output.append(line)
+                if print_stdout:
+                    print(">:", line, end="")
+                    sys.stdout.flush()
+            return "".join(output)
```
FedeAlonso, line 77: Why are you storing the lines in an array if then you are going to join it in a string? Is not better to store 'em directly in a string?

jstourac: Execution efficiency :slightly_smiling_face:  https://realpython.com/python-string-concatenation/#efficiently-concatenating-many-strings-with-join-in-python.

FedeAlonso: > The .join() method is cleaner, more Pythonic, and more readable than concatenating many strings together in a loop using the augmented concatenation operator (+=), as you saw before.
You learn something new every day.
Thanks for the tip @jstourac !

FedeAlonso (approved):

ods_ci/utils/scripts/util.py
jstourac: Main point for the efficiency here is this note, though :slightly_smiling_face: 

> Unlike the concatenation operators, Python’s .join() doesn’t create new intermediate strings in each iteration. Instead, it creates a single new string object by joining the elements from the input iterable with the selected separator string. This behavior is more efficient than using the regular concatenation operators in a loop.

PR #1257, opened 2024-03-06
fix: New xpath on model serving runtime selection dropdown

FedeAlonso:
The xpath for the Model Serving selection dropdown has changed in the 2.8

Fix for tests:
- ODS-2276: Verify RHODS Admins Can Import A Custom Serving Runtime Template By Uploading A YAML file
- ODS-2279: Verify RHODS Admins Can Delete A Custom Serving Runtime Template

image


jstourac (approved):

bdattoma:
doesn't the UI requires to select the protocol as well?

FedeAlonso:
> doesn't the UI requires to select the protocol as well?

Not on this tests

PR #1266, opened 2024-03-07
Add query to ModelInfo and Tokenize endpoints

bdattoma:
Adding query and response validation against ModelInfo and Tokenize endpoints for all the models in `ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/422__model_serving_llm_models.robot` and `flan-t5-small` in `ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/422__model_serving_llm_models.robot`.

Besides, this PR wants to start rewriting the logic for response validation which is becoming too complex and not scalable on the differences btw runtimes.
It is not affecting Generate and GenerateStream endpoint validation, they will be handled in another PR later.


tarukumar (approved):

bdattoma:
PR validation (full model suite): `rhods-smoke/4799/`

PR #1298, opened 2024-03-14
fix: add productbug tag to ODS-1969

FedeAlonso:
Add productbug tag to ODS-1969 so it does not run in regular pipelines


jstourac (approved):

asanzgom (commented):
~Can you please add on [Documentation] the JIRA of the Product Bug?~

asanzgom (approved):

PR #1301, opened 2024-03-15
Fix deletion confirmation message change in RHOAI 2.9.0 Dashboard

jiridanek:
SeleniumLibrary . Page Should Contain Type ${item_title} to confirm deletion.


jstourac:
This is supposed to be against 2.9 release. Please don't merge until we decide whether we'll branch 2.8 already.

jstourac (approved):

jiridanek:
@jstourac I set Milestone to 2.9.0. Is it still appropriate to have the PR as draft, when the changes in it are actually complete? I understand it is a practical safety measure, but still it feels wrong to me.

jstourac:
Okay, obviously different understanding. Personally I don't mind having PRs as drafts even if they are complete since it's safer than the `do-not-merge` label. But feel free to override me, I believe I may be just sole in this opinion.

kobihk (approved):

PR #1302, opened 2024-03-15
[RHOAIENG-3757] Kueue: Reduce time of setting up of e2e tests

sutaakar:
As a result the local execution of Kueue tests no longer require `kconfig` file, you just need to be logged into the cluster (local Kubeconfig is authenticated, same as for other tests).

### DIFF
--- a/ods_ci/tests/Tests/650__distributed_workloads/test-run-kueue-e2e-tests.robot
+++ b/ods_ci/tests/Tests/650__distributed_workloads/test-run-kueue-e2e-tests.robot
@@ -9,13 +9,9 @@ Resource          ../../../tasks/Resources/RHODS_OLM/install/oc_install.robot
 
 
 *** Variables ***
-${KUEUE_DIR}            kueue
-${KUEUE_REPO_URL}       %{KUEUE_REPO_URL=https://github.com/opendatahub-io/kueue.git}
-${KUEUE_REPO_BRANCH}    %{KUEUE_REPO_BRANCH=dev}
-${JOB_GO_BIN}           %{WORKSPACE=.}/go-bin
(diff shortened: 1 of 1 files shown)


jiripetrlik (dismissed):

jiripetrlik (dismissed):

ods_ci/tests/Resources/Page/OCPLogin/OCPLogin.robot
```diff
@@ -15,3 +15,9 @@ Login To OCP Using API
   [Arguments]    ${username}      ${password}
   ${rc}    ${out}=    Run And Return Rc And Output    oc login $(oc whoami --show-server) -u ${username} -p ${password}  #robocop:disable
   Should Be Equal As Integers    ${rc}    ${0}
+
+Login To OCP Using API And Kubeconfig
```
jgarciao, line 19: It's possible this is not needed. Can you look at this example? https://github.com/red-hat-data-services/ods-ci/blob/1ce419d3d1dee0f4954daf9f6e5d59bf5c2d49ca/ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/436__test-run-data-science-pipelines-operator-e2e-tests.robot#L31

sutaakar: Thanks for the tip.
Makes me think, what is a meaning of `kconfig`, which is available in ods-ci in the job environment, if `.kube/config` holds the current token?

sutaakar: Implemented, feel free to review again

ChughShilpa (approved):

jiripetrlik (approved):

PR #1303, opened 2024-03-15
fix: Verify successful pipeline via UI from Pipelines->Runs view

FedeAlonso:
Modify the "Verify Successful Pipeline Run Via Project UI" keyword so now it uses Pipelines->Runs (the other way was deprecated)

It will also fix the Error with "ODS-2271 - Verify Elyra Pipelines In SDS-Based Images" We've found during the execution of Sanity in 2.8 RC3.
Test results: 
image

Also, a wait was added because during the testing phase we had issues trying to run the test, because the automation was so fast that some of the changes in the pipeline had no time to get auto saved, and when trying to run the pipeline an "error" message was shown:
Screenshot from 2024-03-15 11-32-50


jstourac (approved):
LGTM, thank you Fede!

lugi0 (approved):
lgtm

PR #1306, opened 2024-03-15
[Fix] Notebook API test

jstourac:
This test was probably failing since the addition of the accelerator profiles. Since I have no idea where is some documentation to the dashboard API which we use here, I grepped my local disk for some keywords and was able to find this file [1] from which I was able to understand what is wrong with this test. As such, please review thoroughly since I could miss something.

[1] https://github.com/opendatahub-io/odh-dashboard/blob/main/frontend/src/types.ts#L670

After merge, I shall backport to `releases/2.8.0` branch too.

---

Screenshot from 2024-03-15 19-45-42

### DIFF
--- a/ods_ci/tests/Tests/400__ods_dashboard/414__ods_dashboard_api.robot
+++ b/ods_ci/tests/Tests/400__ods_dashboard/414__ods_dashboard_api.robot
@@ -1,6 +1,8 @@
 *** Settings ***
 Documentation     Suite for a basic security test of Dashboard APIs. The tests verifies that user
-...               reach endpoints based on their user permissions
+...               reach endpoints based on their user permissions.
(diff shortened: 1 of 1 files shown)


jiridanek:
> Since I have no idea where is some documentation to the dashboard API which we use here

I'm only aware that Dashboard is mentioned in API tiers, https://access.redhat.com/articles/7047935

jiridanek (approved):

jstourac:
> I'm only aware that Dashboard is mentioned in API tiers, https://access.redhat.com/articles/7047935

Yeah, I know that one, but I had something more descriptive for use in mind. But I suppose that it's Tier 4, then it doesn't make sense to document it properly, IDK.

PR #1309, opened 2024-03-19
RHOAIENG-3425_Validate_Trusted_CA_Bundle

mattmahoneyrh:
The purpose of this new test case is to validate the operator DSCI support of Trusted CA Bundles.


mattmahoneyrh:
Lint Warnings should now be resolved.

mattmahoneyrh:
Lint Warnings should all be resolved, with the exception of line lengths which could not determine a solution without causing syntax errors.

PR #1311, opened 2024-03-20
[Draft]Add authorino operator only for odh nightly

aloganat:


jiridanek (approved):

jstourac:
I'm not sure whether we sill need and want this? :thinking:

jiridanek:
Imo yes, authorino is going to come sooner or later, and nobody removed it from upstream odh, as far as I know.

jstourac:
> Imo yes, authorino is going to come sooner or later, and nobody removed it from upstream odh, as far as I know.

Yeah, but this PR is to install it against ODH-nighly only, which was okay up to last week for RHOAI, but then plan changed, so I'm not sure whether this is still needed :slightly_smiling_face:

jstourac:
Just for the reference - Red Hat Authorino is now available on OCP4.15 - https://issues.redhat.com/browse/RHOAIENG-4710.

PR #1312, opened 2024-03-20
Migrate pipelines tests to DSPv2 (pipelines-api, pipelines-kfp)

diegolovison:
Dispite of `ray` integration tests the others looks good
```
sh ods_ci/run_robot_test.sh --set-urls-variables true --extra-robot-args '-L DEBUG' --test-case 'ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/431__data-science-pipelines-api.robot' --skip-oclogin
sh ods_ci/run_robot_test.sh --set-urls-variables true --extra-robot-args '-L DEBUG' --test-case 'ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/432__data-science-pipelines-kfp.robot' --skip-oclogin
```


jgarciao (approved):
It is possible that there are things to polish, but I think it's better to merge now and fix in follow-up PR, as we need this code for fixing other tests

Raghul-M (approved):

PR #1315, opened 2024-03-24
Remove workaround for resolved product bugs RHOAIENG-2724, RHOAIENG-2869

manosnoam:
These RHOAI bugs were recently resolved:
- https://issues.redhat.com/browse/RHOAIENG-2869
- https://issues.redhat.com/browse/RHOAIENG-2724

Thus removing the ProductBug tag and workaround.

image

PR #1296, opened 2024-03-14
Adding link to ticket bugs in some tests

bdattoma:

PR #1318, opened 2024-03-25
Backport required changes from master to releases/2.8.0

jstourac:
This PR contains following fixes from the `master` branch:

* edafbcc301fcb40a1227fb4e0c3ab05edcebe860 (#1296)
* 5e7f4b116e0c83cac3a91846cc31bddfdd934bd1 (#1302)
* b6fb7f74f99b626cb1561e85defaeb86893d0ad2 (#1307)
* 94852466b134710ca41c92cc3d77944b11287497 (#1314)
* 3e35a714e878bcce49bd62bcaff3bc5c0db79181 (#1315)
* abfb8a5d6623302040f18b5202b9aed044e5f01c (#1316)
* fc76761919c306b46ad60fecd0bcacd0e5a5e0b4 (#1317)
* 439e2b31a9612c356fe2b4670d4aca3b2504dd2a (#1306)

---

I kept the order in which these commits were merged in the `master` branch and provided both link to the particular commit and also link to the particular PR so it's clearer to understand later on.


jiridanek:
Would it make sense to create this PR through `git cherry-pick`, so that the hashes of the original commits are preserved? Looking in the github webui, I don't see the "cherry picked from commit" lines in commit messages. It is added with `-x`, https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-cherry-pick.html

Or do a merge with merge commit which (maybe) also preserves commit hashes?

There is this "Daggy Fixes and Cherry-picking" from https://queue.acm.org/detail.cfm?id=1595636 which I found from https://stackoverflow.com/questions/2922652/git-is-there-a-way-to-figure-out-where-a-commit-was-cherry-picked-from sounds interesting, btw. But it's too late to use it for the commits above, and I don't exactly understand how this is supposed to work in Git.

jstourac:
> Would it make sense to create this PR through `git cherry-pick`, so that the hashes of the original commits are preserved in commit messages? Looking in the github webui, I don't see the "cherry picked from commit" lines in commit messages. It is added with `-x`, https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-cherry-pick.html

I don't know how to do these things in UI so yeah - I used `git cherry-pick`. Can do again with the `-x` will think about it tomorrow.

> Or do a merge with merge commit which (maybe) also preserves commit hashes?

I was also thinking about this, to keep the merge commit there, yes.

jstourac:
@jiridanek I have updated this PR so that all commits now have the info they are being cherry-picked from somewhere.

For others: I removed the 3e35a714e878bcce49bd62bcaff3bc5c0db79181 and added 5db9e26c59c9923a14e4db77e3f3702859037f79.

jiridanek (approved):

PR #1323, opened 2024-03-27
Add DependencyLibrary to support test case/suite dependency

manosnoam:
Robot Framework `DependencyLibrary` allows creating dependency between test cases and test suites. 
For example, we can create the foloowing logic:
If test A must be executed before test B, because test B expects some resources created by test A, 
but test A failed - then test B will not be executed and marked as skipped.

Adopting `DependencyLibrary` Keywards often in our sequential tests can save a lot of run-time and diagnostic time, in case of failed tests that were dependent on previous tests.

This commit includes a first use of the KW `Depends On Test` in: 420__model_serving.robot suite.

image

### DIFF
--- a/ods_ci/tests/Resources/Common.robot
+++ b/ods_ci/tests/Resources/Common.robot
@@ -2,6 +2,7 @@
 Library   SeleniumLibrary
 Library   JupyterLibrary
 Library   OperatingSystem
+Library   DependencyLibrary
 Library   Process
 Library   RequestsLibrary
 Library   ../../libs/Helpers.py
--- a/ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
+++ b/ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
@@ -44,7 +44,7 @@ Verify Openvino_IR Model Via UI
     [Documentation]    Test the deployment of an openvino_ir model
     [Tags]    Smoke
     ...       ODS-2054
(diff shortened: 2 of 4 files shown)


jstourac (approved):
I'm not in favor of having the dependent tests in general (setup and teardown seem nicer to me), but I agree that with our current situation and that setup may be slow for our use case, this can help. Thank you.

Also, there are some major upgrades of python dependencies included with this, so hopefully all will work as expected. Would be nice to run a couple of more tests (smoke?) with this PR too, though.

tarukumar (approved):

tarukumar:
Agree with Jan on running the smoke to make sure nothing is broken

manosnoam:
> I'm not in favor of having the dependent tests in general (setup and teardown seem nicer to me),

Counting on test setup (which means call test teardown later too) brings more code + more test runtime + more maintenance + debug time on falied tests that counts on earlier tests, as in my example:
`Test Inference Without Token Authentication` which runs in smoke after `Verify Openvino_IR Model Via UI` is expecting to have the model interface already running. Currently if you try to run `Verify Openvino_IR Model Via UI` solely, it will always fail. This is where `DependencyLibrary` is so effective.

Once we start to adopt it in our code, which anyhow uses a lot of dependency - without specifying it in the tests sometimes - we would save much more setup code lines, and failure analysis time.

> Also, there are some major upgrades of python dependencies included with this, so hopefully all will work as expected. Would be nice to run a couple of more tests (smoke?) with this PR too, though.
Yes, running as we speak :)

manosnoam:
I verified on 2.9 and the the pass rate was similar as we had for 2.9 before this commit.

PR #1328, opened 2024-03-27
Migrate to pipeline v2 (DataSciencePipelinesKfp)

diegolovison:
Tested on PSI and looks good


jstourac (changes requested):
Hi @diegolovison, to avoid same situation as this morning, could you please comply with the python linters before this is merged so that the CI is green? Thanks!

jiridanek (approved):

jstourac (approved):

PR #1329, opened 2024-03-27
remove clean all unused floating ips

kobihk:
When we run 2 jobs almost at the same time the first one allocate FIP 
but, at the same time, the second job delete the FIPs of the first one.
so we remove the clean all unused floating IPs and take care to delete the FIPs and AWS records as part of the delete cluster job.


jstourac (approved):

jstourac:
@kobihk we should probably merge it once we merge relevant change in our Jenkins, right? So that we don't spoil FIPs in OpenStack in the meantime.

kobihk:
> @kobihk we should probably merge it once we merge relevant change in our Jenkins, right? So that we don't spoil FIPs in OpenStack in the meantime.

Because of the $description of this PR I prefer to merge it as soon as we can and I'll take care to delete the FIP leftover later.
Thank you @jstourac

PR #1332, opened 2024-03-28
Fix linter errors in iris_pipeline.py

jiridanek:
* Apply `poetry run ruff check ods_ci/ --fix`

* Apply `poetry run black ods_ci/`

### DIFF
--- a/ods_ci/tests/Resources/Files/pipeline-samples/v2/iris_pipeline.py
+++ b/ods_ci/tests/Resources/Files/pipeline-samples/v2/iris_pipeline.py
@@ -1,61 +1,48 @@
-from typing import List
 import kfp
-from kfp import compiler
-from kfp import dsl
-from kfp import Client
-from kfp.dsl import Dataset
-from kfp.dsl import Input
-from kfp.dsl import Model
-from kfp.dsl import Output
-from kfp.dsl import ClassificationMetrics
+from kfp import Client, compiler, dsl
+from kfp.dsl import ClassificationMetrics, Dataset, Input, Model, Output
 
 
-@dsl.component(
-    base_image="quay.io/opendatahub/ds-pipelines-sample-base:v1.0",
-    packages_to_install=['pandas==2.2.0']
-)
+@dsl.component(base_image="quay.io/opendatahub/ds-pipelines-sample-base:v1.0", packages_to_install=["pandas==2.2.0"])
 def create_dataset(iris_dataset: Output[Dataset]):
     import pandas as pd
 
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Resources/Files/pipeline-samples/v2/iris_pipeline.py
```diff
@@ -66,57 +53,48 @@ def train_model(
...
     train_model(
-        normalized_iris_dataset=normalize_dataset_task
-        .outputs['normalized_iris_dataset'],
-        n_neighbors=neighbors)
+        normalized_iris_dataset=normalize_dataset_task.outputs["normalized_iris_dataset"], n_neighbors=neighbors
+    )
+
 
-endpoint = 'http://ml-pipeline-ui-kubeflow.apps.rmartine.dev.datahub.redhat.com/'
+endpoint = "http://ml-pipeline-ui-kubeflow.apps.rmartine.dev.datahub.redhat.com/"
```
jiridanek, line 98: > ```endpoint = "http://ml-pipeline-ui-kubeflow.apps.rmartine.dev.datahub.redhat.com/"```
Using http protocol is insecure. Use https instead

jiridanek: I'd ignore this

jstourac: I don't mind if you put it to ignore, but this bit seems to me like something that is irrelevant or will be overwritten on the actual run? I don't think this hostname will be working.

jstourac (approved):

ods_ci/tests/Resources/Files/pipeline-samples/v2/iris_pipeline.py
jgarciao: This sample pipeline is a verbatim copy from https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/docs/example_pipelines/iris/iris-pipeline.py. 

Currently in ods-ci we use the compiled version (iris_pipeline.yaml), which doesn't use that endpoint

jgarciao (approved):

jgarciao (approved):

ods_ci/tests/Resources/Files/pipeline-samples/v2/iris_pipeline.py
jiridanek: That's helpful, so probably the correct formatter fix should've been to disable formatter for this file, actually

PR #1333, opened 2024-03-28
`Open Data Science Project Details Page` also switch to a requested tab

manosnoam:
This PR brings in a new UI action, which is required for RHOAI 2.9 UI: Switch to DS section by tab ID. 
For example, to switch to the storage section we now need to call:
`Open Data Science Project Details Page         project_title=${PRJ_TITLE}, tab_id=cluster-storages`

image


bdattoma (dismissed):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
```diff
@@ -44,6 +45,11 @@ Open Data Science Project Details Page
         Click Link    ${project_title}
         Wait Until Project Is Open    project_title=${project_title}
     END
+    IF  "${tab_id}" != "${NONE}"
+        Click Button    //button[@role="tab" and @aria-controls="${tab_id}"]
+        Element Should Be Visible   //button[@role="tab" and @aria-controls="${tab_id}" and @aria-selected="true"]
+        Sleep    1s    reason=The tab page needs to load content according to user permissions
```
jstourac, line 51: Please, add the discussed https://issues.redhat.com/browse/RHOAIENG-4980 for reference here - once it is fixed, the performance should be improved and so we should update it here too, I think?

Hm, thinking about it more - wouldn't be better to use something like `Wait Until Page Contains`? With will wait just the right amount of time and no need to reference the issue above since we won't need to amend this code?

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
```diff
@@ -191,7 +189,7 @@
     [Tags]    Sanity    Tier1
     ...       ODS-1825    ODS-1972
     [Documentation]    Verifies users can add a Data connection to AWS S3
-    [Setup]    Run Keywords    Open Data Science Project Details Page       project_title=${PRJ_TITLE}
+    [Setup]    Run Keywords    Open Data Science Project Details Page       project_title=${PRJ_TITLE}    tab_id=workbenches
```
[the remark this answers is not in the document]
jstourac: cheap to fix this warning

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/434__data-science-pipelines-ui.robot
```diff
@@ -28,70 +28,57 @@ Verify User Can Create, Run and Delete A DS Pipeline From DS Project Details Pag
     ...                DS Project UI
     [Tags]    Smoke
     ...       ODS-2206    ODS-2226    ODS-2633
-
```
jstourac, line 31: I know that this is more about the personal preferences, but I kind of liked the way this code is split into smaller sections. It's a bit more readable for me. Since this PR isn't about code-style refactoring, do you mind to put these deleted lines back? :stuck_out_tongue_closed_eyes:

jstourac (commented):
Thank you, Noam. I put some comments.

For the more complex tests in workbenches, we will probably have to check and add necessary tab switching in the tests where probably needed now. Anyway, this is something that can be done in a followup PR.

bdattoma:
there are some missing places where the change should be applied:
- ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/425__model_serving_ovms_on_kserve.robot
- ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/424__model_serving_bias_metrics.robot
- ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/421__model_serving_gpu.robot
- ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/422__model_serving_llm_UI.robot
- ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/422__model_serving_llm_other_runtimes_UI.robot

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
manosnoam: I thought about mentioning it's a workaround for RHOAIENG-4980, however, quite often we already use sleep when expecting for backend response (e.g. look for `Sleep    5s    reason=let's not overload the API`). In this case we merely wait just 1 second, so it's legitimate I suppose.

I also prefer to wait for a certain element rather than static wait, but there is no such common element that appears after the tab is selected (`//button[@role="tab" and @aria-controls="${tab_id}" and @aria-selected="true"]`).
I asked Dev to add something like `//div[@aria-controls="details-section" and @aria-selected="false"]` which turns to `true` when detais section is ready.

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/434__data-science-pipelines-ui.robot
manosnoam: The line brakes within a KW/Test is not in our code style across 99% of our RF code + for me it looks confusing. @bdattoma what do you think ?

ods_ci/tests/Tests/400__ods_dashboard/415__ods_dashboard_projects/415__ods_dashboard_projects.robot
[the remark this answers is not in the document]
manosnoam: Actually we have so many lines in our code that have few extra chars above limit.
I tend to ignore it unless the extra characters are more than 10 (in this case it's only 4 more).

manosnoam:
> there are some missing places where the change might be applied, could you pls check these ones if not already done?

I know, that is intentionally. The tab_id is optional + in these tests I realized the switch to a certain tab is not required.

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/434__data-science-pipelines-ui.robot
bdattoma: I think there is not explicit mention to blank lines withing keywords or test cases. 

For some test cases, having blank lines highlights that it could make sense to split into multiple TCs. If that's not the case, i don't personally have a strong opinion about that.
For keywords, I don't have a strong opinion, but sometime I see that long kw/tc in RF might be difficult to read especially when there are many args. 

I

jstourac (approved):

bdattoma (approved):

PR #1337, opened 2024-03-28
Replace black and isort with ruff

jiridanek:
It's very fast, even without filtering for changed files the way isort/black did

image


ods_ci/utils/scripts/ocm/ocm.py
```diff
@@ -1221,8 +1221,9 @@ def update_ocm_policy(self):
         write_data_in_json(self.update_policies_json, data)
 
         schedule_cluster_upgrade = (
-            "ocm --v={} post /api/clusters_mgmt/v1/clusters/{}/upgrade_policies"
-            " --body {}".format(self.ocm_verbose_level, cluster_id, self.update_policies_json)
+            "ocm --v={} post /api/clusters_mgmt/v1/clusters/{}/upgrade_policies" " --body {}".format(
```
jstourac, line 1224:
```suggestion
            "ocm --v={} post /api/clusters_mgmt/v1/clusters/{}/upgrade_policies --body {}".format(
```
?

jstourac:
Thank you for this. One comment, otherwise LGTM.

So to be sure, because I know nothing :slightly_smiling_face:  - this replaces all the checks we had with black and isort? Or is it just some subset now?

jiridanek:
> So to be sure, because I know nothing 🙂 - this replaces all the checks we had with black and isort? Or is it just some subset now?

It's the same

* https://docs.astral.sh/ruff/faq/#how-does-ruffs-formatter-compare-to-black
* https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort

ods_ci/utils/scripts/ocm/ocm.py
jiridanek: makes sense; the previous less-than-ideal result of `poetry run ruff format` is most likely a ruff bug, actually

jstourac (approved):

bdattoma (approved):

PR #1340, opened 2024-04-02
Update new runtime image for tgis with snyk fixes

tarukumar:


jstourac:
@tarukumar whole of this should be backported to the `releases/2.8.0` branch too, I suppose?

bdattoma (approved):

jstourac:
> > @tarukumar whole of this should be backported to the `releases/2.8.0` branch too, I suppose?
> 
> yes

Would it be possible to share what is the best test to execute and test this change against 2.8.z (if you have it in memory) so I don't have to check sources once backporting this? :slightly_smiling_face:

FedeAlonso (approved):

jstourac (approved):

PR #1341, opened 2024-04-02
Add ruff to pre-commit

jiridanek:
This is opt-in feature to run ruff as pre-commit.

Enable this with

```
poetry run pre-commit install
```

Then, every time you commit, ruff will run on the Python code.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,11 @@
+repos:
+
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.3.4
+  hooks:
+    # Run the linter.
+    - id: ruff
+      args: [ --fix ]
+    # Run the formatter.
+    - id: ruff-format
--- a/poetry.lock
+++ b/poetry.lock
@@ -171,17 +171,17 @@ files = [
 
 [[package]]
 name = "awscli"
-version = "1.32.73"
+version = "1.32.79"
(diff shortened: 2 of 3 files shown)


jstourac:
1. Thank you for this, Jiri. I suppose that it works just fine in your setup where you have global pre-commit hook configured. There was some discussion how this will work in a setup where there is a pre-commit hook per repo? Has this been tested?
Note: not sure if I expressed myself correctly :smile: 

2. Anyway, if this is something that should be enabled/disabled manually per user - we should put some note somewhere in our readme/doc. WDYT?

3. Also, what happens once the `pre-commit` dependency is updated in the `pyproject.toml` file? Is it necessary to enable it again so the updated dependency is used or is it grabbed automatically?

jiridanek:
1. You said it exactly right. Actually, what I have is the unsupported configuration that requires workarounds to make it work. https://stackoverflow.com/questions/67793193/cowardly-refusing-to-install-hooks-with-core-hookspath-set/78243646#78243646 What everybody else has is the recommended setup and this should work seamlessly with it.

2. I wanted to first see how it works for others. I added a topic to tomorrow's meeting, proposing to merge this so people can experiment how it works for them.

3. The pre-commit tool is responsible for setting up a local git hook, which is a shell script. The way the shell script works is that it detects active venv and runs pre-commit from a venv. I'm willing to take it on faith that it will work fine. Pre-commit is what everybody uses.

jgarciao (approved):

PR #1346, opened 2024-04-04
RHOAIENG-5092 - Validate Authorino Resources

mattmahoneyrh:
Main purpose of this PR is the introduction of a new test case that validates RHOAI operator integration with Authorino. The test case validates that Authorino resources are properly created.

  Note: This is an initial test case, with more test coverage expected.

Secondary inclusion in this PR is a fix to the Service Mesh 'Managed' test case, for checking that the istiod pod is running.


mattmahoneyrh:
Tests passed in Jenkins

jstourac (approved):

PR #1352, opened 2024-04-08
Add common KW `Open "Get Started" For App Name`

manosnoam:
- Replacing duplicated code to open the "Get Started" of Explorer cards
- Fix the Selenium error while trying to select an Explorer card: `Element is not clickable at point. Other element would receive the click`

Before this fix, our tests failed on:

image

With this fix it is resolved:

image

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
@@ -204,35 +204,32 @@ Remove Disabled Application From Enabled Page
    Wait Until Page Does Not Contain Element    xpath://div[@id='${app_id}']
    Capture Page Screenshot  ${app_id}_removed.png
 
-
-Verify Service Provides "Enable" Button In The Explore Page
-  [Documentation]   Verify the service appears in Applications > Explore and, after clicking on the tile, the sidebar opens and there is an "Enable" button
+Open "Get Started" For App Name
+  [Documentation]   Open Explore page and select a Card titled "${app_name}", to view its "Get Started" on the sidebar
   [Arguments]  ${app_name}    ${app_id}=${NONE}
   Menu.Navigate To Page    Applications    Explore
   Wait For RHODS Dashboard To Load    expected_page=Explore
   IF    "${app_id}" == "${NONE}"
-      ${card_locator}=    Set Variable    //*[.='${app_name}']/../../..//input[@type='radio']
+      ${card_locator}=    Set Variable    //*[contains(@data-testid,"cardtitle") and contains(text(),'${app_name}')]
   ELSE
       ${card_locator}=    Set Variable    ${CARDS_XP}\[@id='${app_id}']
   END
   ${status}=    Open Get Started Sidebar And Return Status    card_locator=${card_locator}
   Capture Page Screenshot
   Run Keyword And Continue On Failure    Should Be Equal    ${status}    ${TRUE}
+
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -343,8 +338,9 @@ Open Get Started Sidebar And Return Status
     [Arguments]  ${card_locator}
     Wait Until Element Is Visible    xpath:${card_locator}
     Wait Until Element Is Enabled     xpath:${card_locator}    timeout=20s     error=Element is not clickbale  #robocop : disable
-    Click Element    ${card_locator}
-        ${status}=  Run Keyword and Return Status  Wait Until Page Contains Element    xpath://div[contains(@class,'pf-v5-c-drawer__panel-main')]
+    Mouse Down    ${card_locator}
+    Mouse Up    ${card_locator}
```
bdattoma, line 342: why?

manosnoam: Because of clicking element fails on "Other element would receive the click", no matter how I try to optimize the Xpath. For example:
`Click Element    //*[contains(@data-testid,"cardtitle") and contains(text(),'Anaconda Professional')]/../../..//input[@type='radio']`

> robot.errors.HandlerExecutionFailed: ElementClickInterceptedException: Message: element click intercepted: Element <input aria-labelledby="anaconda-ce" id="anaconda-ce-selectable-card-id" name="odh-explore-selectable-card" class="pf-v5-c-radio__input" type="radio" aria-invalid="false" data-ouia-component-type="PF5/Radio" data-ouia-safe="true" data-ouia-component-id="OUIA-Generated-Radio-1"> is not clickable at point (613, 240). Other element would receive the click: <label class="pf-v5-c-radio__label" for="anaconda-ce-selectable-card-id"></label>
>   (Session info: chrome=122.0.6261.69)

bdattoma: the card should be clickable yes. In the past they disabled this option, but iirc they put it back. This would need a double check

manosnoam: That is what I do :) click on the whole card:
image

bdattoma: yeah but why not using Click Element instead of Mouse Down/Up?

jstourac: @bdattoma yes, they have put it back already

@manosnoam I don't think so - you click on the child div element instead of on the whole card:

Screenshot from 2024-04-08 10-38-26

manosnoam: I thought so too, that clicking on `//*[contains(@data-testid,"cardtitle") and contains(text(),'Anaconda Professional')]` will work, but as I explained, this fails in our tests. 
In "Browser Developer mode" you can find the element, but in Selenium runtime, it fails to click on it with error `Other element would receive the click`. Try to debug test `Verify Anaconda Professional Is Available In RHODS Dashboard Explore/Enabled Page` and see...

jstourac (approved):

jstourac (approved):

bdattoma (approved):

PR #1364, opened 2024-04-09
Add ibmcloud among supported platform for RHOAI install task

bdattoma:


bdattoma:
image

ods_ci/tasks/Tasks/rhods_olm.robot
```diff
@@ -11,7 +11,7 @@ Library          String
 ${cluster_type}                 selfmanaged
 ${image_url}                    ${EMPTY}
 ${RHODS_OSD_INSTALL_REPO}       None
-@{SUPPORTED_TEST_ENV}           AWS   AWS_DIS   GCP   PSI   PSI_DIS   ROSA
+@{SUPPORTED_TEST_ENV}           AWS   AWS_DIS   GCP   PSI   PSI_DIS   ROSA    IBM_Cloud
```
jstourac, line 14: It's a nitpick, but could be possible to:
```suggestion
@{SUPPORTED_TEST_ENV}           AWS   AWS_DIS   GCP   PSI   PSI_DIS   ROSA   IBM_Cloud
```

jstourac (approved):

ods_ci/tasks/Tasks/rhods_olm.robot
bdattoma: done

ods_ci/tasks/Tasks/rhods_olm.robot
```diff
@@ -11,7 +11,7 @@ Library          String
 ${cluster_type}                 selfmanaged
 ${image_url}                    ${EMPTY}
 ${RHODS_OSD_INSTALL_REPO}       None
-@{SUPPORTED_TEST_ENV}           AWS   AWS_DIS   GCP   PSI   PSI_DIS   ROSA
+@{SUPPORTED_TEST_ENV}           AWS   AWS_DIS   GCP   PSI   PSI_DIS   ROSA   IBM_Cloud
```
jstourac, line 14: Hm, one more thing - I noticed now that we use uppercase for all the other variants. Wouldn't it be better to use `IBM_CLOUD` instead of `IBM_Cloud`? (same would have to apply also for the Jenkins code later on).

jstourac (dismissed):

ods_ci/tasks/Tasks/rhods_olm.robot
bdattoma: I used Camel Case cos the others are all abbreviation. 
I don't think it's a big deal, I can make it upper case but don't see much value in one of the other option

jstourac: Upper case - easier to remember (don't have to think what part is uppercase and what part not); I don't insist. If nobody else has objections.

kobihk: Maybe we can use only IBM - 3 characters as most of the other platforms

jiridanek (dismissed):

ods_ci/tasks/Tasks/rhods_olm.robot
bdattoma: I'll make it upper case, cheap change.

Kobi I wanted to keep the original Cloud name like we do for the other providers

jiridanek (approved):

ods_ci/tasks/Tasks/rhods_olm.robot
jiridanek: Don't know why, but i get FORTRAN vibes from this uppercase thing https://www.fortran90.org/src/best-practices.html

kobihk (approved):

PR #1365, opened 2024-04-09
Verify That There Are No Sample Pipelines After Creating Pipeline Server

Raghul-M:
Added Automation Test for RHOAIENG-4382


Raghul-M:
Verified success with Jenkins Build #5069

PR #1366, opened 2024-04-09
Speed-up execution of keyword "Is RHODS Self-Managed" in local

jgarciao:
The keyword `RHOSi Setup` calls keyword `Is RHODS Self-Managed`, which is using the Robot Framework OpenShiftLibrary to perform `Oc Get ...` 

The problem is that the OpenShiftLibrary library takes more than 10s to load for the first time, slowing down development in local. 

This PR modifies the `Is RHODS Self-Managed` keyword not to use the OpenShift Library in order to speed up execution.

Tested successfully in local and rhods-ci-pr-test/2683

Original:
speed-if-rhods-is-self-managed-original-managed

With this PR:
speed-if-rhods-is-self-managed-updated-managed

### DIFF
--- a/ods_ci/tests/Resources/OCP.resource
+++ b/ods_ci/tests/Resources/OCP.resource
@@ -65,7 +65,7 @@ Is RHODS Self-Managed
     ...                 in the cluster (e.g., usually on OSD, ROSA).
     ...                 Returns ${TRUE} if RHODS Self-Managed is installed or if PRODUCT=ODH
     IF  "${PRODUCT}" == "ODH"  RETURN     ${TRUE}
-    ${rc} =    Run And Return Rc    oc get catalogsource addon-managed-odh-catalog -n ${OPERATOR_NAMESPACE}
+    ${rc}=    Run And Return Rc    oc get catalogsource addon-managed-odh-catalog -n ${OPERATOR_NAMESPACE}
     Run Keyword And Return    Run Keyword And Return Status      Should Not Be Equal As Numbers    ${rc}    ${0}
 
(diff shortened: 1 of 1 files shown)


jiridanek (approved):

FedeAlonso (approved):

ods_ci/tests/Resources/OCP.resource
```diff
@@ -65,11 +65,8 @@ Is RHODS Self-Managed
     ...                 in the cluster (e.g., usually on OSD, ROSA).
     ...                 Returns ${TRUE} if RHODS Self-Managed is installed or if PRODUCT=ODH
     IF  "${PRODUCT}" == "ODH"  RETURN     ${TRUE}
-    ${is_managed} =    Run Keyword And Return Status    OpenshiftLibrary.Oc Get
-    ...                                                      kind=CatalogSource
-    ...                                                      name=addon-managed-odh-catalog
-    ...                                                      namespace=${OPERATOR_NAMESPACE}
-    Run Keyword And Return    Evaluate    not ${is_managed}
+    ${rc}=    Run And Return Rc    oc get catalogsource addon-managed-odh-catalog -n ${OPERATOR_NAMESPACE}
```
jstourac, line 68: Would be nice to put a comment above this line so we assure that nobody else will change it back to use OpenshiftLibrary again in the future :slightly_smiling_face:

jgarciao: I've added it

jiridanek (approved):

PR #1368, opened 2024-04-10
Verify That There Are No Sample Pipelines After Creating Pipeline Server in RHOAI

Raghul-M:
Added Automation Test for RHOAIENG-4382

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataSciencePipelines.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataSciencePipelines.resource
@@ -139,13 +139,17 @@ Verify There Is No "Error Displaying Pipelines" After Creating Pipeline Server
     Run Keyword And Warn On Failure    Page Should Not Contain     Error displaying pipelines
 
 Verify That There Are No Sample Pipelines After Creating Pipeline Server
-    [Documentation]     Verify creation of sample pipelines by default when deploying dsp in RHOAI
-    ${status}=      Run Keyword And Return Status   Wait Until Page Contains    Create pipeline    timeout=10s
-    IF  ${status}
-        Log     Default Sample Pipelines are Not Available and No Pipelines Created
-    ELSE
-        Log     Default Sample Pipelines are Available
+    [Documentation]     Verify creation of sample pipelines by default when deploying dsp in RHOAI/ODH
+    Log    ${PRODUCT}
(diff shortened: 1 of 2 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataSciencePipelines.resource
```diff
@@ -138,6 +138,15 @@ Verify There Is No "Error Displaying Pipelines" After Creating Pipeline Server
     Run Keyword And Ignore Error       Wait Until Page Contains    Error displaying pipelines    timeout=60s
     Run Keyword And Warn On Failure    Page Should Not Contain     Error displaying pipelines
 
+Verify That There Are No Sample Pipelines After Creating Pipeline Server
```
jgarciao, line 141: I believe in ODH the pipeline samples will be available after creating a pipeline server. You could 
call your keyword `Verify Sample Pipelines Existence After Creating Pipeline Server` and fail (not log) if the samples are created for RHOAI .

Something like this (not tested):
```
${samples_created}=      Run Keyword And Return Status   Wait Until Page Contains    Create pipeline    timeout=10s
IF  "${PRODUCT}" == "RHODS" and ${samples_created}
    Fail    Default Sample Pipelines should not be created after creating a pipeline server in ${PRODUCT}
ELSE IF     "${PRODUCT}" == "ODH" and not ${samples_created}
    Fail    Default Sample Pipelines should be created after creating a pipeline server in ${PRODUCT}
END
```

jgarciao (changes requested):

Raghul-M:
Verified success with Jenkins Build #5099

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataSciencePipelines.resource
jiridanek: > I believe in ODH the pipeline samples will be available after creating a pipeline server. You could

if so, then that's a bug, https://issues.redhat.com/browse/RHOAIENG-4382, @HumairAK is working on removing the samples from both ODH and RHOAI

Raghul-M: Yes, @jiridanek  modified the test to fail if samples are available in either RHOAI or ODH

diegolovison (approved):

jiridanek (approved):

jgarciao (approved):

PR #1380, opened 2024-04-15
Adapt Trusted CA bundles tests to ODH

asanzgom:


asanzgom:
rhods-ci-pr-test/2692/ In progress

jstourac:
Basically LGTM.

I just don't feel very comfortable with the different approaches we have to configure product specific constants - somewhere we count respective values in the running code directly, elsewhere we expect to use [this file](https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/test-variables-odh-overwrite.yml).

mattmahoneyrh:
@asanzgom Has this PR been by Jenkins job for both ODH and RHOAI?

asanzgom:
Triggered odh-smoke/52/

bdattoma:
> Basically LGTM.
> 
> I just don't feel very comfortable with the different approaches we have to configure product specific constants - somewhere we count respective values in the running code directly, elsewhere we expect to use [this file](https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/test-variables-odh-overwrite.yml).

Agree with Jan's point

asanzgom:
> @asanzgom Has this PR been by Jenkins job for both ODH and RHOAI? Also, this is also a fix to the Serverless test, right?

I have triggered for ODH now odh-smoke/52/ and will trigger on another cluster with RHOAI and yes, I moved common Keyword and fixed compatibility backwards for the serverless Test Case

asanzgom:
odh-smoke/55/ passed

asanzgom:
Ran successful on RHOAI rhods-smoke/5103

ods_ci/tests/Resources/RHOSi.resource
```diff
@@ -70,6 +70,7 @@ Protect Sensitive Variables In Keywords
 Initialize Global Variables
     [Documentation]   Sets global variables
     [Arguments]    ${force_fetch}=False
+    Assign Vars According To Product    ${PRODUCT}
```
jgarciao, line 73: ${PRODUCT} is already a global variable so you don't need this argument

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
```diff
@@ -73,10 +73,11 @@ Validate Trusted CA Bundles State Removed
 *** Keywords ***
 Suite Setup
     [Documentation]    Suite Setup
+    Assign Vars According To Product    ${PRODUCT}
```
jgarciao, line 76: This is not needed, as it will be already called in RHOSI Setup

ods_ci/tests/Tests/100__deploy/130__operators/140__serverless_operator/140__serverless_operator.robot
```diff
@@ -42,23 +42,13 @@ Suite Setup
     [Documentation]    Suite Setup
     Set Library Search Order    SeleniumLibrary
     RHOSi Setup
+    Assign Vars According To Product    ${PRODUCT}
```
jgarciao, line 45: This is not needed (already done in RHOSi Setup)

jgarciao (changes requested):

mattmahoneyrh (dismissed):

jgarciao:
> I just don't feel very comfortable with the different approaches we have to configure product specific constants - somewhere we count respective values in the running code directly, elsewhere we expect to use [this file](https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/test-variables-odh-overwrite.yml).

I think the code at `Assign Vars According To Product` could be moved to the existing `Initialize Global Variables` and use it only for global variables  (variables that might be used in more than one test suite). For variables that will be used only in a test suite, they can be declared in the test suite itself and then overwritten for ODH using [test-variables-odh-overwrite](https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/test-variables-odh-overwrite.yml).

ods_ci/tests/Resources/RHOSi.resource
asanzgom: I had to include it as an input param because of:
`Validate Trusted CA Bundles State Managed :: The purpose of this t... | FAIL |
Parent suite setup failed:
Keyword 'RHOSi.Assign Vars According To Product' expected 1 argument, got 0.`

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
asanzgom: I had to include it as an input param because of:
`Validate Trusted CA Bundles State Managed :: The purpose of this t... | FAIL |
Parent suite setup failed:
Keyword 'RHOSi.Assign Vars According To Product' expected 1 argument, got 0.`

ods_ci/tests/Tests/100__deploy/130__operators/140__serverless_operator/140__serverless_operator.robot
asanzgom: I had to include it as an input param because of:
`Validate Trusted CA Bundles State Managed :: The purpose of this t... | FAIL |
Parent suite setup failed:
Keyword 'RHOSi.Assign Vars According To Product' expected 1 argument, got 0.`

asanzgom:
> > I just don't feel very comfortable with the different approaches we have to configure product specific constants - somewhere we count respective values in the running code directly, elsewhere we expect to use [this file](https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/test-variables-odh-overwrite.yml).
> 
> I think the code at `Assign Vars According To Product` could be moved to the existing `Initialize Global Variables` and use it only for global variables (variables that might be used in more than one test suite). For variables that will be used only in a test suite, they can be declared in the test suite itself and then overwritten for ODH using [test-variables-odh-overwrite](https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/test-variables-odh-overwrite.yml).

It is already moved inside Initialize Global Variables

ods_ci/tests/Tests/100__deploy/130__operators/140__serverless_operator/140__serverless_operator.robot
jgarciao: I still see this line to be removed

jgarciao (changes requested):

ods_ci/tests/Resources/RHOSi.resource
asanzgom: Removed

jgarciao (approved):

jgarciao (approved):

jgarciao:
Merging after the agreement in today's Automation Meeting

PR #1384, opened 2024-04-16
fix: ODS-2054 issues

FedeAlonso:
During the nightlies exectuion (e.g.: autotrigger-smoke/124) the test "ODS-2054 - Verify Openvino_IR Model Via UI" was failing, and it was because the flow has been modified
image

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
@@ -119,9 +119,8 @@ Select Model Server
     [Documentation]    If there are multiple model servers to choose from, this keyword
     ...    Selects the given one
     [Arguments]    ${model_server}
-    Sleep    1s     reason=It's taking a little bit time to load the Model Server options
     ${selectable}=    Run Keyword And Return Status
-    ...    Page Should Contain Element     xpath://span[.="Model servers"]/../../..//button[@aria-label="Options menu"]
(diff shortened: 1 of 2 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -120,6 +119,7 @@ Select Model Server
     [Documentation]    If there are multiple model servers to choose from, this keyword
     ...    Selects the given one
     [Arguments]    ${model_server}
+    Sleep    1s     reason=It's taking a little bit time to load the Model Server options
```
jstourac, line 122: Isn't there some possibility to wait until some particular element is present/enabled on the page instead of this 1 sec wait?

jstourac (dismissed):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
bdattoma: +1, sleep risks to be unstable

FedeAlonso: @jstourac @bdattoma modified in order to use an active wait. 
Tested in possitive and negative cases.
Thanks for pointing it ;)

FedeAlonso:
Test results after modifications: 
image

bdattoma (approved):

asanzgom (approved):

PR #1387, opened 2024-04-17
Fix required xpaths for the selection of the model server template

jstourac:
This change is necessary for the latest RHOAI 2.9 preRC build.

---

Local check:

```
./ods_ci/run_robot_test.sh --extra-robot-args '-i ODS-2626 -i ODS-2519 -i ODS-2552 -i ODS-2555 -i ODS-2268' --open-report true
```

CI: `rhods-smoke/5113/`

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
@@ -76,7 +76,7 @@ Serve Model
     Select Project    ${project_name}
     Set Model Name    ${model_name}
     Select Model Server    ${model_server}
-    IF    "${serving_runtime}" != "${NONE}"    Select Kserve Serving Runtime    ${serving_runtime}
+    IF    "${serving_runtime}" != "${NONE}"    Set Model Server Runtime    ${serving_runtime}
     SeleniumLibrary.Wait Until Page Contains Element    xpath://span[.="Model framework (name - version)"]
     Select Framework    ${framework}
     IF    ${existing_data_connection}==${TRUE}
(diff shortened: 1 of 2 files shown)


manosnoam:
To verify KW `Create Model Server` please run suite `420__model_serving.robot`.
In addition, consider adding it directly into `Serve Model` since `Create Model Server` is always called before `Serve Model` (except in test `Verify GPU Model Deployment Via UI`, but it seems OK to have that there too).

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -411,12 +411,12 @@ Deploy Kserve Model Via UI  #robocop: disable
     Click Button    Deploy
     Wait Until Page Does Not Contain Element    xpath=${KSERVE_MODAL_HEADER}   timeout=60s
 
-Select Kserve Serving Runtime
+Set Model Server Runtime
     [Documentation]    Opens the Runtime dropdown in the deploy modal for kserve models and select the given runtime
```
FedeAlonso, line 415: The doc says "...for **kserve** models...". Shouldn't it be more generic (like in the title, which now it's not about KServe)

FedeAlonso (changes requested):

jstourac:
> To verify KW Create Model Server please run suite 420__model_serving.robot.

That is the reason I run the ODS-2268 - is it really necessary to run the whole suite for this change? :thinking: 

> In addition, consider adding it directly into Serve Model since Create Model Server is always called before Serve Model (except in test Verify GPU Model Deployment Via UI, but it seems OK to have that there too).

I'm not gonna do that - I'm not gonna touch the code more for this change now. I'm not the expert here and this was supposed to be just an xpath change to work with the current build. If we want more changes, I am okay to hand this over to somebody who is more experienced here and knows the code.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
jstourac: updated

FedeAlonso (approved):

manosnoam (approved):

PR #1391, opened 2024-04-17
Add self-signed CA bundle support for kserve grpc/http requests

lugi0:
This PR adds support for passing self-signed (or anyway customized) CA bundles to Kserve GRPC and HTTP inference requests.
It is assumed that the request will either be insecure or have a valid CA bundle attached to it.

For GRPC requests, we use either the `-insecure` option or `-cacert` option (incompatible with each other). `-cacert` expects the file path to a valid CA Bundle.
For HTTP requests, we use the `verify` argument provided by the Requests library in Robot Framework - the argument accepts `True`/`False` (secure/insecure request) or a valid path to a CA bundle. 

Both options are abstracted into the `Query Model Multiple Times` keyword, which has a new argument `cert` which defaults to `False` (i.e. insecure requests for both GRPC and HTTP).  In order to validate the request, we need to pass a valid CA bundle file path as the `cert` argument.

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.resource
@@ -212,13 +212,18 @@ Click Minus Button
 Query Model With GRPCURL
     [Arguments]    ${host}    ${port}    ${endpoint}    ${json_body}
     ...            ${json_header}=${NONE}    ${insecure}=${FALSE}    ${plaintext}=${FALSE}
-    ...            ${background}=${NONE}    ${skip_res_json}=${FALSE}    &{args}
+    ...            ${background}=${NONE}    ${skip_res_json}=${FALSE}    ${cabundle_file}=${False}    &{args}
     ${cmd}=    Set Variable    grpcurl -d ${json_body}
     IF    $json_header != None
         ${cmd}=    Catenate    ${cmd}    -H ${json_header}
     END
     IF    ${insecure} == ${TRUE}
         ${cmd}=    Catenate    ${cmd}    -insecure
+    ELSE  #-insecure and -cacert are incompatible with each other. Assuming we want either option always on.
+        IF    ${cabundle_file}==${False}
+            Log    The call will fail because you have not provided a valid CA bundle file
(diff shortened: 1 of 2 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.resource
```diff
@@ -206,13 +206,18 @@ Click Minus Button
 Query Model With GRPCURL
     [Arguments]    ${host}    ${port}    ${endpoint}    ${json_body}
     ...            ${json_header}=${NONE}    ${insecure}=${FALSE}    ${plaintext}=${FALSE}
-    ...            ${background}=${NONE}    ${skip_res_json}=${FALSE}    &{args}
+    ...            ${background}=${NONE}    ${skip_res_json}=${FALSE}    ${cabundle_file}=${False}    &{args}
```
bdattoma, line 209: can we use the same arg name as for the other keyword?

bdattoma:
lgtm. Could you use the cert in at least one of the test cases we have?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.resource
lugi0: the other keyword uses `verify` because it's a built-in of the requests library, I wouldn't want to mix the two to be honest

lugi0:
@bdattoma IMHO it would be better to leave the implementation of a TC to another PR

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/ModelServer.resource
bdattoma: I was referring to `cert` vs `cabundle_file`

lugi0: Done

mwaykole (approved):

tarukumar (approved):

bdattoma (approved):

PR #1397, opened 2024-04-22
Pipeline for manual testing

diegolovison:
Long term, create a run and double check for a failure. Rerun and double check if the previous failure is matching.


jgarciao:
Hi @diegolovison . Is this sample still needed? If  yes, could you store it in https://github.com/red-hat-data-services/ods-ci/tree/master/ods_ci/tests/Resources/Files/pipeline-samples/v2/cache-disabled/ ? Thanks!

PR #1398, opened 2024-04-22
Rhoaieng 6358 restore custom ca bundle

mattmahoneyrh:
This fix will restore the original DSCi customCABundle to original value.

Verified fix by populating customCABundle with a CA prior to running the tests, and then ensuring that the CA was restored.

### DIFF
--- a/ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
+++ b/ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
@@ -25,6 +25,8 @@ Validate Trusted CA Bundles State Managed
     ...    each non-reserved namespace.
     [Tags]    Operator    Smoke    ODS-2638    TrustedCABundle-Managed
 
+    ${saved_custom_ca_bundle}=    Get Custom CA Bundle Value In DSCI     ${DSCI_NAME}    ${OPERATOR_NS}
+
     Wait Until Keyword Succeeds    2 min    0 sec
     ...    Is Resource Present    project    ${TEST_NS}    ${TEST_NS}    ${IS_PRESENT}
 
@@ -39,12 +41,14 @@ Validate Trusted CA Bundles State Managed
     Wait Until Keyword Succeeds    2 min    0 sec
     ...    Is CA Bundle Value Present    ${TRUSTED_CA_BUNDLE_CONFIGMAP}    ${CUSTOM_CA_BUNDLE}    ${TEST_NS}    ${IS_PRESENT}
 
-    [Teardown]     Restore DSCI Trusted CA Bundle Settings
+    [Teardown]     Restore DSCI Trusted CA Bundle Settings    ${saved_custom_ca_bundle}
 
 Validate Trusted CA Bundles State Unmanaged
     [Documentation]  The purpose of this test case is to validate Trusted CA Bundles when in state Unmanaged
     ...    With Trusted CA Bundles Unmanaged, ConfigMap odh-trusted-ca-bundle will not be managed by the operator.
     [Tags]    Operator    Smoke    ODS-2638    TrustedCABundle-Unmanaged
+
+    ${saved_custom_ca_bundle}=    Get Custom CA Bundle Value In DSCI     ${DSCI_NAME}    ${OPERATOR_NS}
 
     Set Trusted CA Bundle Management State    ${DSCI_NAME}    Unmanaged    ${OPERATOR_NS}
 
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
```diff
@@ -85,8 +91,11 @@ Suite Teardown
 
 Restore DSCI Trusted CA Bundle Settings
     [Documentation]    Restore DSCI Trusted CA Bundle settings to original tate
+    [Arguments]    ${custsom_ca_value}
+
     Set Custom CA Bundle Value In DSCI    ${DSCI_NAME}   ''    ${OPERATOR_NS}
```
jstourac, line 96: Is there a reason to keep this line?

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
```diff
@@ -85,8 +91,11 @@ Suite Teardown
...
     Set Trusted CA Bundle Management State    ${DSCI_NAME}    Managed    ${OPERATOR_NS}
```
jstourac, line 97: The management state would also deserve similar check for original value.

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
```diff
@@ -25,6 +25,8 @@ Validate Trusted CA Bundles State Managed
     ...    each non-reserved namespace.
     [Tags]    Operator    Smoke    ODS-2638    TrustedCABundle-Managed
 
+    ${saved_custom_ca_bundle}=    Get Custom CA Bundle Value In DSCI     ${DSCI_NAME}    ${OPERATOR_NS}
```
jstourac, line 28: Calling this function in a suite setup would be even better, if possible.

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
mattmahoneyrh: No, removed the line.

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
mattmahoneyrh: yes, though lets make this another PR since similar approach is needed for additional operator test cases, if that is OK with you.

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
mattmahoneyrh: Do you have an example of how to save a value in a global, and then access the value in a test case (was not able to determine a working example)?

mattmahoneyrh:
Jenkins run passed.

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
jstourac: Hm, I would expect that creating a `*** Variables ***` section and assigning a variable defined there a value in the setup function would work, see e.g. here https://docs.robotframework.org/docs/variables.

There is another possibility to use `Set Global Variable`, but I don't like that approach.

I can live with the variant as is right now, but would make sense to have it in some common setup phase instead in every test :slightly_smiling_face:

jstourac (approved):

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
mattmahoneyrh: There are additional test cases where saving/restoring state is also needed, and a better approach is needed. Similarly for Management state, and in particular when testing "Managed" and the start configuration is Removed or Unmanaged.

jstourac: yeah, can't disagree

bdattoma: You can use `Set Suite Variable <var_name> <var_value>`

apodhrad (approved):
As a hot fix it is ok but Jan's suggestion about storing the value in a suite setup is a very valid suggestion.

ods_ci/tests/Tests/100__deploy/130__operators/130__rhods_operator/139__trusted_ca_bundles.robot
jstourac: Followup #1401

PR #1402, opened 2024-04-23
Fix/420 model serving dashboard tests

FedeAlonso:
During the execution of the 2.9 QGs we've found that some of the tests related with the suite 420_model_serving were not working properly. Trying to fix it we've found that also there were some race condition
I've modified the tests a little bit in order to use it's own resources (DSP name 'model-serving-project' is used in multiple suites, which is not a good practice) and also now all the DSProjects related with this suite are deleted via CLI
Test results: 
image


bdattoma:
lgtm. However I'd ask @manosnoam to review the change because he brought in the re-use of existing projects in these tests.

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -216,18 +217,12 @@ Create Openvino Models
...
     END
 
+
 Model Serving Suite Teardown
     [Documentation]    Suite teardown steps after testing DSG. It Deletes
     ...                all the DS projects created by the tests and run RHOSi teardown
-    # Even if kw fails, deleting the whole project will also delete the model
-    # Failure will be shown in the logs of the run nonetheless
-    IF    ${MODEL_CREATED}
-        Clean All Models Of Current User
```
manosnoam, line 225: @FedeAlonso Thanks for the PR. 
Can you run all sanity tests of 420 suite, and verify that the `Model Serving Suite Teardown` works for all  ?
Specifically, replacing the KW `Clean All Models Of Current User` in that teardown, with `Delete All DS Projects With Name Like    ${PRJ_TITLE}` might not work as expected (the current user might have other projects to be deleted, or similar project names of another user should not be deleted).

FedeAlonso: @manosnoam I've noticed that we were using the DSProject name "model-serving-project" in multiple suites, so what I've done here is using the DSProject name "model-serving-project-420" as a base name (depending on the test it uses this one or adds something at the end), because it's a good practice to isolate the resources for each suite or test (preventing race conditions and this kind of hazards). So now almost each of this tests creates it's own DSProject.
With that modification, I've found that when trying to execute all this tests in a small cluster, It ran out of resources. So what I've done is that the deletion of the DSProjects comes as a Test Teardown instead of a Suite Tear Down.
I've also checked that the models are deleted when the DSProject is deleted and it works too.
I'll post the Test results below.

FedeAlonso:
Test Results of the complete Suite after the latest modifications:
image

manosnoam (commented):
lgtm, besides one suggestion

manosnoam (dismissed):

ods_ci/tests/Resources/CLI/DSProjects/DSProjects.resource
```diff
@@ -0,0 +1,20 @@
+*** Settings ***
+Documentation    Collection of keywords to interact with the administration of DS Projects via CLI
+
+
+*** Keywords ***
+
+Delete All DS Projects With Name Like
+    [Documentation]    Delete all accelerator profiles with a name that partially or totally matches with the given
```
jstourac, line 8: the doc is misleading

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -177,7 +183,9 @@ Verify Editing Existing Model Deployment
     Verify Model Status    ${MODEL_NAME}    success
     Run Keyword And Continue On Failure    Verify Model Inference    ${MODEL_NAME}    ${INFERENCE_INPUT_OPENVINO}
     ...    ${EXPECTED_INFERENCE_OUTPUT_OPENVINO}    token_auth=${FALSE}
-    [Teardown]    Run Keyword If Test Failed    Get Modelmesh Events And Logs
+    [Teardown]   Run Keywords    Model Serving Test Teardown
```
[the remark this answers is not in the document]
jstourac: Are we sure that we want to do this in this order? First the `Model Serving Test Teardown` (DS projects deletion) and then the `Get Modelmesh Events And Logs`? Don't we loose any data by the deletion before we get the events and logs? No logs from pods running in the DS project are gathered?

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
```diff
@@ -217,19 +226,13 @@ Create Openvino Models
...
-    # Will only be present on SM cluster runs, but keyword passes
-    # if file does not exist
+    [Documentation]    Suite teardown steps after testing DSG.
     Remove File    openshift_ca.crt
     Close All Browsers
     RHOSi Teardown
+
+Model Serving Test Teardown
+    [Documentation]    Test teardown steps after test. It Deletes
+    ...                all the DS projects created by the tests and run RHOSi teardown
```
jstourac, line 236: RHOSi teardown too? :thinking:

jstourac (commented):
LGTM in general, we should be careful what we put for the deletion for the DS projects with this grep, though.

Put some minor comments.

asanzgom (approved):

FedeAlonso:
Test results with the latest changes:
image

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/420__model_serving.robot
FedeAlonso: Copypasting like a rookie.
Thanks for noticing!
Fixed in [b1a4d1c](https://github.com/red-hat-data-services/ods-ci/pull/1402/commits/b1a4d1c5afd7abc376bb0f9ef49f64c00be504b0)

ods_ci/tests/Resources/CLI/DSProjects/DSProjects.resource
FedeAlonso: Fixed in [b1a4d1c](https://github.com/red-hat-data-services/ods-ci/pull/1402/commits/b1a4d1c5afd7abc376bb0f9ef49f64c00be504b0)

manosnoam (approved):

PR #1405, opened 2024-04-24
Fix odh-nightly on managed cluster

tarukumar:


jstourac:
Yeah, this is ugly. Is there a way we can move our nightly from managed to self-managed? If not, would you mind to put a comment on the line why we're using self-managed install type on the managed cluster? So it's more obvious to anyone who will look the code in the future that this is desired...

tarukumar:
> Yeah, this is ugly. Is there a way we can move our nightly from managed to self-managed cluster? If not, would you mind to put a comment on the line why we're using self-managed install type on the managed cluster? So it's more obvious to anyone who will look the code in the future that this is desired...

Tarun: I know it is ugly but we do not have consistent solution because odh-nightly is only designed for  SM installation. Regarding us of managed cluster we use this for explore testing because OSD is more stable and controlled in better than OCP. Regarding the comment i will put something but not more detailed

aloganat (approved):

jstourac (approved):

jstourac:
> Tarun: I know it is ugly but we do not have consistent solution because odh-nightly is only designed for SM installation. Regarding us of managed cluster we use this for explore testing because OSD is more stable and controlled in better than OCP. Regarding the comment i will put something but not more detailed

Thank you, Tarun. That is all I wanted :slightly_smiling_face:

PR #1407, opened 2024-04-24
Remove workaround needed for self-signed S3

apodhrad:
Since 2.8 we can specify CA bundles in DSCI


jgarciao (approved):
It is possible we'll add a dedicated test to verify configuring custom ca's using this system, as [it is still a valid configuration for pipelines](https://access.redhat.com/documentation/en-us/red_hat_openshift_ai_self-managed/2.8/html/installing_and_uninstalling_openshift_ai_self-managed/working-with-certificates_certs#using-certificates-with-data-science-pipelines_certs), but I agree on removing it now to simplify the code.

Raghul-M (approved):

jstourac (approved):

apodhrad:
Verified with these test cases
* ODS-1932
* ODS-2209

PR #1409, opened 2024-04-25
Fix image reference in DSPAv2 to a digest

apodhrad:


diegolovison (approved):

jstourac (approved):

jstourac:
Will we need to update this in our tests regularly?

jgarciao (approved):

PR #1412, opened 2024-04-29
Update the tgis runtime for rhoai_2_8_2

tarukumar:


jstourac:
@tarukumar I suppose this should be rebased against `releases/2.8.0` branch? :slightly_smiling_face:

jstourac:
For reference - this was closed in favor of #1417

PR #1418, opened 2024-05-03
[Fix] check for the CUDA version installed on image

jstourac:
Up to now we used `nvidia-smi` for the CUDA version check, but because of the [1] we should rather use the `nvcc` tool for such operation.

Also, correct version of CUDA is 12.1 in our images now, see [2].

This is a followup of #1413 [3].

[1] https://issues.redhat.com/browse/RHOAIENG-6673
[2] https://github.com/red-hat-data-services/notebooks/blob/rhoai-2.9/cuda/ubi9-python-3.9/Dockerfile#L29
[3] aae86d286bc8ed7fd8cfb9c0c468b86cf6654764

---

CI: good - failed test is unrelated issue to this PR
Screenshot from 2024-05-03 13-55-05


FedeAlonso:
Thank you, Jan!
LGTM

FedeAlonso (approved):

lugi0 (approved):

PR #1425, opened 2024-05-07
Add pyright typechecker to GitHub Actions

jiridanek:


.github/workflows/code_quality.yaml
```diff
@@ -82,6 +82,9 @@ jobs:
       - run: poetry run pyright
         if: "!cancelled()"
 
+      - run: poetry run pyright
```
kobihk, line 85: I think we don't need it since we have the same code in lines # 82-83

kobihk (changes requested):

.github/workflows/code_quality.yaml
jiridanek: 🤦🏽 thanks, I'm blind

jiridanek:
This has been already merged before, 
* https://github.com/red-hat-data-services/ods-ci/pull/1351

PR #1433, opened 2024-05-09

apodhrad:


apodhrad:
Hi @mattmahoneyrh @asanzgom this implementation works for multiline as well as single line strings

apodhrad:
We also need to merge this changes to 2.8 and 2.9

mattmahoneyrh:
Ran this PR after setting the custom CA bundle with configure-trusted-ca-bundle-dis.sh, all trusted-ca-bundle test cases passed (tag: ODS-2638) when run from localhost.

mattmahoneyrh (approved):

asanzgom (approved):

PR #1435, opened 2024-05-10
Update codeflare-sdk tests to run against latest release tag

Srihari1192:
closes https://issues.redhat.com/browse/RHOAIENG-5070


sutaakar (approved):

ChughShilpa (approved):
lgtm

jiripetrlik (approved):

sutaakar:
Makes me think, is there a need to work with Git at all? Why not to install latest version directly from pip?

Srihari1192:
> Makes me think, is there a need to work with Git at all? Why not to install latest version directly from pip?

Yes we can install either way , but for test execution we need the respective repo branch aswell

sutaakar:
ok, fine for me

PR #1436, opened 2024-05-10
Add Training operator smoke tests to DW stack

sutaakar:


ChughShilpa (dismissed):
lgtm 

Just one small change request, Instead of using `  %{ODH_NAMESPACE=redhat-ods-applications}`, we can use 
`${APPLICATIONS_NAMESPACE}` variable which fetches value from test variable file. Hence, our smoke tests run on ODH nightly as well as rhoai downstream .

https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/test-variables.yml.example#L81

Currently our smoke tests fails in https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/odh/job/autotrigger-smoke/

sutaakar:
@ChughShilpa thanks, good catch
should be fixed now

sutaakar:
This needs to be fixed in other files too, though it is out of scope for this PR.

ChughShilpa (approved):

ChughShilpa:
> @ChughShilpa thanks, good catch should be fixed now

Thanks Karel

ChughShilpa:
> This needs to be fixed in other files too, though it is out of scope for this PR.

Yes right, i will raise the PR for this

jiripetrlik (approved):

PR #1439, opened 2024-05-13

apodhrad:
This was already merged to the master branch but we need it also for 2.9 smoke tests

* master #1433
* 2.8 branch #1440


asanzgom (approved):

jstourac (approved):

jstourac:
Normally, we shouldn't backport to these branches for non-stable releases, but since we're expecting the patch release exceptionally, then this is good. Thanks!

PR #1440, opened 2024-05-13

apodhrad:
This was already merged to the master branch but we need it also for 2.8 smoke tests

* master #1433
* 2.9 branch #1439

### DIFF
--- a/ods_ci/libs/Helpers.py
+++ b/ods_ci/libs/Helpers.py
@@ -310,3 +310,13 @@ class Helpers:
             SyntaxError: EOL while scanning string literal (<string>, line 1)
         """
         return string is None or string == ""
+
+    @keyword
+    def multiline_to_oneline_string(self, multiline_string, delimeter=" "):
+        """
+        Converts a mutliline string into a oneline string with a provided delimeter.
+        Robot Framework doesn't properly handle multi-line strings and throws
+            Evaluating expression '"...".replace('', '\n')' failed:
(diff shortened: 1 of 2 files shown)


jstourac:
Has this been tested with 2.8 branch?

apodhrad:
No :(
Let's verify it on 2.8.2

jstourac:
You know, the changes LGTM and shouldn't break anything hopefully. But I'm not even sure we truly need it for 2.8 - are the tests that this fixes actually present in this branch and are being run? I haven't checked personally, so asking :slightly_smiling_face:

apodhrad:
They are part of the smoke tests, so the goal here is to make at least the smoke tests as stable as possible and reduce the work related to failure investigations for 2.8.3, 2.8.4 , ... for several months

mattmahoneyrh (approved):

apodhrad:
In 2.8 we would need to add also other commits

jstourac:
@apodhrad for 2.8, the smokes are basically always green in connected env. Do I understand correctly that you see these tests failing in disconnected?

apodhrad:
> @apodhrad for 2.8, the smokes are basically always green in connected env. Do I understand correctly that you see these tests failing in disconnected?

worse, these tests are green but they delete the ca bundles which could cause other failures, so it depends on the order of executed tests

apodhrad:
@jstourac @mattmahoneyrh I have added changes related to restoring the ca bundles + the my recent change.
I have tested that with 2.8.2.GA where the tests are green and the CA bundles were properly restored.

PR #1442, opened 2024-05-14
Fix "F841", # Local variable `response` is assigned to but never used

jiridanek:
I'm not entirely sure we really want this lint enabled, but overall, I think probably yes. Please review, thanks.


ods_ci/utils/scripts/testconfig/generateTestConfigFile.py
```diff
@@ -142,7 +142,7 @@ def get_dashboard_url():
     """
     Get dashboard url for the open data science.
     """
-    host_jsonpath = "{.spec.host}"
+    _host_jsonpath = "{.spec.host}"
```
jstourac, line 145: Does it make sense to keep this line?

ods_ci/libs/Helpers.py
```diff
@@ -99,7 +99,7 @@ def update_notification_email_address(self, cluster_name, email_address, addon_n
     @keyword
     def convert_to_hours_and_minutes(self, seconds):
         """Converts seconds in hours and minutes"""
-        m, s = divmod(int(seconds), 60)
+        m, _s = divmod(int(seconds), 60)
```
jstourac, line 102: Personally, I would slightly prefer to have just `_` in all these variables you renamed (this way it is more visible for myself that we're ignoring this variable), but I don't really insist.

jstourac (approved):
LGTM, thanks. Put two comments that can be addressed. I can live with this to be merged as is, though.

ods_ci/libs/Helpers.py
jiridanek: This is related to the previous comment. I wanted to preserve the variable name "for documentation purposes". I'll go and put just a `_` there; that's probably how this is most often done.

ods_ci/utils/scripts/testconfig/generateTestConfigFile.py
jiridanek: removed line

jstourac (dismissed):

kobihk (approved):

jstourac (approved):

PR #1447, opened 2024-05-15
Parameterize the model deployment for tgis and using ratio to match model response

tarukumar:
Parameterize the model deployment for tgis so that we can utilize same script for other runtime as well and using ratio to match model response


mwaykole:
looks good have you verified it ?

tarukumar:
> looks good have you verified it ?



> looks good have you verified it ?

yes

lugi0 (approved):

bdattoma (approved):

PR #1453, opened 2024-05-16
Exclude image digest verification on nightly builds

apodhrad:
Nightly builds can contain images with tags

### DIFF
--- a/ods_ci/tests/Tests/100__deploy/100__installation/102__post_install.robot
+++ b/ods_ci/tests/Tests/100__deploy/100__installation/102__post_install.robot
@@ -293,6 +293,7 @@ Verify All The Pods Are Using Image Digest Instead Of Tags
     ...       Tier1
     ...       ODS-2406
     ...       ExcludeOnODH
+    ...       ExcludeOnNightly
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/100__deploy/100__installation/102__post_install.robot
```diff
@@ -292,6 +292,7 @@ Verify All The Pods Are Using Image Digest Instead Of Tags
     ...       Tier1
     ...       ODS-2406
     ...       ExcludeOnODH
+    ...       ExcludeOnNightly
```
jstourac, line 295: Do we plan to use this tag? AFAIK we don't use this in Jenkins anywhere? :thinking:

jiridanek (approved):

ods_ci/tests/Tests/100__deploy/100__installation/102__post_install.robot
apodhrad: Hi @jstourac , you are right, we do not use it atm. But I plan to set up a disconnected nightly job and having such a tag would allow us to reduce the number of failed tests.

apodhrad:
Hi @jiridanek @jstourac we can close this merge request without merging.
Today I received info that nightly builds builds will be the same as RC builds once we onboard on konflux.
So, no need to to introduce such a new tag.

PR #1445, opened 2024-05-14
Allow user to provide custom-manifests to run ODS_CI on PR

vaibhavjainwiz:
Allow user to provide custom-manifests to run ODS_CI on PR.

Custom-manifest in below syntax should be define on `test-variable.yaml` file.
```
CUSTOM_MANIFESTS:
  kserve:
    manifests:
    - uri: https://github.com/vaibhavjainwiz/kserve/tarball/test_pr
      contextDir: config
      sourcePath: ''
```

Values of custom manifest could be passed using Jenkins as inline parameter(component1:repo_org:repo_name:branch_name:context_dir:source_path)
```
python3 -u ./utils/scripts/testconfig/generateTestConfigFile.py --custom_manifests=kserve:vaibhavjainwiz:kserve:test_pr:config: --components=kserve:Managed
```

PR #1458, opened 2024-05-17
[fix] disable the custom manifest support temporarily

jstourac:
Reason is that this doesn't work when empty array is generated in test-variables.yml file. This should be easy to fix, but I'm lazy to do it.

This is a followup of #1445.

Tested and job passed for me.


jiridanek (approved):

aloganat (approved):
lgtm

PR #1466, opened 2024-05-21
Handle custom_manifest empty condition

vaibhavjainwiz:


jstourac:
LGTM in general, two questions:

1. Have you verified it works with some run?
2. Have you considered of not adding any `CUSTOM_MANIFESTS` section into the generated variables file when there is no need to do so? Advantage of this approach would be that people don't need to amend their locally saved variable files they use for execution.

vaibhavjainwiz:
@jstourac 
I have tested my changes by running below rhods-smoke test.
https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods/job/rhods-smoke/5500/console

In this job, I didn't pass any parameter in CUSTOM_MANIFEST. I had cancelled this job before its completed but we could verify that Setup till RHODS setup is successfully completed.
```
Rhods Olm :: Perform and verify RHODS OLM tasks                       | PASS |
```

bdattoma (approved):

jstourac (approved):

PR #1468, opened 2024-05-21
Verify RHODS version before and after upgrade

adolfo-ab:
Refactor upgrade test to check if version after upgrade is greater than the initial version. The previous approach that checked if the pods in the operator namespace were Pending after 10s was causing the test to fail, because the pods were already running before 10s, and playing with that sleep timer might make the test flaky. I think the right way to confirm if the upgrade has been successful is to check the CSV version and that the pods in the operator namespace are running once the CSV version has been updated

rh-pre-commit.version: 2.2.0
rh-pre-commit.check-secrets: ENABLED

### DIFF
--- a/ods_ci/tests/Tests/100__deploy/120__upgrades/121__during_upgrade.robot
+++ b/ods_ci/tests/Tests/100__deploy/120__upgrades/121__during_upgrade.robot
@@ -27,17 +27,16 @@ Long Running Jupyter Notebook
     Close Browser
 
 Upgrade RHODS
-    [Documentation]    Appprove the install plan for the upgrade
+    [Documentation]    Approve the install plan for the upgrade and make sure that upgrade has completed
     [Tags]  ODS-1766
     ...     Upgrade
     ${initial_version} =  Get RHODS Version
+    ${initial_creation_date} =  Get Operator Pod Creation Date
     ${return_code}    ${output}    Run And Return Rc And Output   oc patch installplan $(oc get installplans -n ${OPERATOR_NAMESPACE} | grep -v NAME | awk '{print $1}') -n ${OPERATOR_NAMESPACE} --type='json' -p '[{"op": "replace", "path": "/spec/approved", "value": true}]'   #robocop:disable
     Should Be Equal As Integers    ${return_code}     0   msg=Error while upgrading RHODS
     Sleep  30s      reason=wait for thirty seconds until old CSV is removed and new one is ready
-    ${is_version_gt} =    Is RHODS Version Greater Than    ${initial_version}
-    IF    ${is_version_gt} == False
-        Fail    RHODS version was not greater than initial version ${initial_version}
-    END
+    RHODS Version Should Be Greater Than    ${initial_version}
+    Operator Pod Creation Date Should Be Updated    ${initial_creation_date}
     OpenShiftLibrary.Wait For Pods Status  namespace=${OPERATOR_NAMESPACE}  timeout=300
 
 TensorFlow Image Test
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/100__deploy/120__upgrades/121__during_upgrade.robot
```diff
@@ -85,3 +88,12 @@ Upgrade Test Teardown
     Log    rhods_aggregate_availability: ${resp.json()["data"]["result"][0]["value"][-1]}
     @{list_values} =    Create List    1
     Run Keyword And Warn On Failure    Should Contain    ${list_values}    ${resp.json()["data"]["result"][0]["value"][-1]}
+
+Is RHODS Version Greater Than
+    [Documentation]    Returns True if:
+    ...    - RHODS version is greater or equal than ${initial_version}
+    [Arguments]  ${initial_version}
+    ${ver} =  Get RHODS Version
+    ${ver} =  Fetch From Left  ${ver}  -
```
jstourac, line 97: Does this work properly with z.stream too? E.g. 2.8.3?

bdattoma:
>  I think the right way to confirm if the upgrade has been successful is to check the CSV version and that the pods in the operator namespace are running once the CSV version has been updated


if something is wrong with upgrade, the pods may remain up and running. So, even if the CSV is updated, the upgrade  process may still be blocked and components not upgraded.

I think one possible option we have is to check that each deployment under redhat-ods-* namespaces got rollout, in other words the pods creation date should be more recent than the time upgrade was initiated.

ods_ci/tests/Tests/100__deploy/120__upgrades/121__during_upgrade.robot
adolfo-ab: I just tested it and it works with z-streams, I could upgrade from 2.9.0 to 2.9.1 using this test. ${RHODS_VERSION} variable is X.Y.Z.

ods_ci/tests/Tests/100__deploy/120__upgrades/121__during_upgrade.robot
```diff
@@ -30,11 +30,14 @@ Upgrade RHODS
...
-    Should Be Equal As Integers    ${return_code}     0   msg=Error while upgradeing RHODS
-    Sleep  10s      reason=wait for ten second until operator goes into init state
-    ${return_code}    ${output}    Run And Return Rc And Output   oc get pod -n ${OPERATOR_NAMESPACE} -l name=rhods-operator --no-headers --output='custom-columns=STATUS:.status.phase'    #robocop:disable
-    Should Contain    ${output}    Pending
+    Should Be Equal As Integers    ${return_code}     0   msg=Error while upgrading RHODS
+    Sleep  30s      reason=wait for thirty seconds until old CSV is removed and new one is ready
+    ${is_version_gt} =    Is RHODS Version Greater Than    ${initial_version}
+    IF    ${is_version_gt} == False
+        Fail    RHODS version was not greater than initial version ${initial_version}
+    END
```
bdattoma, line 40: Maybe you could create a keyword like `RHODS Version Should Be Greater than` which fails if the comparison fails. It would be more in RobotFramework style

adolfo-ab: Done

ods_ci/tests/Tests/100__deploy/120__upgrades/121__during_upgrade.robot
```diff
@@ -30,11 +30,14 @@
...
+    ${initial_version} =  Get RHODS Version
     ${return_code}    ${output}    Run And Return Rc And Output   oc patch installplan $(oc get installplans -n ${OPERATOR_NAMESPACE} | grep -v NAME | awk '{print $1}') -n ${OPERATOR_NAMESPACE} --type='json' -p '[{"op": "replace", "path": "/spec/approved", "value": true}]'   #robocop:disable
-    Should Be Equal As Integers    ${return_code}     0   msg=Error while upgradeing RHODS
-    Sleep  10s      reason=wait for ten second until operator goes into init state
-    ${return_code}    ${output}    Run And Return Rc And Output   oc get pod -n ${OPERATOR_NAMESPACE} -l name=rhods-operator --no-headers --output='custom-columns=STATUS:.status.phase'    #robocop:disable
-    Should Contain    ${output}    Pending
+    Should Be Equal As Integers    ${return_code}     0   msg=Error while upgrading RHODS
+    Sleep  30s      reason=wait for thirty seconds until old CSV is removed and new one is ready
+    ${is_version_gt} =    Is RHODS Version Greater Than    ${initial_version}
+    IF    ${is_version_gt} == False
```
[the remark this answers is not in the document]
adolfo-ab: Done

bdattoma (approved):

jstourac (dismissed):

asanzgom (dismissed):

jstourac (approved):

bdattoma (approved):

PR #1470, opened 2024-05-21
Bumping default retry to 150 in Wait Until Operator Subscription Last Condition Is

CFSNM:


lugi0 (dismissed):

ods_ci/tests/Resources/Page/Operators/ISVs.resource
```diff
@@ -66,7 +66,8 @@ Wait Until Operator Subscription Last Condition Is    # robocop: disable
     [Documentation]    Keeps checking until the subscription status matches the expected status
     [Arguments]    ${type}    ${status}    ${reason}
     ...    ${subcription_name}    ${namespace}=openshift-operators
-    ...    ${retry}=20    ${retry_interval}=3s
```
bdattoma, line 69: instead of changing the default values (450s is an high timeout for a default IMO), could you set the parameters in the corresponding keyword calls?

CFSNM: done, added a default value of 60 retry, and using 150 in the desired calls

jstourac (approved):

bdattoma (approved):

PR #1471, opened 2024-05-21
[Jupyter] RHOAIENG-7490 update test tags - remove many from Sanity

jstourac:
This change is to reduce the execution time of the Sanity tests. Removing many tests from Sanity and keeping it only in Smoke or Tier1.

* https://issues.redhat.com/browse/RHOAIENG-7490

### DIFF
--- a/ods_ci/tests/Tests/500__jupyterhub/custom-image.robot
+++ b/ods_ci/tests/Tests/500__jupyterhub/custom-image.robot
@@ -11,7 +11,7 @@ Library          JupyterLibrary
 Library          OpenShiftLibrary
 Suite Setup      Custom Notebook Settings Suite Setup
 Suite Teardown   End Web Test
-Test Tags       JupyterHub
+Test Tags        JupyterHub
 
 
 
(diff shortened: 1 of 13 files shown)


jstourac:
> could you list the test cases which remain in Sanity?

Right now it's basically tests in these files:

```
minimal-cuda-test.robot
multiple-gpus.robot
notebook_size_verification.robot
plugin-verification.robot
test-jupyterlab-notebook.robot
test-minimal-image.robot
test-pipChanges-not-permanent.robot
test-s3-cc-fraud.robot
```

Looking on this, we may want to reduce it also by cuda and gpu tests that may also take quite a long time to run. I didn't bump into that because the example execution run didn't run on the GPU :slightly_smiling_face: 

We can do it in a separate followup PR though.

TBH - 15mins execution time is very tight... personally I would like to implement more non-UI tests to run and check the overall state of the component so that we can test more things in this timeframe...

bdattoma:
> > could you list the test cases which remain in Sanity?
> 
> Right now it's basically tests in these files:
> 
> ```
> minimal-cuda-test.robot
> multiple-gpus.robot
> notebook_size_verification.robot
> plugin-verification.robot
> test-jupyterlab-notebook.robot
> test-minimal-image.robot
> test-pipChanges-not-permanent.robot
> test-s3-cc-fraud.robot
> ```
> 
> Looking on this, we may want to reduce it also by cuda and gpu tests that may also take quite a long time to run. I didn't bump into that because the example execution run didn't run on the GPU 🙂
> 
> We can do it in a separate followup PR though.
> 
> TBH - 15mins execution time is very tight... personally I would like to implement more non-UI tests to run and check the overall state of the component so that we can test more things in this timeframe...

I'd agree with having some additional non-UI tests

jgarciao (approved):

PR #1472, opened 2024-05-21
Test Validate Argo Workflow for DSP Component

asanzgom:
rhods-ci-pr-test/2782


ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -264,7 +264,7 @@ Wait For DSCInitialization CustomResource To Be Ready
 Apply DataScienceCluster CustomResource
     [Documentation]
     [Arguments]        ${dsc_name}=${DSC_NAME}
-    ${file_path} =    Set Variable    tasks/Resources/Files/
+    ${file_path} =    Set Variable    ods_ci/tasks/Resources/Files/
```
bdattoma, line 267: this is going to break CI/CD

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -295,31 +295,31 @@ Apply DataScienceCluster CustomResource
 Create DataScienceCluster CustomResource Using Test Variables
     [Documentation]
     [Arguments]    ${dsc_name}=${DSC_NAME}
-    ${file_path} =    Set Variable    tasks/Resources/Files/
+    ${file_path} =    Set Variable    ods_ci/tasks/Resources/Files/
```
bdattoma, line 298: an improvement could be to make this a Suite variable rather than repeating it

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -295,31 +295,31 @@ Apply DataScienceCluster CustomResource
...
     Copy File    source=${file_path}dsc_template.yml    destination=${file_path}dsc_apply.yml
-    Run    sed -i 's/<dsc_name>/${dsc_name}/' ${file_path}dsc_apply.yml
+    Run    sed -i ' ' -e 's/<dsc_name>/${dsc_name}/' ${file_path}dsc_apply.yml
     FOR    ${cmp}    IN    @{COMPONENT_LIST}
         IF    $cmp not in $COMPONENTS
-            Run    sed -i 's/<${cmp}_value>/Removed/' ${file_path}dsc_apply.yml
+            Run    sed -i ' ' -e 's/<${cmp}_value>/Removed/' ${file_path}dsc_apply.yml
```
bdattoma, line 303: is it for fixing sed on MacOs?

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
```diff
@@ -0,0 +1,60 @@
...
+    ${return_code}          ${output}               Run And Return Rc And Output
+    ...                     oc apply -f ods_ci/tests/Resources/Files/argo/crd.workflows.yaml
+    Log To Console          ${output}
+    Should Be Equal As Integers
+    ...                     ${return_code}
+    ...                     0
+    ...                     msg=Error while applying the provided file
+    Open Installed Operators Page
+    Navigate to Installed Operators
+    ${is_operator_installed}                        Is Operator Installed                           ${OPERATOR_NAME}
```
bdattoma, line 28: could you try avoiding OCP UI to check operators? it's been always flaky as method

bdattoma, line 28: which operator? I guess ODH/RHOAI? could you make it clear?

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
```diff
@@ -0,0 +1,60 @@
...
+    IF    ${is_operator_installed}    Uninstall ODH Operator
+    ODH Operator Should Be Uninstalled
+    Open OperatorHub
+    Install ODH Operator
```
bdattoma, line 32: why?
I think it'd safer not to put a such test in Tier1.
Also, it says "ODH", does it do the same for RHOAI?

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
```diff
@@ -0,0 +1,60 @@
...
+    Resource Status Should Be
+    ...                     oc get DataScienceCluster default-dsc -o json | jq '.status.conditions[] | select(.type=="data-science-pipelines-operatorReady") | .status'
+    ...                     data-science-pipelines-operatorReady
+    ...                     "False"
+    ${output}    Run And Return Rc And Output
+    ...    oc delete crd -l app.kubernetes.io/part-of=data-science-pipelines-operator -l app.opendatahub.io/data-science-pipelines-operator=true --ignore-not-found
+    Log To Console          ${output}
+    ${output}               Run And Return Rc And Output
+    ...                     oc delete DataScienceCluster default-dsc
+    Log To Console          ${output}
```
bdattoma, line 47: don't we need some checks on the result of the operation? e.g., return code

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
asanzgom: correct @bdattoma

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
asanzgom: We are leaving tests on Tier1 by default, the name ODH in that keyword already existed so I would have to refactor it in order to change name

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
asanzgom: Adding checks

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
asanzgom: It works with both products, what is defined on test-variables.yaml

asanzgom: So Far it is working fine..

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
asanzgom: it's strange because on local, that part of the path was necessary, but on ci/cd is it different?

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
asanzgom: Ok I will refactor it

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
jstourac: 1. Such OS specific change shouldn't be part of such PR as this one is. It should be in a separate PR or commit at least. Otherwise someone (e.g. me) will put it back deliberately, because there is no explanation for this change and for regular Linux user it doesn't make sense.
2. If you do this change could you, please, check that this won't stop working to users with standard GNU sed? E.g. trying something similar, looks like it doesn't seem to work on my machine:
```
$ sed -i ' ' -e 's/my/Removed/' ahoj
sed: can't read  : No such file or directory
```
this seem to work on my machine:
```
$ sed -i'' -e 's/my/Removed/' ahoj
```

asanzgom: @jstourac let me try your change. It should be compatible for Mac users too, if it passes CI/CD then it is suitable for Linux, but I need to be also able to execute on local

tarukumar: it depends on the sed version. Kindly check the version

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
asanzgom: changed to work on CI/CD

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
tarukumar: Agree with Berto ..will suggest to go with cli approach ..it will cause problem between different openshift version

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
```diff
@@ -0,0 +1,62 @@
+*** Settings ***
+Documentation       Test Cases to verify Serverless installation
```
tarukumar, line 2: This seems wrong base don the TC

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
tarukumar: could you let me know what version of sed you are using ?

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
asanzgom: ok changing

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
asanzgom: `$ sw_vers -productVersion
14.5`

bdattoma: I think we should start limiting the usage of bash commands like sed where possible and where it doesn't complicate too much. 
We have `Create File From Template` keyword in RF which is comfortable for this kind of operations. Alternatively, we can use kustomize

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
bdattoma: > We are leaving tests on Tier1 by default


I'm not sure is a good reason for adding a potential destructive test in Tier1

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
bdattoma: hmm i don't think `./ods_ci/..` is  ok for CI/CI. you should leave as it was

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -265,7 +265,7 @@ Wait For DSCInitialization CustomResource To Be Ready
 Apply DataScienceCluster CustomResource
     [Documentation]
     [Arguments]        ${dsc_name}=${DSC_NAME}
-    ${file_path} =    Set Variable    tasks/Resources/Files/
+    ${file_path} =    Set Variable    /tasks/Resources/Files/
```
bdattoma, line 268: pls remove the `/`

asanzgom: removed

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
asanzgom: on which Tier do you suggest to set it?

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
jstourac: > could you let me know what version of sed you are using ?

It's not just about `sed` version but about the tool implementation... since MacOS was historically based on BSD, I believe that what Adrian is using is some kind of BSD sed implementation, whether on our Fedora machines we have a GNU sed implementation.

You can install GNU sed on MacOS too in case you want to keep this compatibility.

asanzgom: @jstourac 
```
$ gsed --version
gsed (GNU sed) 4.9
```

jstourac: You mean you have this issue when using `gsed`? Or what does your comment mean?

asanzgom: yes, it was not compatible with change it is

jstourac: Interesting, are you sure that this is the default sed that is used during the script execution? Please, be aware that script calls `sed` command, not `gsed`. Thus you need to have an alias for it or some other configuration so that `gsed` is used in the script.

Anyway, I don't want to spam this thread too much. We're slightly diverging here.

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
```diff
@@ -0,0 +1,72 @@
...
+    ${return_code}          ${output}               Run And Return Rc And Output
+    ...                     oc apply -f ./ods_ci/tests/Resources/Files/argo/crd.workflows.yaml
+    Should Be Equal As Integers
+    ...                     ${return_code}
+    ...                     0
+    ...                     msg=${output}
+    Can Uninstall RHODS Operator
+    Should Be Equal As Integers ${return_code}      0
+    Open OperatorHub
+    Install ODH Operator
```
bdattoma, line 30: Uninstall RHODS but installing ODH?

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
```diff
@@ -0,0 +1,72 @@
...
+    ...                     30s
+    ...                     Resource Status Should Be
+    ...                     oc get DataScienceCluster default-dsc -o json | jq '.status.conditions[] | select(.type=="data-science-pipelines-operatorReady") | .status'
+    ...                     CapabilityDSPv2Argo
+    ...                     "False"
+    Wait Until Keyword Succeeds
+    ...                     5 min
+    ...                     30s
+    ...                     Resource Status Should Be
+    ...                     oc get DataScienceCluster default-dsc -o json | jq '.status.conditions[] | select(.type=="data-science-pipelines-operatorReady") | .status'
```
bdattoma, line 43: maybe you could replace these lines with a keyword which makes clear what it is expecting: "Something Should Be"

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
asanzgom: This logic is being replaced with the olm install script execution on upcoming commit

bdattoma: ok but still, if we need to merge this PR before that change, I'ld like to understand: are we re-installing ODH instead of RHOAI? if yes, is that ok?

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -260,11 +261,17 @@ Apply DataScienceCluster CustomResource
...
+        ${rc}=    Set Variable    1
+        TRY
+            WHILE    ${rc} != 0    limit=2m
+                Sleep    5s
+                ${rc}    ${output}=    Run And Return Rc And Output
+                ...    oc apply -f ${file_path}dsc_apply.yml
+            END
+        EXCEPT    WHILE loop was aborted    type=start
+            Fail    msg=Error detected while applying DSC CR
+        END
```
bdattoma, line 273: cannot you use Wait Until Keyword Succeedes?

bdattoma: Also if your problem is that the CRD is not present, you can first check its existence and then apply the yaml

ods_ci/tests/Resources/RHOSi.resource
```diff
@@ -97,6 +97,7 @@ Assign Vars According To Product
     IF    "${PRODUCT}" == "RHODS"
         Set Suite Variable    ${OPERATOR_APPNAME}     "Red Hat OpenShift AI"
         Set Suite Variable    ${OPERATOR_DEPLOYMENT_NAME}    rhods-operator
+        Set Suite Variable    ${OPERATOR_NAMESPACE}    redhat-ods-operator
```
bdattoma, line 100: This variable is already in the test-variables.yaml file, there shouldn't be the need of adding it here

ods_ci/tests/Resources/RHOSi.resource
```diff
@@ -109,6 +110,7 @@ Assign Vars According To Product
     ELSE IF    "${PRODUCT}" == "ODH"
         Set Suite Variable    ${OPERATOR_APPNAME}  "Open Data Hub Operator"
         Set Suite Variable    ${OPERATOR_DEPLOYMENT_NAME}    opendatahub-operator-controller-manager
+        Set Suite Variable    ${OPERATOR_NAMESPACE}    opendatahub-operators
```
bdattoma, line 113: same here https://github.com/red-hat-data-services/ods-ci/pull/1472/files#r1783943697

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
```diff
@@ -0,0 +1,202 @@
...
+        Wait For Mutatingwebhook To Be Available
+        Apply DataScienceCluster CustomResource     ${DSC_NAME}     True    ${custom_cmp}
+    END
+
+Cleanup Olm Install Dir
+    [Documentation]    Clean up olm install repo
+    ${return_code}=    Run and Watch Command    rm -vRf ${EXECDIR}/${OLM_DIR}    timeout=10 min
+    Should Be Equal As Integers    ${return_code}    0    msg=Error while cleaning up olminstall directory
+
+Teardown Detect Pre-existing Install Of Argo Workflows
```
bdattoma, line 153: if the test fails before re-installing RHOAI, the cluster won't have RHOAI operator installed hence this teardown and the rest of the tests will fail. Could you have RHOAI re-installed if broken in the teardown as well>

ods_ci/tests/Resources/RHOSi.resource
asanzgom: Removing

ods_ci/tests/Resources/RHOSi.resource
asanzgom: Removing

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
asanzgom: This is to make the test case less prone to failures, not only about the CRD but I noticed that sometimes it failed at first try and then by manually doing the oc apply -f it worked just some seconds after, so this is to give a few more tries

ods_ci/tests/Tests/100__deploy/130__operators/150__dsp_operator/150__dsp_operator.robot
asanzgom: the problem is that teardown is always executed no matter whether the test case fails or is successful, and if it is successful the operator is already installed. The keywords that are prone to fail come after reinstalling RHOAI:
```
Detect Pre-existing Install Of Argo Workflows And Block RHOAI Install
    [Documentation]    Detect Pre-existing Install Of Argo Workflows And Block RHOAI Install
    [Tags]                  Operator                ODS-2651                Tier3
    Fetch Image Url And Update Channel
    Check Whether DSC Exists And Save Component Statuses
    Fetch Cluster Type By Domain
    IF    "${CLUSTER_TYPE}" == "selfmanaged"
        Uninstall RHODS In Self Managed Cluster
        Create Argo Workflow From Template
        Install RHODS In Self Managed Cluster Using CLI     ${CLUSTER_TYPE}     ${IMAGE_URL}

    ELSE IF    "${CLUSTER_TYPE}" == "managed"
        Uninstall RHODS In OSD
        Create Argo Workflow From Template
        Install RHODS In Managed Cluster Using CLI      ${CLUSTER_TYPE}     ${IMAGE_URL}
    END
    Restore Datasciencecluster If Existed
    Wait For Failed Conditions
    ${return_code}          ${output}               Run And Return Rc And Output
    ...                     oc delete DataScienceCluster default-dsc
    Log To Console          ${output}
    Should Be Equal As Integers
    ...                     ${return_code}
    ...                     0
    ...                     msg=Error deleting DataScienceCluster CR
    [Teardown]      Teardown Detect Pre-existing Install Of Argo Workflows
```

bdattoma: cannot you check if RHOAI is installed in the teardown to decide if to install or not?

ods_ci/tests/Resources/Page/OCPDashboard/OperatorHub/InstallODH.robot
```diff
@@ -17,7 +17,7 @@ Open OperatorHub
     OperatorHub Should Be Open
 
 Install ODH Operator
-    Install Operator    opendatahub
+    Install Operator    ${OPERATOR_DEPLOYMENT_NAME}
```
CFSNM, line 20: shouldnt this be ${OPERATOR_NAME}, as we have in the Uninstall ODH Operator keyword?

PR #1474, opened 2024-05-22
Add option in fetch_new_test.py to automatically fetch previous release branch

bdattoma:
In order to ease the integration in our CI/CI, this PR adds the option `--ref2-auto` and `--selector-attribute ` allowing the script to automatically select the previous releases branch of ods-ci wrt `--ref1`

Example of usage: 
```
python3 ods_ci/utils/scripts/fetch_tests.py --test-repo git@github.com:red-hat-data-services/ods-ci.git --ref1 releases/2.8.0 --ref2-auto true --selector-attribute creatordate -A new-arg-file.txt
```
```
python3 ods_ci/utils/scripts/fetch_tests.py --test-repo git@github.com:red-hat-data-services/ods-ci.git --ref1 master  --ref2-auto true --selector-attribute creatordate -A new-arg-file.txt
```
Output: 
```
---| Computing differences |----
Done. Found 30 new tests in releases/2.8.0 which were not present in origin/releases/2.7.0
```
```
---| Computing differences |----
Done. Found 14 new tests in master which were not present in origin/releases/2.9.0
```

### DIFF
--- a/ods_ci/utils/scripts/fetch_tests.py
+++ b/ods_ci/utils/scripts/fetch_tests.py
@@ -63,18 +63,16 @@ def get_branch(ref_to_exclude, selector_attribute):
     List the remote branches and sort by last commit date (ASC order), exclude $ref_to_exclude and get latest
     """
     ref_to_exclude_esc = ref_to_exclude.replace("/", r"\/")
-    if "master" in ref_to_exclude or "main" in ref_to_exclude:
-        cmd = f"git branch -r --sort={selector_attribute} | grep releases/"
-    else:
-        cmd = rf"git branch -r --sort={selector_attribute} | grep releases/  | sed  's/.*{ref_to_exclude_esc}$/current/g' |  grep -zPo '[\S\s]+(?=current)'"
+    cmd = f"git branch -r --sort={selector_attribute} | grep releases/"
+    if "master" not in ref_to_exclude and "main" not in ref_to_exclude:
+        cmd += rf" | sed  's/.*{ref_to_exclude_esc}$/current/g' |  grep -zPo '[\S\s]+(?=current)'"
     ret = execute_command(cmd)
     branches = ret.split(" ")
-    ret = branches[-1].split("\x00")[0].strip().replace("\n", "")
-    if ret != "":
-        print(f"Done. {ret} branch selected as ref_2")
-        return ret
-    else:
+    branch = branches[-1].split("\x00")[0].strip().replace("\n", "")
+    if not branch or "fatal:" in branch:
         raise Exception("Failed to auto-selecting ref_2 branch.")
+    print(f"Done. {branch} branch selected as ref_2")
+    return branch
 
 
 def extract_test_cases_from_ref(repo_local_path, ref, auto=False, selector_attribute=None, ref_to_exclude=None):
--- a/ods_ci/utils/scripts/fetch_tests.py
+++ b/ods_ci/utils/scripts/fetch_tests.py
@@ -25,7 +25,7 @@ def get_repository(test_repo):
     if "http" in test_repo or "git@" in test_repo:
         print("Cloning repo ", test_repo)
         cloned = True
-        ret = execute_command(f"git clone {test_repo} {repo_local_path}")
+        execute_command(f"git clone {test_repo} {repo_local_path}")
     elif not os.path.exists(test_repo):
         raise FileNotFoundError(f"local path {test_repo} was not found")
     else:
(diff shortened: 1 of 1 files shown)


ods_ci/utils/scripts/fetch_tests.py
```diff
@@ -20,39 +21,58 @@ def get_repository(test_repo):
...
-def extract_test_cases_from_ref(repo_local_path, ref):
+def get_branch(ref_to_exclude, selector_attribute):
+    """
+    List the remote branches and sort by last commit date (ASC order), exclude $ref_to_exclude and get latest
+    """
+    ref_to_exclude_esc = ref_to_exclude.replace("/", r"\/")
+    if "master" in ref_to_exclude or "main" in ref_to_exclude:
+        cmd = f"git branch -r --sort={selector_attribute} | grep releases/"
+    else:
+        cmd = rf"git branch -r --sort={selector_attribute} | grep releases/  | sed  's/.*{ref_to_exclude_esc}$/current/g' |  grep -zPo '[\S\s]+(?=current)'"
```
kobihk, line 69: because the the beginning of the cmd is the same maybe you can do it like:
```suggestion
    cmd = f"git branch -r --sort={selector_attribute} | grep releases/"
    if "master" not in ref_to_exclude or "main" not in ref_to_exclude:
        cmd += rf" | sed  's/.*{ref_to_exclude_esc}$/current/g' |  grep -zPo '[\S\s]+(?=current)'"
```

ods_ci/utils/scripts/fetch_tests.py
```diff
@@ -20,39 +21,58 @@ def get_repository(test_repo):
...
+    ret = execute_command(cmd)
+    branches = ret.split(" ")
+    ret = branches[-1].split("\x00")[0].strip().replace("\n", "")
```
kobihk, line 72: maybe "branch" instead of generic "ret"
could you please give an example where we need the strip and replace?  
```suggestion
    branch = branches[-1].split("\x00")[0].strip().replace("\n", "")
```

ods_ci/utils/scripts/fetch_tests.py
```diff
@@ -20,39 +21,58 @@ def get_repository(test_repo):
...
+    if ret != "":
```
kobihk, line 73: if the string is not empty we can check like:
```suggestion
    if ret:
```

ods_ci/utils/scripts/fetch_tests.py
```diff
@@ -20,39 +21,58 @@ def get_repository(test_repo):
...
+        print(f"Done. {ret} branch selected as ref_2")
+        return ret
+    else:
+        raise Exception("Failed to auto-selecting ref_2 branch.")
```
kobihk, line 77: we will get to this line only if the ret string is empty 
otherwise, the return from the last line will occur 
```suggestion
    raise Exception("Failed to auto-selecting ref_2 branch.")
```

kobihk (changes requested):

ods_ci/utils/scripts/fetch_tests.py
bdattoma: it helps knowing ret is expected to be a string. But ok will change

ods_ci/utils/scripts/fetch_tests.py
bdattoma: without replace: 
image

Strip maybe can be avoided here, it was necessary in a previous version of the script.

In any case, without both it works, but output goes to new line when unnecessary

ods_ci/utils/scripts/fetch_tests.py
bdattoma: I changed it slightly:

```code
if not branch or "fatal:" in branch:
        raise Exception("Failed to auto-selecting ref_2 branch.")
 print(f"Done. {branch} branch selected as ref_2")
 return branch
```

It's still wrong to do a error check like this one, we should rather check the return code of the command. But the function we have "execute_command" doesn-t return it and I'd prefer not enlarging the scope of the PR so much.

ods_ci/utils/scripts/fetch_tests.py
bdattoma: done

ods_ci/utils/scripts/fetch_tests.py
bdattoma: done

jiridanek (approved):

kobihk (approved):

PR #1477, opened 2024-05-23
Use ${APPLICATIONS_NAMESPACE} variable

ChughShilpa:
Use ${APPLICATIONS_NAMESPACE} variable so that test runs on both ODH build and RHOAI build


sutaakar (approved):
Thanks

jiripetrlik (approved):

PR #1478, opened 2024-05-23
RHOAIENG-7580 - Add Pod Ready Check

mattmahoneyrh:
The DSC component test cases check for the existence of given component pods. However, the pod may exist but is in an error state such as due to a pull error. Thus, validate not only that the pod exists but that it is also in a Ready state.

Remove Training Operator from Tier1 Tag


VaishnaviHire (approved):
/lgtm

PR #1486, opened 2024-05-28
Add new granite and intructlab merlinite-7b-lab model for tgis and vllm

tarukumar:


ods_ci/tests/Resources/Files/llm/model_expected_responses.json
```diff
@@ -42,6 +42,22 @@
...
+                    }
+                },
+                "merlinite-7b-lab": {
+                    "response_tokens": 20,
+                    "response_text": "\n\nWater boils at 100 degrees Celsius or 212",
+                    "streamed_response_text": "{ 'inputTokenCount': 8 } { 'generatedTokenCount': 2, 'text': '\n' } { 'generatedTokenCount': 3, 'text': '\n' } { 'generatedTokenCount': 4, 'text': 'Wate' } { 'generatedTokenCount': 5, 'text': 'r b' } { 'generatedTokenCount': 6, 'text': 'oil' } { 'generatedTokenCount': 7, 'text': 's a' } { 'generatedTokenCount': 8, 'text': 't' } { 'generatedTokenCount': 9, 'text': ' ' } { 'generatedTokenCount': 10, 'text': '1' } { 'generatedTokenCount': 11, 'text': '0' } { 'generatedTokenCount': 12, 'text': '0 degree' } { 'generatedTokenCount': 13, 'text': 's Ce' } { 'generatedTokenCount': 14, 'text': 'ls' } { 'generatedTokenCount': 15, 'text': 'iu' } { 'generatedTokenCount': 16, 'text': 's o' } { 'generatedTokenCount': 17, 'text': 'r' } { 'generatedTokenCount': 18, 'text': ' ' } { 'generatedTokenCount': 19, 'text': '2' } { 'generatedTokenCount': 20, 'text': '12', 'stopReason': 'MAX_TOKENS' }",
```
bdattoma, line 57: we should stop using these 2 values and move them inside the tgis-runtime key like for the tokenize response text

tarukumar: sure ..that could be enhancement given that time frame, i would prefer to keep this as of now and change it later

bdattoma: we already have the support for not using these fields, it's not an enhancement you need to implement. 
But if you prefer keeping it as it is, I will approve

tarukumar: I know we have the support for this..what i don't want to do at this moment to change the all things because my plan is to do in another PR which will have only this change nothing else

bdattoma (approved):

mwaykole (approved):

PR #1507, opened 2024-06-06
Modify Pipeline UI test in Smoke to use pip_index_url (making it compatible with disconnected clusters)

jgarciao:
Implements [RHOAIENG-3473](https://issues.redhat.com/browse/RHOAIENG-3473):
- Modifies Pipelines UI test in Smoke to use pip_index_url and pip_trusted_host. The values should be available in the global variables PIP_INDEX_URL and PIP_TRUSTED_HOST (for connected and disconnected clusters).
- The test will:
  -  Create a custom ConfigMap in the data science project containing the values for pip_index_url and pip_trusted_host
  -  Run iris_pipeline_pip_index_url_compiled.yaml  sample file, which reads the ConfigMap (using kfp-kubernetes, new dependency) and injects the values as env vars in the pipeline tasks. 
   -  Note:  I had to manually modify iris_pipeline_pip_index_url_compiled.yaml in order to use PIP_TRUSTED_HOST, as currently the kfp 2.7.0 is using pip_index_url for both values 
- Tested successfully in disconnect/job/psi/job/rhods-smoke/10/console


pyproject.toml
```diff
@@ -30,12 +30,14 @@ pexpect = "^4.8.0"
 python-openstackclient = "^6.2.0"
 awscli = "^1.27.100"
 kfp = "==2.7.0"
+kfp-kubernetes = "^1.2.0"
```
diegolovison, line 33: Why this was added? What problem we solve with this dependency ?

jgarciao: It's used by iris_pipeline_disconnected.py to read values from the ConfigMap and store them in the task as env var (I've added now in the description of the PR)

diegolovison (dismissed):
Have tested this PR using a notebook and manually adding the configmap.

ods_ci/tests/Resources/OCP.resource
```diff
@@ -71,6 +71,14 @@ Is RHODS Self-Managed
     ${rc}=    Run And Return Rc    oc get catalogsource addon-managed-odh-catalog -n ${OPERATOR_NAMESPACE}
     Run Keyword And Return    Run Keyword And Return Status      Should Not Be Equal As Numbers    ${rc}    ${0}
 
+Is Disconnected Cluster
```
apodhrad, line 74: Hi @jgarciao I would rather use a different terminology, for example `Is PIP Mirror Available` from 2 reasons
1. Testing with PIP miror is possible also in an online env
2. Deciding if a cluster is disconnected just according to presence of PIP vars is risky IMHO

ods_ci/tests/Tests/400__ods_dashboard/430__data_science_pipelines/434__data-science-pipelines-ui.robot
```diff
@@ -150,3 +153,17 @@ Verify Data Science Parameter From A Duplicated Run Are The Same From The Compil
     # look for spec.params inside ${PIPELINE_TEST_FILEPATH} source code
     Should Contain    ${input_parameters}    model_obc
     Should Contain    ${input_parameters}    iris-model
+
+Configure Test For Disconnected Cluster
```
apodhrad, line 157: In fact, we do not configure it specifically for a disconnected env - we just set PIP mirror.

jgarciao: I've refactored the code to just talk about pip_index_url, which will be used for connected and disconnected clusters

ods_ci/tests/Resources/OCP.resource
jgarciao: I've refactored the code to just talk about pip_index_url, which will be used for connected and disconnected clusters

jgarciao:
Adding label do-not-merge until PIP_INDEX_URL and PIP_TRUSTED_HOST are configured in Jenkins

apodhrad (approved):

diegolovison (approved):

PR #1508, opened 2024-06-06
[fix] installation of servicemesh dependant operator on managed cluster

jstourac:
This all comes from this issue [1]. Basically, we need to narrow down
the name of subscription we create for servicemesh operator so that it
matches what rhods-operator expects. Even though it's not good how the
rhoai-operator detects this.

Also, we should install Authorino operator as a dependency operator.

* [1] https://issues.redhat.com/browse/RHOAIENG-6692


aloganat (approved):
LGTM , verified the changes by executing from jenkins.

jiridanek (approved):

bdattoma (approved):

PR #1509, opened 2024-06-06
RHOAIENG-7298 - New DSC Component Dashboard and Datasciencepipelines Test Cases

mattmahoneyrh:
Adding DSC component Dashboard and Datasciencepipelines Test Cases. 

Extended resource check timeouts.

Improved pod ready check to include all pods with given label-selector.

### DIFF
--- a/ods_ci/tests/Resources/RHOSi.resource
+++ b/ods_ci/tests/Resources/RHOSi.resource
@@ -93,10 +93,14 @@ Assign Vars According To Product
         Set Suite Variable    ${OPERATOR_DEPLOYMENT_NAME}    rhods-operator
         Set Suite Variable    ${OPERATOR_LABEL_SELECTOR}    name=rhods-operator
         Set Suite Variable    ${AUTHORINO_CR_NS}    redhat-ods-applications-auth-provider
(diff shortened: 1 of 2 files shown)


mattmahoneyrh:
Jenkins runs passed RHOAI 2.10 and ODH 2.12.1.

zdtsw (approved):

CFSNM (approved):
LGTM

bdattoma:
lgtm. Question: what would happen when setting a component to Unmanaged when the component is not supposed to accept that value? 
@zdtsw is there a unit testing covering this case maybe?

bdattoma (approved):

zdtsw:
> lgtm. Question: what would happen when setting a component to Unmanaged when the component is not supposed to accept that value? @zdtsw is there a unit testing covering this case maybe?

I do not think we enabled on component level user can set value to unmanaged, only the serverless servicemesh and trustcca  support this value. 
in short, user wont allowed to set it, get rejected by patch or create DSC CR with it

mattmahoneyrh:
@bdattoma  an attempt to set the managementState to an unsupported value would be disallowed/fail, with a message indicating such as well as the supported values.

Can add negative tests. Good suggestion.

PR #1511, opened 2024-06-07
Add support for embedded endpoint for vllm

tarukumar:
The model which support generation text are not supported for embedding.  If model architectures  is a XXXForCausalLM. CausalLM means that it generates text. It should not be used for embedding.  If model architectures  is a XXModel. This means that the model just generates embeddings. It should be used for embeddings.


ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/422__model_serving_llm_models.robot
```diff
@@ -593,6 +593,47 @@ Verify User Can Serve And Query A ibm-granite/granite-8b-code-base Model
...
+    ...    overlays=${overlays}
+    Deploy Model Via CLI    isvc_filepath=${INFERENCESERVICE_FILLED_FILEPATH}
+    ...    namespace=${test_namespace}
+    Wait For Model KServe Deployment To Be Ready    label_selector=serving.kserve.io/inferenceservice=${model_name}
+    ...    namespace=${test_namespace}    runtime=${RUNTIME_NAME}    timeout=900s
+    ${pod_name}=  Get Pod Name    namespace=${test_namespace}    label_selector=serving.kserve.io/inferenceservice=${model_name}
+    Run Keyword If    "${KSERVE_MODE}"=="RawDeployment"
+    ...    Start Port-forwarding    namespace=${test_namespace}    pod_name=${pod_name}
+    IF     "${RUNTIME_NAME}" == "tgis-runtime" or "${KSERVE_MODE}" == "RawDeployment"
+            Skip   msg=Embedding endpoint is not supported for tgis as well as model architectures with "XXModel"
```
lugi0, line 625: What's the point of doing this? Wouldn't it make more sense to skip the test entirely if it's RawDeployment or using tgis?
If I understand the goal of this correctly it's to test the embeddings endpoint, so IMHO there's no benefit to deploying the model regardless and then skipping the query if it doesn't support the endpoint under test

tarukumar: I included this step to ensure smooth model deployment, whether it's in raw mode or otherwise. That's why I incorporated the skip part into the query section rather than at the initial stage.

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/422__model_serving_llm_models.robot
```diff
@@ -593,6 +593,47 @@ Verify User Can Serve And Query A ibm-granite/granite-8b-code-base Model
     ...    AND
     ...    Run Keyword If    "${KSERVE_MODE}"=="RawDeployment"    Terminate Process    llm-query-process    kill=true
 
+Verify User Can Serve And Query A intfloat/e5-mistral-7b-instruct Model
+    [Documentation]    Basic tests for preparing, deploying and querying a LLM model
+    ...                using Kserve using TGIS standalone or vllm runtime
+    [Tags]    RHOAIENG-768912    VLLM
```
bdattoma, line 599: you meant to say "RHOAIENG-7689"?

tarukumar: done

bdattoma (approved):

mwaykole (approved):

aloganat (approved):

PR #1513, opened 2024-06-07
Add support for tgi metricesfor vllm

mwaykole:


mwaykole:
Screenshot from 2024-06-08 17-51-31

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/vllm/426__model_serving_vllm_metrics.robot
```diff
@@ -89,7 +103,7 @@ Verify Vllm Metrics Are Present
     Set Suite Variable    ${token}
     Metrics Should Exist In UserWorkloadMonitoring    ${thanos_url}    ${token}    ${SEARCH_METRICS}
 
-Verify Vllm Metrics Values Match Between UWM And Endpoint
+Verify Vllm And tgi Metrics Values Match Between UWM And Endpoint
```
lugi0, line 106: This does not work - Line 98/112 fetches the metrics to test for by looking at those that start with `vllm:`, so TGI metrics are not being tested at all here.
You will need to either change the logic of the `Get Vllm Metrics And Values` in Helpers.py or develop an equivalent keyword for TGI metrics

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/vllm/426__model_serving_vllm_metrics.robot
```diff
@@ -52,10 +52,24 @@ ${TEST_NS}=                   vllm-gpt2
...
+...                           tgi_request_generated_tokens_bucket
+...                           tgi_request_generated_tokens_count
+...                           tgi_request_generated_tokens_sum
 
 *** Test Cases ***
-Verify User Can Deploy A Model With Vllm Via CLI
+Verify User Can Deploy A Model With Vllm And tgi Via CLI
     [Documentation]    Deploy a model (gpt2) using the vllm runtime and confirm that it's running
     [Tags]    Tier1    Sanity    Resources-GPU    RHOAIENG-6264   VLLM
     ${rc}    ${out}=    Run And Return Rc And Output    oc apply -f ${DL_POD_FILEPATH}
```
lugi0, line 75: This test is not using tgi to deploy the model, it's using vLLM, please revert the name of this TC

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/vllm/426__model_serving_vllm_metrics.robot
```diff
@@ -75,8 +89,8 @@ Verify User Can Deploy A Model With Vllm Via CLI
     ...    inference_type=chat-completions    n_times=3    query_idx=8
     ...    namespace=${TEST_NS}    string_check_only=${TRUE}    validate_response=${FALSE}
 
-Verify Vllm Metrics Are Present
-    [Documentation]    Confirm vLLM metrics are exposed in OpenShift metrics
+Verify Vllm And tgi Metrics Are Present
+    [Documentation]    Confirm vLLM and tgi metrics are exposed in OpenShift metrics
```
lugi0, line 93: IMHO you should create a separate test to validate that TGI metrics are present - as far as I understand you've already encountered the issue where only vLLM ones would get exposed, so it'd be better to separate the two and have clear reporting on which ones (if any) are failing to get exposed.

lugi0 (changes requested):
Test currently does not validate TGI metrics' values, and I would suggest separating the testing of TGI and vLLM metrics in two different TCs for better visibility into their behaviour

ods_ci/tests/Tests/400__ods_dashboard/420__model_serving/LLMs/vllm/426__model_serving_vllm_metrics.robot
mwaykole: ack

PR #1520, opened 2024-06-12
fix mktemp on macOS

lugi0:
The creation of a temp dir in the test-output folder for ods-ci appears to be broken in macOS, causing the `run_robot_test.sh` script to fail when run on a vanilla system using all defaults.
This variation of the command will create a directory called `ods-ci-$(date +%Y-%m-%d-%H-%M)-XXXXXXXXXX)` within the TEST_ARTIFACT_DIR directory (i.e. by default `ods_ci/test-output/`).

### DIFF
--- a/ods_ci/run_robot_test.sh
+++ b/ods_ci/run_robot_test.sh
@@ -362,7 +362,7 @@ if ! ${SUBFOLDER}; then
   case "$(uname -s)" in
     Darwin)
         # shellcheck disable=SC2046
-        TEST_ARTIFACT_DIR=$(mktemp -d  "${TEST_ARTIFACT_DIR}" -t "${TEST_ARTIFACT_DIR}"/ods-ci-$(date +%Y-%m-%d-%H-%M)-XXXXXXXXXX)
+        TEST_ARTIFACT_DIR=$(mktemp -d  -t ods-ci-$(date +%Y-%m-%d-%H-%M)-XXXXXXXXXX) -p "${TEST_ARTIFACT_DIR}"
          ;;
     Linux)
         # shellcheck disable=SC2046


ods_ci/run_robot_test.sh
```diff
@@ -362,7 +362,7 @@ if ! ${SUBFOLDER}; then
   case "$(uname -s)" in
     Darwin)
         # shellcheck disable=SC2046
-        TEST_ARTIFACT_DIR=$(mktemp -d  "${TEST_ARTIFACT_DIR}" -t "${TEST_ARTIFACT_DIR}"/ods-ci-$(date +%Y-%m-%d-%H-%M)-XXXXXXXXXX)
+        TEST_ARTIFACT_DIR=$(mktemp -d  -t ods-ci-$(date +%Y-%m-%d-%H-%M)-XXXXXXXXXX) -p "${TEST_ARTIFACT_DIR}"
```
jstourac, line 365: Since it is same as below, could you, please, at least unify the lines so it's obvious it's same or you can fallthrough (I would keep the `case` in case we'll realize that we need to differentate in the future again).

FedeAlonso (approved):
LGTM

ods_ci/run_robot_test.sh
lugi0: Good catch, did not notice that 😄

jstourac (approved):
Hopefully this won't break it for some other MacOS users with older systems since I don't understand how it worked for them fine :upside_down_face: 

Anyway, LGTM now, thank!

asanzgom:
I tried it on Mac It does not throw any error, but does not either create a directory called `ods-ci-$(date +%Y-%m-%d-%H-%M)-XXXXXXXXXX) `within the TEST_ARTIFACT_DIR directory (i.e. by default `ods_ci/test-output/)`
After triggering Test Case with changes on run_robot_tests.sh:
```
$ ls -lu /Users/asanzgom/Work/ods-ci/ods_ci/ods_ci/test-output/
total 1272
-rw-r--r-- 1 asanzgom staff 272741 12 jun 12:21 log.html
-rw-r--r-- 1 asanzgom staff 127930 12 jun 12:21 output.xml
-rw-r--r-- 1 asanzgom staff 232884 12 jun 12:21 test_report.html
-rw-r--r-- 1 asanzgom staff   443 12 jun 12:21 xunit_test_result.xml
```

jiridanek (approved):

lugi0:
@asanzgom Where are you launching the script from? The root of the repo or from its `ods_ci` subfolder?
From your logs I can see an `ods_ci` too many in the path
`/Users/asanzgom/Work/ods-ci/ods_ci/ods_ci/test-output/`

asanzgom (approved):

asanzgom:
> @asanzgom Where are you launching the script from? The root of the repo or from its `ods_ci` subfolder? From your logs I can see an `ods_ci` too many in the path `/Users/asanzgom/Work/ods-ci/ods_ci/ods_ci/test-output/`

Sorry for the late reply @lugi0 I always triggerthe tests on `ods-ci/ods_ci`

jiridanek:
@coderabbitai full review

PR #1528, opened 2024-06-13
fix: use activeModelState to wait ISVC status

FedeAlonso:
We are having some timeout issues with de deployment of models (e.g.: autotrigger-smoke/116)

After some discussion, we thought that the best way to fix it is by waiting for `.status.modelStatus.states.activeModelState == Loaded`

image

### DIFF
--- a/ods_ci/tests/Resources/CLI/ModelServing/llm.resource
+++ b/ods_ci/tests/Resources/CLI/ModelServing/llm.resource
@@ -824,7 +824,10 @@ Wait For Model KServe Deployment To Be Ready
     ...                the deployment has the expected pods and containers
     [Arguments]    ${label_selector}    ${namespace}    ${runtime}
     ...    ${timeout}=300s    ${exp_replicas}=${1}
-    Wait For Pods To Be Ready    label_selector=${label_selector}    namespace=${namespace}    timeout=${timeout}
+    # Get the model name from the label_selector variable
+    ${splitted_str}    Split String    ${label_selector}    =
+    ${model_name}    Set Variable    ${splitted_str}[1]
+    Wait For ISVC To Be Loaded     ${model_name}     ${namespace}
     Verify Model KServe Deployment    ${label_selector}    ${namespace}    ${runtime}    ${exp_replicas}
 
 Verify Model KServe Deployment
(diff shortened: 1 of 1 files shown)


asanzgom (approved):

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
```diff
@@ -857,3 +860,25 @@ Verify Model KServe Deployment
...
     ...    ${n_containers}  ${containers}
+
+Wait For ISVC To Be Loaded
+    [Documentation]    Waits for the ISVC to be Loaded (i.e., Success status in the grid)
+    [Arguments]    ${isvc_name}    ${namespace}    ${retries}=36
+    ${rc}    ${isvc_status}=    Run And Return Rc And Output
+    ...    oc get isvc ${isvc_name} -n ${namespace} -o=jsonpath='{.status.modelStatus.states.activeModelState}'
+    ${condition1}=    Evaluate    "${isvc_status}"=="${EMPTY}"
+    ${condition2}=    Evaluate    ${rc}==0
+    ${condition3}=    Evaluate    ${retries} > 0
```
jstourac, line 871: Why to have these 4 lines duplicated here before the while branch? Wouldn't be better to delete these 4 lines and put the `Sleep    5s` call at the end of the while loop?

Also, isn't it possible to utilize `oc wait` for this somehow :thinking: (don't insist on this one, but would be even nicer, I believe)

FedeAlonso: I have modified the code in order to not have duplicities.
I'm with you that this one is not an elegant and smart solution, but also I don't think it's possible to combine "oc wait" with the output of a jsonpath.

asanzgom (approved):

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
```diff
@@ -857,3 +860,23 @@ Verify Model KServe Deployment
     ${n_containers}=    Get Length    ${containers}
     Run Keyword And Continue On Failure    Verify Deployment    ${model_pods}  ${exp_replicas}
     ...    ${n_containers}  ${containers}
+
+Wait For ISVC To Be Loaded
```
jgarciao, line 864: I think that if you combine the keyword `Wait Until Keyword Succeeds` with a new keyword `ISVC Should Be Loaded` the code could be much simpler

Example:

```
Wait For ISVC To Be Loaded
    [Documentation]    Waits for the ISVC to be Loaded (i.e., Success status in the grid)
    [Arguments]    ${isvc_name}    ${namespace}    ${retries}=36
    Wait Until Keyword Succeeds   retry=${retries}    5s    ISVC Should Be Loaded    ${isvc_name}    ${namespace}
```

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
jstourac: > I have modified the code in order to not have duplicities.

Thank you.

> but also I don't think it's possible to combine "oc wait" with the output of a jsonpath.

After very brief check of oc wait help, something like this could work:
```
oc wait --for=jsonpath='{.status.modelStatus.states.activeModelState}' isvc ${isvc_name} -n ${namespace} --timeout=180s
```

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
jstourac: Hehe, I knew there was something like this but I completely forgot the name of this, I was searching for it already for this PR to comment, but failed: #1518.

manosnoam (approved):
lgtm

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
FedeAlonso: Done in 0030dd0
Thanks @jgarciao and @jstourac  for the recomendations :)

jstourac (dismissed):

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
jstourac: @FedeAlonso out of curiosity - this above didn't work?

jgarciao (approved):

asanzgom (approved):

PR #1538, opened 2024-06-17
fix: wait pods to be Ready in KServe deployment

FedeAlonso:
Wait the pods to be in Ready Status during the deployment of a KServe model, before waiting for the ISVC to be in Loaded Status.
Also increase the waiting in order to be ~5 mins.
image

### DIFF
--- a/ods_ci/tests/Resources/CLI/ModelServing/llm.resource
+++ b/ods_ci/tests/Resources/CLI/ModelServing/llm.resource
@@ -824,6 +824,7 @@ Wait For Model KServe Deployment To Be Ready
     ...                the deployment has the expected pods and containers
     [Arguments]    ${label_selector}    ${namespace}    ${runtime}
     ...    ${timeout}=300s    ${exp_replicas}=${1}
+    Wait For Pods To Be Ready    label_selector=${label_selector}    namespace=${namespace}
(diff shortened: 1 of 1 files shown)


CFSNM (approved):

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
```diff
@@ -824,6 +824,7 @@ Wait For Model KServe Deployment To Be Ready
     ...                the deployment has the expected pods and containers
     [Arguments]    ${label_selector}    ${namespace}    ${runtime}
     ...    ${timeout}=300s    ${exp_replicas}=${1}
+    Wait For Pods To Be Ready    label_selector=${label_selector}    namespace=${namespace}
```
lugi0, line 827: If we don't specify a timeout here it's going to wait for 10 minutes, is that ok? 
IMHO yes so approving :D

lugi0 (approved):

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
FedeAlonso: image
The wait is active (not waiting 10 minutes and then checking)
So it's ok for me too :)

PR #1557, opened 2024-06-24
Add sleep to ovms test, skip fail if UWM configmaps already present

lugi0:


mwaykole (approved):

ods_ci/tests/Resources/Common.robot
```diff
@@ -482,14 +482,24 @@ Enable User Workload Monitoring
     [Documentation]    Enable User Workload Monitoring for the cluster for user-defined-projects
     ${return_code}    ${output}    Run And Return Rc And Output   oc apply -f ${UWM_ENABLE_FILEPATH}
     Log To Console    ${output}
-    Should Be Equal As Integers    ${return_code}     0   msg=Error while applying the provided file
+    IF    "already exists" in ${output}
+        Log    configmap already existed on the cluster, continuing
+        RETURN
+    ELSE
+        Should Be Equal As Integers    ${return_code}     0   msg=Error while applying the provided file
+    END
```
bdattoma, line 490: the point is that "oc apply" should not fail if the CM already exist, that's weird.

lugi0: You can take a look at https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/rhods/job/rhods-smoke/5934/TestReport/ , it gets an error from the server and has RC 1 so the keyword fails

bdattoma: yes, I saw that. Strange that it never happened before

bdattoma (approved):

ods_ci/tests/Resources/Common.robot
bdattoma: Ah, could it be that `Should Be Equal As Integer` was recently added or some tests running before it in the queue now create the same CM?

jiridanek (approved):

ods_ci/tests/Resources/Common.robot
lugi0: From the git logs it seems that these keywords were untouched since last November, so my only guess is that now we hit a different scenario where UWM is already enabled when these get called

PR #1563, opened 2024-06-26
Add support for ovms cli test-case

mwaykole:
can be used 
sh run_robot_test.sh --skip-oclogin true --extra-robot-args '-i RHOAIENG-ovns --variable RUNTIME_NAME:ovns-runtime  --variable KSERVE_MODE:RawDeployment --variable PROTOCOL:http --variable MODEL_FORMAT:onnx'


mwaykole:
Screenshot from 2024-06-27 10-42-03

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
```diff
@@ -0,0 +1,109 @@
...
+    ${test_namespace}=  Set Variable     ovns-test-dir
+    ${model_name}=  Set Variable    test-dir
+    Setup Test Variables    model_name=test-dir    use_pvc=${USE_PVC}    use_gpu=${USE_GPU}
+    ...    kserve_mode=${KSERVE_MODE}
+    ${EXPECTED_INFERENCE_SECURED_OUTPUT}  Set Variable    {"model_name":"${MODEL_NAME}__isvc-83d6fab7bd","model_version":"1","outputs":[{"name":"Plus214_Output_0","datatype":"FP32","shape":[1,10],"data":[-8.233053,-7.7497034,-3.4236815,12.3630295,-12.079103,17.266596,-10.570976,0.7130762,3.321715,1.3621228]}]}  #robocop: disable
+
+    Set Project And Runtime    runtime=${RUNTIME_NAME}     protocol=${PROTOCOL}     namespace=${test_namespace}
+    ...    download_in_pvc=${DOWNLOAD_IN_PVC}    model_name=${model_name}
+    ...    storage_size=10Gi
+    ${requests}=    Create Dictionary    memory=10Gi
```
lugi0, line 40: The onnx model does not need 10Gi of memory

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
```diff
@@ -0,0 +1,109 @@
...
+    ...    namespace=${test_namespace}
+
+    ${pod_name}=  Get Pod Name    namespace=${test_namespace}    label_selector=serving.kserve.io/inferenceservice=${model_name}
+
+    Run Keyword If    "${KSERVE_MODE}"=="RawDeployment"
+    ...    Start Port-forwarding    namespace=${test_namespace}    pod_name=${pod_name}  local_port=8888
+    ...    remote_port=8888    process_alias=ovns-process
+
+    Run Keyword And Continue On Failure    Verify OVNS Model Inference Via CLI    ${MODEL_NAME}    ${INFERENCE_INPUT}
+    ...    ${EXPECTED_INFERENCE_SECURED_OUTPUT}    token_auth=${FALSE}  test_namespace=${test_namespace}    kserve_mode=${KSERVE_MODE}
```
lugi0, line 59: Reuse the existing `Query Model Multiple Times` or `Verify Model Inference` keyword instead of creating a new one here hardcoded to OVMS

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
```diff
@@ -0,0 +1,109 @@
+*** Settings ***
+Documentation     Collection of CLI tests to validate the model serving stack for different Large Language Models (LLM).
+...               These tests leverage on TGIS Standalone Serving Runtime
+Resource          ../../Resources/OCP.resource
+Resource          ../../Resources/CLI/ModelServing/llm.resource
+Library            OpenShiftLibrary
+Suite Setup       Suite Setup
+Suite Teardown    Suite Teardown
+Test Tags         KServe-OVNS   Sanity
```
lugi0, line 9: The `KServe-OVNS` tag does not exist

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
```diff
@@ -0,0 +1,109 @@
+*** Settings ***
```
lugi0, line 1: There is already a test suite for OVMS_on_kserve, why not extend that one?

lugi0 (changes requested):
The abbreviation everywhere should be changed from `ovns` to `OVMS`

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
```diff
@@ -744,7 +745,8 @@ Create PVC And Download Model From S3
     Create File From Template    ${DOWNLOAD_PVC_FILEPATH}    ${DOWNLOAD_PVC_FILLED_FILEPATH}
     ${rc}    ${out}=    Run And Return Rc And Output    oc -n ${namespace} apply -f ${DOWNLOAD_PVC_FILLED_FILEPATH}
     Should Be Equal As Integers    ${rc}    ${0}
-    Wait For Pods To Be Ready    label_selector=name=download-${model_name}    namespace=${namespace}
+    # commented for now as for ovns the model id downloaded very fast
+    # Wait For Pods To Be Ready    label_selector=name=download-${model_name}    namespace=${namespace}
```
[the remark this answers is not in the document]
mwaykole: it should not actually but ill verify it 
as the next line is waiting for the pod to Succeed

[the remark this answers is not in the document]
lugi0: They have different timeouts for different steps (getting the pod up and running vs. downloading the model in the PVC), so both are needed

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -581,3 +581,52 @@ Set Up Project
     ELSE
         Log    message=Skipping UserWorkloadMonitoring enablement.
     END
+
+Verify OVNS Model Inference Via CLI
+    [Documentation]    Verifies that the inference result of a model is equal to an expected output
```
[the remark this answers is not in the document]
lugi0: all the test cases in `model_serving_ovms_on_kserve` use `Verify Model Inference` / `Verify Model Inference With Retries` with OVMS on Kserve specifically.
Most if not all of the tests in the LLM subfolder use `Query Model Multiple Times`, and I think that keyword along with `runtime_query_formats.json` / `model_expected_responses.json` can be extended for OVMS

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -581,3 +581,52 @@ Set Up Project
...
+    [Arguments]    ${model_name}    ${inference_input}    ${expected_inference_output}    ${token_auth}=${FALSE}
+    ...    ${test_namespace}=${NONE}    ${kserve_mode}="Serverless"    ${retries}=${10}
+    ${url}=   Set Variable     ${NONE}
+    ${inference_output}=    Set Variable     ${NONE}
+    IF    '${kserve_mode}' == 'Serverless'
+        ${rc}    ${url}=    Run And Return Rc And Output
+        ...    oc get ksvc ${model_name}-predictor -n ${test_namespace} -o jsonpath='{.status.url}'
+         Should Be Equal As Integers    ${rc}    ${0}
+        ${curl_cmd}=   Set Variable    curl -sk ${url}/v2/models/${model_name}/infer -d ${inference_input}
```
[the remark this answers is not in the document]
lugi0: It will not, the host in the ISVC does not have the protocol in front, see the `Query Model Multiple Times` keyword https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tests/Resources/CLI/ModelServing/llm.resource#L376

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -581,3 +581,52 @@ Set Up Project
...
+    END
+
+    IF     '${kserve_mode}' == "RawDeployment"
+            ${url}=   Set Variable    http://localhost:8888/v2/models/${model_name}/infer
```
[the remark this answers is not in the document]
lugi0: You should not be using this keyword at all, and if you decide to create a new keyword make it dynamic based on its arguments rather than hardcoded please

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
lugi0: It does have UI tests in it right now since those are the only tests we have for OVMS on Kserve, I think it can be extended to have CLI tests as well

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
```diff
@@ -0,0 +1,109 @@
...
+${INFERENCE_INPUT}=    @tests/Resources/Files/modelmesh-mnist-input.json
+${INFERENCE_INPUT_OPENVINO}=    @tests/Resources/Files/openvino-example-input.json
+
+*** Test Cases ***
+Verify User Can Serve And Query A ovns Model
+    [Documentation]    Basic tests for preparing, deploying and querying a LLM model
+    ...                using Kserve and TGIS runtime
+    [Tags]    RHOAIENG-ovns
+    ${test_namespace}=  Set Variable     ovns-test-dir
+    ${model_name}=  Set Variable    test-dir
```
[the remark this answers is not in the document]
lugi0: You cannot remove it, you use it further down in the code. Modify the following line to use the variable rather than hardcode the model name.

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
lugi0: I gave you examples in https://github.com/red-hat-data-services/ods-ci/pull/1563/files#r1656901690

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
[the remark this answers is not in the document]
mwaykole: i tried almost my full day on Friday on it looks like both kw has it own limitation 
while  Query Model Multiple Times  when we add input /op in json its failing to load vaues i have tried multiple combination but still same value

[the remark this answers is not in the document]
mwaykole: another kw uses UI for multiple things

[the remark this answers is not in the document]
lugi0: Replied in Slack but pasting here as well for better tracking:
for 1) the fix is very simple, you just need to add the runtime ovns-runtime to runtime_query_formats.json (and again it should not be ovns but ovMs )  with its own query format.
For 2) it's even easier, you just need to update the Verify Model Inference keyword to use an additional argument that specifies whether you want to test via UI or via API - in the latter case you can skip any UI steps, e.g. the first two calls of the keyword and the call to get the model route via UI in the Get Model Inference keyword called within

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
```diff
@@ -0,0 +1,109 @@
...
+    Setup Test Variables    model_name=test-dir    use_pvc=${USE_PVC}    use_gpu=${USE_GPU}
+    ...    kserve_mode=${KSERVE_MODE}
+    ${EXPECTED_INFERENCE_SECURED_OUTPUT}  Set Variable    {"model_name":"${MODEL_NAME}__isvc-83d6fab7bd","model_version":"1","outputs":[{"name":"Plus214_Output_0","datatype":"FP32","shape":[1,10],"data":[-8.233053,-7.7497034,-3.4236815,12.3630295,-12.079103,17.266596,-10.570976,0.7130762,3.321715,1.3621228]}]}  #robocop: disable
+
+    Set Project And Runtime    runtime=${RUNTIME_NAME}     protocol=${PROTOCOL}     namespace=${test_namespace}
+    ...    download_in_pvc=${DOWNLOAD_IN_PVC}    model_name=${model_name}
+    ...    storage_size=10Gi
```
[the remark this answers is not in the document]
mwaykole: changed to 5GB

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
```diff
@@ -745,21 +746,25 @@ Create PVC And Download Model From S3
...
+    Run Keyword And Ignore Error   Wait For Pods To Be Ready    label_selector=name=download-${model_name}    namespace=${namespace}
     Wait For Pods To Succeed    label_selector=name=download-${model_name}    namespace=${namespace}
     ...    timeout=${download_timeout}
     ${rc}    ${out}=    Run And Return Rc And Output    oc delete pods -l name=download-${model_name} -n ${namespace}
     Should Be Equal As Integers    ${rc}    ${0}
 
 Extract Service Port
     [Arguments]    ${service_name}    ${namespace}    ${protocol}
-    IF    "${protocol}" == "grpc"
+    ${name}     Set Variable    name
```
lugi0, line 757: Can you use a different name for this variable? it's confusing

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
```diff
@@ -745,21 +746,25 @@ Create PVC And Download Model From S3
...
+    IF    '${protocol}' == 'grpc'
         ${port_name}=    Set Variable    h2c
+    ELSE IF    '${protocol}' == 'TCP'
+           ${name}=    Set Variable    protocol
+            ${port_name}=    Set Variable    TCP
```
lugi0, line 762: fix indentation

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
```diff
@@ -745,21 +746,25 @@ Create PVC And Download Model From S3
     Create File From Template    ${DOWNLOAD_PVC_FILEPATH}    ${DOWNLOAD_PVC_FILLED_FILEPATH}
     ${rc}    ${out}=    Run And Return Rc And Output    oc -n ${namespace} apply -f ${DOWNLOAD_PVC_FILLED_FILEPATH}
     Should Be Equal As Integers    ${rc}    ${0}
-    Wait For Pods To Be Ready    label_selector=name=download-${model_name}    namespace=${namespace}
+    Run Keyword And Ignore Error   Wait For Pods To Be Ready    label_selector=name=download-${model_name}    namespace=${namespace}
```
lugi0, line 749: I suggest using `Run Keyword And Continue On Failure` instead

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,37 +261,63 @@ Get Model Inference
...
+
+            IF     '${kserve_mode}' == "RawDeployment"
+                    ${url}=   Set Variable    http://localhost:${service_port}${end_point}
+                    ${curl_cmd}=     Set Variable    curl -sk ${url} -d ${inference_input}
+            END
+
+            IF      ${token_auth}
+                    ${token}     Create Inference Access Token   ${project_title}    ${DEFAULT_BUCKET_SA_NAME}
+                    ${curl_cmd}=     Catenate    ${curl_cmd}    -H "Authorization: Bearer ${token}"
+            END
```
lugi0, line 304: fix indentation here

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,37 +261,63 @@ Get Model Inference
...
+                ${curl_cmd}=   Set Variable    curl -sk ${url}${end_point} -d ${inference_input}
+            END
+
+            IF     '${kserve_mode}' == "RawDeployment"
+                    ${url}=   Set Variable    http://localhost:${service_port}${end_point}
+                    ${curl_cmd}=     Set Variable    curl -sk ${url} -d ${inference_input}
+            END
+
+            IF      ${token_auth}
+                    ${token}     Create Inference Access Token   ${project_title}    ${DEFAULT_BUCKET_SA_NAME}
```
lugi0, line 302: it's missing a `=` after the variable name, are you sure this works?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,37 +261,63 @@ Get Model Inference
...
-    ELSE IF  ${self_managed}
-        Fetch Openshift CA Bundle
-        ${curl_cmd}=     Catenate    ${curl_cmd}    --cacert openshift_ca.crt
     END
+    IF  ${deployment_mode} == 'Cli'
+            IF    '${kserve_mode}' == 'Serverless'
+                ${rc}    ${url}=    Run And Return Rc And Output
+                ...    oc get ksvc ${model_name}-predictor -n ${project_title} -o jsonpath='{.status.url}'
+                 Should Be Equal As Integers    ${rc}    ${0}
+                ${curl_cmd}=   Set Variable    curl -sk ${url}${end_point} -d ${inference_input}
```
lugi0, line 293: instead of using `-k` you should also implement the certificates (L280-285)

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,37 +261,63 @@ Get Model Inference
...
+            END
+
+            IF     '${kserve_mode}' == "RawDeployment"
+                    ${url}=   Set Variable    http://localhost:${service_port}${end_point}
+                    ${curl_cmd}=     Set Variable    curl -sk ${url} -d ${inference_input}
```
lugi0, line 298: instead of using -k you should also implement the certificates (L280-285)

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,37 +261,63 @@ Get Model Inference
...
-    ...    ${project_title}=${NONE}
-    ${self_managed}=    Is RHODS Self-Managed
-    ${url}=    Get Model Route Via UI    ${model_name}
-    ${kserve}=    Run Keyword And Return Status    SeleniumLibrary.Page Should Contain    Single-model serving enabled
-    ${curl_cmd}=     Set Variable    curl -s ${url} -d ${inference_input}
-    IF    ${token_auth}
-        IF    "${project_title}" == "${NONE}"
-            ${project_title}=    Get Model Project    ${model_name}
+    [Arguments]    ${model_name}    ${inference_input}    ${token_auth}=${FALSE}    ${kserve_mode}=Serverless
+    ...    ${project_title}=${NONE}   ${deployment_mode}='UI'   ${service_port}=8888   ${end_point}=${NONE}
```
lugi0, line 265: You are changing the order of the arguments, this will break existing tests calling this keyword expecting the 4th argument to be `project_title`. Preserve the original order by adding your new arguments after the already existing ones

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -311,7 +337,8 @@ Verify Model Inference With Retries
     ...                endpoint exposed.
     ...                This is a temporary mitigation meanwhile we find a better way to check the model
     [Arguments]    ${model_name}    ${inference_input}    ${expected_inference_output}
-    ...            ${token_auth}=${FALSE}    ${project_title}=${NONE}    ${retries}=${5}
+    ...            ${token_auth}=${FALSE}    ${project_title}=${NONE}     ${deployment_mode}=UI  ${kserve_mode}=Serverless
+    ...            ${service_port}=${NONE}   ${end_point}=${NONE}    ${retries}=${5}
```
lugi0, line 341: You are changing the order of the arguments, this will break existing tests calling this keyword expecting the 6th argument to be `retries`. Preserve the original order by adding your new arguments after the already existing ones

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
```diff
@@ -0,0 +1,111 @@
...
+${USE_GPU}=    ${FALSE}
+${KSERVE_MODE}=    Serverless    #RawDeployment   #Serverless
+${MODEL_FORMAT}=   onnx
+${PROTOCOL}=     http
+${OVERLAY}=      ${EMPTY}
+${MODELS_BUCKET}=    ${S3.BUCKET_1}
+${INFERENCE_INPUT}=    @tests/Resources/Files/modelmesh-mnist-input.json
+${INFERENCE_INPUT_OPENVINO}=    @tests/Resources/Files/openvino-example-input.json
+#${EXPECTED_INFERENCE_SECURED_OUTPUT}=    {"model_name":"${MODEL_NAME}__isvc-83d6fab7bd","model_version":"1","outputs":[{"name":"Plus214_Output_0","datatype":"FP32","shape":[1,10],"data":[-8.233053,-7.7497034,-3.4236815,12.3630295,-12.079103,17.266596,-10.570976,0.7130762,3.321715,1.3621228]}]}  #robocop: disable
+#${EXPECTED_INFERENCE_OUTPUT_OPENVINO}=    {"model_name":"${MODEL_NAME}__isvc-8655dc7979","model_version":"1","outputs":[{"name":"Func/StatefulPartitionedCall/output/_13:0","datatype":"FP32","shape":[1,1],"data":[0.99999994]}]}  #robocop: disable
```
lugi0, line 26: Remove if not needed

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
```diff
@@ -0,0 +1,111 @@
...
+    [Documentation]    Basic tests for preparing, deploying and querying model
+    ...                using Kserve and ovms runtime
+    [Tags]    OVMS
+    ...       Tier1
+    ...       Smoke
+    ...       OpenDataHub
+    ...       RHOAIENG-9045
+    Setup Test Variables    model_name=test-dir    use_pvc=${USE_PVC}    use_gpu=${USE_GPU}
+    ...    kserve_mode=${KSERVE_MODE}
+    ${EXPECTED_INFERENCE_SECURED_OUTPUT}  Set Variable    {"model_name":"${model_name}__isvc-83d6fab7bd","model_version":"1","outputs":[{"name":"Plus214_Output_0","datatype":"FP32","shape":[1,10],"data":[-8.233053,-7.7497034,-3.4236815,12.3630295,-12.079103,17.266596,-10.570976,0.7130762,3.321715,1.3621228]}]}  #robocop: disable
```
lugi0, line 39: If you need this set it in the Variables section instead of the test

lugi0 (changes requested):
Still some changes needed IMHO. Also test this PR against modelmesh to ensure it's not breaking its keywords.

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
mwaykole: if i set in in the variable it will fail as model_name is defined in the later phase of the test

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
mwaykole: what do u suggest ?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
mwaykole: ack moved it

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
mwaykole: ack moned it

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
mwaykole: done

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
mwaykole: done added

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
lugi0: You changed this in line 124/125 instead of 748/749 :)

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
lugi0: perhaps something like `protocol_name`, `jsonpath_name`, `jsonpath_parent`. As long as the variable isn't called `name` and its value is `name` it would be better.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,37 +261,62 @@ Get Model Inference
...
+            ${curl_cmd}=     Catenate    ${curl_cmd}    --cacert openshift_ca_istio_knative.crt
+        ELSE IF  ${self_managed}
+            Fetch Openshift CA Bundle
+            ${curl_cmd}=     Catenate    ${curl_cmd}    --cacert openshift_ca.crt
         END
-        ${token}=    Get Access Token Via UI    ${project_title}
-        ${curl_cmd}=     Catenate    ${curl_cmd}    -H "Authorization: Bearer ${token}"
     END
-    IF  ${kserve}
+    IF  ${deployment_mode} == 'Cli'
```
lugi0, line 288: For the CLI case you're forcing the Knative CA, but what if we want to have a CLI test case for modelmesh in a SM environment? You should add the openshift CA as well based on the conditions in 280/283

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,37 +261,62 @@ Get Model Inference
...
-    ...    ${project_title}=${NONE}
-    Open Model Serving Home Page
-    Switch Model Serving Project    ${project_title}
-    ${inference_output}=    Get Model Inference    ${model_name}    ${inference_input}    ${token_auth}
-    ...    project_title=${project_title}
+    ...    ${project_title}=${NONE}    ${deployment_mode}='UI'   ${kserve_mode}=Serverless
+    ...    ${service_port}=${NONE}   ${end_point}=${NONE} 
+    IF   ${deployment_mode} == 'UI'
+            Open Model Serving Home Page
+            Switch Model Serving Project    ${project_title}
```
lugi0, line 315: fix the indentation

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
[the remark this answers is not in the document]
lugi0: It is still 10Gi

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
lugi0: It is still 10Gi

lugi0 (changes requested):

ods_ci/tests/Tests/1000__model_serving/1006__model_serving_ovms_on_kserve_cli.robot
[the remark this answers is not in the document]
mwaykole: done reduced to 5

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
mwaykole: changed to protocol_name

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,44 +261,74 @@ Get Model Inference
...
-        Fetch Openshift CA Bundle
-        ${curl_cmd}=     Catenate    ${curl_cmd}    --cacert openshift_ca.crt
+    IF  ${deployment_mode} == 'Cli'
+        IF  ${self_managed}
+            Fetch Openshift CA Bundle
+            ${cert}=   Set Variable   --cacert openshift_ca.crt
+        ELSE
+            Fetch Knative CA Certificate    filename=openshift_ca_istio_knative.crt
+            ${cert}=    Set Variable    --cacert openshift_ca_istio_knative.crt
+        END
```
lugi0, line 294: This is not equivalent to 279-284, you're forcing a CA to be always added to the query. 
```
IF kserve
  ...
ELSE IF self-managed
  ...
END
```

Means that the kserve CA takes priority over the self-managed one, and if we're not using kserve and we're not in a SM cluster no CA should be added to the query
What you did:
```
IF self-managed
  ...
ELSE
  ...
END
```
Means that the self-managed CA takes priority over everything, if and if it's not a self-managed cluster you always default to the knative CA even if kserve is not being used, it's completely different.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,44 +261,74 @@ Get Model Inference
...
+        ELSE
+            Fetch Knative CA Certificate    filename=openshift_ca_istio_knative.crt
+            ${cert}=    Set Variable    --cacert openshift_ca_istio_knative.crt
+        END       
+        IF    '${kserve_mode}' == 'Serverless'
+               ${rc}    ${url}=    Run And Return Rc And Output    oc get ksvc ${model_name}-predictor -n ${project_title} -o jsonpath='{.status.url}'
+               Should Be Equal As Integers    ${rc}    0
+               ${curl_cmd}=    Set Variable    curl -s ${url}${end_point} -d ${inference_input} 
+        END
+        IF    '${kserve_mode}' == 'RawDeployment'
```
lugi0, line 300: this should be an ELSE IF

lugi0 (changes requested):
the CA portion is still wrong imho - please test this against the full regression suite of both kserve and modelmesh.

mwaykole:
smoke and normal test tested works fine nothing else is breaking

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,44 +261,74 @@ Get Model Inference
...
-    ELSE IF  ${self_managed}
-        Fetch Openshift CA Bundle
-        ${curl_cmd}=     Catenate    ${curl_cmd}    --cacert openshift_ca.crt
+    IF  ${deployment_mode} == 'Cli'
+        IF  ${self_managed}
+            Fetch Openshift CA Bundle
+            ${cert}=   Set Variable   --cacert openshift_ca.crt
+        ELSE IF  ${self_managed}
+            Fetch Knative CA Certificate    filename=openshift_ca_istio_knative.crt
+            ${cert}=    Set Variable    --cacert openshift_ca_istio_knative.crt
```
lugi0, line 293: you used the same variable for both conditions

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,44 +261,74 @@ Get Model Inference
...
+        END       
+        IF    '${kserve_mode}' == 'Serverless'
+               ${rc}    ${url}=    Run And Return Rc And Output    oc get ksvc ${model_name}-predictor -n ${project_title} -o jsonpath='{.status.url}'
+               Should Be Equal As Integers    ${rc}    0
+               ${curl_cmd}=    Set Variable    curl -s ${url}${end_point} -d ${inference_input} 
+        
+        ELSE IF    '${kserve_mode}' == 'RawDeployment'
+                    ${url}=   Set Variable    http://localhost:${service_port}${end_point}
+                    ${curl_cmd}=    Set Variable    curl -s ${url} -d ${inference_input} --cacert openshift_ca_istio_knative.crt
+        END
```
lugi0, line 303: indentation is wrong

lugi0 (changes requested):
If you have launched the tests please share the job id so I can look at the logs, thanks

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
mwaykole: okay i am kserve via 
oc get isvc test-dir -o jsonpath='{.metadata.name}' IMO its best way to validate it  the code also works fine validated in Jenkins @lugi0 plz review once u are back

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -261,44 +261,76 @@ Get Model Inference
...
-    IF  ${kserve}
-        Fetch Knative CA Certificate    filename=openshift_ca_istio_knative.crt
-        ${curl_cmd}=     Catenate    ${curl_cmd}    --cacert openshift_ca_istio_knative.crt
-    ELSE IF  ${self_managed}
-        Fetch Openshift CA Bundle
-        ${curl_cmd}=     Catenate    ${curl_cmd}    --cacert openshift_ca.crt
+    IF  ${deployment_mode} == 'Cli'
+        ${rc}   ${cmd_op}   Run And Return Rc And Output    oc get isvc ${model_name} -o jsonpath='{.metadata.name}'
+        Should Be Equal As Integers    ${rc}    0
+         IF  "${cmd_op}" == "${model_name}"
```
lugi0, line 290: This doesn't make sense, you're returning the model name from the ISVC and then checking that it's equal to the model name, it will always be true.
This condition will be true also when deploying models with modelmesh. 

You can achieve what you're looking for here with something like 
```
oc get isvc {model_name} -o jsonpath='{.metadata.annotations.serving\.kserve\.io/deploymentMode}' -n {my_ns}
```
It will return `ModelMesh` if it's using modelmesh or an empty string otherwise.

lugi0 (changes requested):
Please share the Jenkins job you've used to validate the changes, thanks

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
lugi0: I've reviewed it, this solution will not work

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
mwaykole: thanks i did not think of modelmesh thing ill apply it

lugi0 (approved):

nehachopra27 (approved):

ods_ci/tests/Resources/CLI/ModelServing/llm.resource
```diff
@@ -745,21 +746,25 @@ Create PVC And Download Model From S3
...
     Should Be Equal As Integers    ${rc}    ${0}
 
 Extract Service Port
     [Arguments]    ${service_name}    ${namespace}    ${protocol}
-    IF    "${protocol}" == "grpc"
+    ${protocol_name}     Set Variable    name
+    IF    '${protocol}' == 'grpc'
         ${port_name}=    Set Variable    h2c
+    ELSE IF    '${protocol}' == 'TCP'
+            ${protocol_name}=    Set Variable    protocol
```
bdattoma, line 761: this variable doesn't container the protocol name, though. Maybe you should give it a different name

PR #1567, opened 2024-06-27
Tag both Operator and ModelRegistry on 112_model_registry

asanzgom:
The Test Case inside this file is shared responsibility between Platform and Model Registry Scrum Teams, so tagging for better identification and organization on Test Reports.


bdattoma (approved):

bdattoma:
then probably you should do the same with other suites

asanzgom:
> then probably you should do the same with other suites

yes, will review and keep updating

jgarciao (approved):

PR #1577, opened 2024-06-28
All pipelines connections are secured by default from the client perspective

diegolovison:
All pipelines connections are secured by default from the client perspective

Self-signed certificate: rhods/job/rhods-smoke/5967/console
Trusted certificate: I didn't find a way to run it with Jenkins. Not going to share the cluster here. Tested locally and passed


jiridanek (approved):

kobihk (approved):

diegolovison:
> It could be useful for some test scenarios, no ?

I don't think a customer will send unencrypted data. The client is a computer that is **not** running in OpenShift.

PR #1586, opened 2024-07-02
Add new vllm runtime to fix RHOAIENG-9343

tarukumar:
https://issues.redhat.com/browse/RHOAIENG-9343


bdattoma (approved):

jstourac (approved):
lgtm

I suppose that this should be backported to `releases/2.11.0` then. Thanks.

PR #1593, opened 2024-07-03
DB Test Secrets Update

tonyxrmdavidson:
This commit will stop the test credentials for the Model-Registry DB being identified as a data leak from prodsec scanning.


nehachopra27 (approved):

FedeAlonso (approved):

grdryn:
@tonyxrmdavidson FYI my understanding is that you don't need to add the `# notsecret` comment when you have an appropriate `.gitleaks.toml` file that covers it. It shouldn't cause any issue to have both, so just FYI :)

tonyxrmdavidson:
> @tonyxrmdavidson FYI my understanding is that you don't need to add the `# notsecret` comment when you have an appropriate `.gitleaks.toml` file that covers it. It shouldn't cause any issue to have both, so just FYI :)

Hi Ger, I think you are correct. The tag #notsecret works for new commits, but will not work for older commits, hence the need for the toml file. If you have the toml file from the start and keep it updated you shouldn't need the tag.

PR #1599, opened 2024-07-04
Update must-gather image for RHOAI 2.11

jstourac:
See:
* https://github.com/red-hat-data-services/rhoai-disconnected-install-helper/blob/main/rhoai-2.11.md?plain=1#L53
* https://quay.io/repository/modh/must-gather/manifest/sha256:4a8d9398fdbc073e32611364d715a7210ce695d8b6bbbe957e3f9de6e4374e45

### DIFF
--- a/ods_ci/tests/Tests/505__must_gather/get-must-gather-logs.sh
+++ b/ods_ci/tests/Tests/505__must_gather/get-must-gather-logs.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 # Redirecting stdout/stderr of must-gather to a file, as it fills up the
 # process buffer and prevents the script from running further.
-oc adm must-gather --image=quay.io/modh/must-gather@sha256:1bd8735d715b624c1eaf484454b0d6d400a334d8cbba47f99883626f36e96657 &> must-gather-results.txt
+oc adm must-gather --image=quay.io/modh/must-gather@sha256:4a8d9398fdbc073e32611364d715a7210ce695d8b6bbbe957e3f9de6e4374e45 &> must-gather-results.txt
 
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/505__must_gather/get-must-gather-logs.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 # Redirecting stdout/stderr of must-gather to a file, as it fills up the
 # process buffer and prevents the script from running further.
-oc adm must-gather --image=quay.io/modh/must-gather@sha256:1bd8735d715b624c1eaf484454b0d6d400a334d8cbba47f99883626f36e96657 &> must-gather-results.txt
+oc adm must-gather --image=quay.io/modh/must-gather@sha256:4a8d9398fdbc073e32611364d715a7210ce695d8b6bbbe957e3f9de6e4374e45 &> must-gather-results.txt
```
CFSNM, line 4: why dont we use the rhoai-2.11 tag?

oc adm must-gather --image=quay.io/modh/must-gather:RHOAI-2.11

jstourac: the tag won't work in the disconnected env for us

also - what I did is just a fast dumb update - we need to find out a different way how to do this since we need to update image sha with each release now; but this should be in scope of the platform team, so I don't want to interfere (and also don't have time to do it myself :) ) - I also talked about it with @apodhrad couple of weeks ago, so he is also aware.

CFSNM (approved):

manosnoam (approved):

PR #1614, opened 2024-07-08
RHOAIENG-2510: Add test to validate the recreation of the KnativeServing resource

CFSNM:
Related ticket: https://issues.redhat.com/browse/RHOAIENG-2510

### DIFF
--- a/ods_ci/tests/Tests/100__deploy/130__operators/140__serverless_operator/140__serverless_operator.robot
+++ b/ods_ci/tests/Tests/100__deploy/130__operators/140__serverless_operator/140__serverless_operator.robot
@@ -36,6 +36,22 @@ Validate DSC creates all Serverless CRs
     #Verify Pod Logs Do Not Contain    ${podname}    ${OPERATOR_NAMESPACE}    ${regex_pattern}    rhods-operator
     Wait For DSC Conditions Reconciled    ${KNATIVESERVING_NS}    default-dsc
 
+Validate DSC Recreates The KnativeServing Resource When Deleted If Kserve is Managed
+    [Documentation]  The purpose of this test case is to validate the recreation of
+    ...    the Serverless KnativeServing resource if is previously deleted if kserve cmp is managed
+    [Tags]   Operator     Tier2     RHOAIENG-2510      ODS-2594
+    ${return_code}    ${output}    Run And Return Rc And Output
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/100__deploy/130__operators/140__serverless_operator/140__serverless_operator.robot
```diff
@@ -36,6 +36,22 @@ Validate DSC creates all Serverless CRs
...
+    ${return_code}    ${output}    Run And Return Rc And Output
+    ...    oc delete KnativeServing knative-serving -n ${KNATIVESERVING_NS} --ignore-not-found
+    Should Be Equal As Integers  ${return_code}   0   msg=Error deleting knative-serving, ${output}
+    Wait Until Keyword Succeeds    2 min    0 sec
+    ...        Check Number Of Resource Instances Equals To      KnativeServing     knative-serving      0
+    Wait Until Keyword Succeeds    6 min    0 sec
+    ...        Resource Should Exist     KnativeServing    knative-serving     ${KNATIVESERVING_NS}
+    Wait Until Keyword Succeeds    6 min    0 sec
+    ...        Resource Status Should Be     oc get KnativeServing knative-serving -n ${KNATIVESERVING_NS} -o json | jq '.status.conditions[] | select(.type=="Ready") | .status'     KnativeServing    "True"    # robocop: disable
+    Wait For Pods Numbers  2    namespace=${KNATIVESERVING_NS}
```
bdattoma, line 52: there are usually more than 2 pods in that namespace, you sure this check won't be flaky?

CFSNM: there are only 2 pods with the label "app.kubernetes.io/component=controller". I can add more conditions for the rest of pods with different labels, but this condition should be accurate

bdattoma: ops, I missed the label selector. 
hmm maybe you could check the total pod number of that NS as well? maybe there is a common label selector

CFSNM: sure, can try that

CFSNM:
This test case is not a valid scenario

PR #1639, opened 2024-07-12
Allow to install gpu-operator-certified in disconnected env

apodhrad:
Field 'startingCSV' is not needed and it causes problems in a disconnected env where we do not have all versions in the latest channel, e.g. in channel 'v24.3' we have only version '24.3.0'.

### DIFF
--- a/ods_ci/tasks/Resources/Provisioning/GPU/NVIDIA/gpu_install.yaml
+++ b/ods_ci/tasks/Resources/Provisioning/GPU/NVIDIA/gpu_install.yaml
@@ -27,7 +27,6 @@ spec:
   name: gpu-operator-certified
   source: certified-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: "gpu-operator-certified.v1.11.1"
 
 ---
 apiVersion: operators.coreos.com/v1alpha1
--- a/ods_ci/tasks/Resources/Provisioning/GPU/nfd_deploy.yaml
+++ b/ods_ci/tasks/Resources/Provisioning/GPU/nfd_deploy.yaml
@@ -7,7 +7,8 @@ spec:
   instance: "" # instance is empty by default
   topologyupdater: false # False by default
   operand:
-    image: registry.redhat.io/openshift4/ose-node-feature-discovery:v4.11
+    # Image digest for registry.redhat.io/openshift4/ose-node-feature-discovery:v4.11
+    image: registry.redhat.io/openshift4/ose-node-feature-discovery@sha256:d6242132d2ddec00c46d22b63015a33af821eace0150ba47d185cd992fee317d
     imagePullPolicy: Always
   workerConfig:
     configData: |
--- a/ods_ci/tasks/Resources/Provisioning/GPU/NVIDIA/gpu_deploy.sh
+++ b/ods_ci/tasks/Resources/Provisioning/GPU/NVIDIA/gpu_deploy.sh
@@ -9,7 +9,7 @@ CHANNEL="$(oc get packagemanifest gpu-operator-certified -n openshift-marketplac
 
 CSVNAME="$(oc get packagemanifests/gpu-operator-certified -n openshift-marketplace -o json | jq -r '.status.channels[] | select(.name == "'$CHANNEL'") | .currentCSV')"
 
-sed -i'' -e "0,/v1.11/s//$CHANNEL/g" -e "s/gpu-operator-certified.v1.11.0/$CSVNAME/g"  "$GPU_INSTALL_DIR/gpu_install.yaml"
+sed -i'' -e "0,/v1.11/s//$CHANNEL/g" "$GPU_INSTALL_DIR/gpu_install.yaml"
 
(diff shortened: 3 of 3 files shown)


ods_ci/tasks/Resources/Provisioning/GPU/nfd_deploy.yaml
```diff
@@ -7,7 +7,8 @@ spec:
   instance: "" # instance is empty by default
   topologyupdater: false # False by default
   operand:
-    image: registry.redhat.io/openshift4/ose-node-feature-discovery:v4.11
+    # Image digest for registry.redhat.io/openshift4/ose-node-feature-discovery:v4.11
+    image: registry.redhat.io/openshift4/ose-node-feature-discovery@sha256:d6242132d2ddec00c46d22b63015a33af821eace0150ba47d185cd992fee317d
```
bdattoma, line 11: so, for GPU operator we are going to install latest version available given that we are removing the `startingCSV` field, shouldn't we adopt the same approach with NFD and use some `latest` (if exists) or some other label?

apodhrad: Hi @bdattoma , that's correct. I noticed that we update the channel but now I have noticed it is because of channel changes in `nfd` package among OCP versions :(

So, I will probably get back the `startingCSV` and update the disconnected automation.

bdattoma: > we always install the latest version from the stable channel


yes. In my original comment I was referring to the NFD CustomResource, which here is fixed on a specific SHA. So I was wondering if we should use some "stable" or "latest" for the CR as well like we do for the operators

ods_ci/tasks/Resources/Provisioning/GPU/NVIDIA/gpu_install.yaml
```diff
@@ -27,7 +27,6 @@ spec:
   name: gpu-operator-certified
   source: certified-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: "gpu-operator-certified.v1.11.1"
```
bdattoma, line 30: could you check if https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tasks/Resources/Provisioning/GPU/NVIDIA/gpu_deploy.sh#L12 would create problems? maybe we can just remove that line from the script now

I don't know why it was put there in the first place, maybe @tarukumar remembers?

ods_ci/tasks/Resources/Provisioning/GPU/nfd_deploy.yaml
bdattoma: >Currently, with this static approach we have to add the image egistry.redhat.io/openshift4/ose-node-feature-discovery@sha256:d6242132d2ddec00c46d22b63015a33af821eace0150ba47d185cd992fee317d explicitly into a disconnected env.


that raises the question: how much frequently would we update this image sha?

bdattoma (approved):

ods_ci/tasks/Resources/Provisioning/GPU/NVIDIA/gpu_install.yaml
apodhrad: I suggest to completely omit the version specification - in gpu_install.yaml as well as in gpu_deploy.sh

apodhrad: @bdattoma I have noticed that the `sed` is incorrect, notice the version diffs
```
sed -i'' -e "0,/v1.11/s//$CHANNEL/g" -e "s/gpu-operator-certified.v1.11.0/$CSVNAME/g"  "$GPU_INSTALL_DIR/gpu_install.yaml"
```

vs

```
startingCSV: "gpu-operator-certified.v1.11.1"
```

bdattoma: ok, are you going to remove the sed part in this PR?

tarukumar: The fixed version was implemented because the the operator was causing the problme was causing issues if the starting CSV version wasn't provided. If this problem no longer occurs, we can remove this .

tarukumar (approved):

ods_ci/tasks/Resources/Provisioning/GPU/NVIDIA/gpu_install.yaml
apodhrad: I have removed the `sed` part and now I'm going to re-execute the installation with these changes.

apodhrad: Hi @bdattoma @tarukumar I have re-executed the job together with a basic GPU test in disconnected as well as in online env and it works

bdattoma (approved):

abhijeet-dhumal (approved):

PR #1648, opened 2024-07-17
Add annotation  so that model can be downloaded without any issue

tarukumar:


mwaykole (approved):

ods_ci/tests/Resources/Files/llm/bucket_secret.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
     serving.kserve.io/s3-usehttps: {{USE_HTTPS}}
     serving.kserve.io/s3-region: {{REGION}}
     serving.kserve.io/s3-useanoncredential: "false"
+    serving.kserve.io/s3-verifyssl: '0'
```
bdattoma, line 9: question: would this prevent from testing with passing a real CA certificate?

tarukumar: Yes, this is intended to address situations where the customer doesn't have any other components enabled in the cluster. It is not related to dashboard test cases where bundle validation is required, so official work will remain unaffected. In this case, you are creating secrets that are not managed by DSCsi or the operator, and we are not validating the bundle.We are not trying to create a data connection here. This is a method available in KServe. If we want to add a bundle, we need to use the bundle created by DSCi and apply the `serving.kserve.io/s3-cabundle-configmap` annotation. I might add this after performing the necessary validation. whihc will be in coming pr for now i will keep this.

PR #1654, opened 2024-07-22
Model Registry Deployment and  Integration test

tonyxrmdavidson:
This test applies the SecuredDB version of model regisrty to an Istio/ODH/ModelMesh cluster.
To review and run it locally you will need the following:

- Pull and checkout this PR
- Have an existing openshift cluster with Red Hat Athorino, OpenDataHub and Red Hat Openshift Service Mesh operators installed.

- Be oc logged into to your cluster

- Set the following local files
      ods_ci/run_robot_test.sh  line 25 SUBFOLDER to be set to true.

  ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot.  In the test case [Tags] (line 45), leave 4 spaces after OpenDataHub and put **_<YOUR_NAME>_** as a local tag to be picked up by the run command.

- run the following command from directory ods-ci/ods_ci 
`sh run_robot_test.sh --test-variable PRODUCT:ODH --test-variable APPLICATIONS_NAMESPACE:opendatahub --test-variable MONITORING_NAMESPACE:opendatahub --test-variable OPERATOR_NAMESPACE:openshift-operators --test-variable NOTEBOOKS_NAMESPACE:opendatahub --extra-robot-args '--variablefile test-variables-odh-overwrite.yml' --include  **<YOUR_NAME>** --skip-oclogin`


tonyxrmdavidson:
Screenshot 2024-07-30 at 14 09 59

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    Create OpenShift Project    ${NAMESPACE_MODEL-REGISTRY}
+
+Run OC Commands And Capture Output
+    [Documentation]  Logs the Domain and Token capture.
+    ${domain}=    Get Domain
+    ${token}=    Get Token
+    Set Suite Variable    ${DOMAIN}    ${domain}
+    Set Suite Variable    ${TOKEN}    ${token}
+    Log    Domain: ${DOMAIN}
+    Log    Token: ${TOKEN}
```
bdattoma, line 108: why printing the token?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    ...    oc delete project ${PRJ_TITLE} --force --grace-period=0
+    Should Be Equal As Integers	  ${return_code}	 0
+    Log    ${output}
+    RHOSi Teardown
+
+Create Project
+    [Documentation]    Create a project using oc command.
+    Create OpenShift Project    ${NAMESPACE_MODEL-REGISTRY}
+
+Run OC Commands And Capture Output
```
bdattoma, line 101: maybe for a better readability of the code you could make the keyword title saying what it does like `Get Cluster Domain And Token`

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    Launch Dashboard    ${TEST_USER.USERNAME}    ${TEST_USER.PASSWORD}    ${TEST_USER.AUTH_TYPE}
+    ...    ${ODH_DASHBOARD_URL}    ${BROWSER.NAME}    ${BROWSER.OPTIONS}
+    Open Data Science Projects Home Page
+    Create Data Science Project    title=${PRJ_TITLE}    description=${PRJ_DESCRIPTION}
+    Create Project
+    Apply ServiceMeshMember Configuration
+    Run OC Commands And Capture Output
+    Run Update Notebook Script
+    Copy File    ${EXAMPLE_ISTIO_ENV}    ${ISTIO_ENV}
+    Append Key Value To Env File    ${ISTIO_ENV}    DOMAIN    ${DOMAIN}
```
bdattoma, line 80: You could use `Create File From Template` which automatically resolves the variables you put in the template file

bdattoma: you find different examples in ods-ci

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    [Arguments]    ${domain_cert}
+    Upload File In The Workbench     filepath=${domain_cert}    workbench_title=${WORKBENCH_TITLE}
+    ...    workbench_namespace=${PRJ_TITLE}
+
+Create Generic Secret
+    [Documentation]    Creates Secret for model registry in a given namespace
+    [Arguments]    ${namespace}    ${secret_name}    ${key_file}    ${crt_file}    ${ca_file}
+    Log    This is the secret name ${secret_name}
+    ${command}=    Set Variable    oc create secret -n ${namespace} generic ${secret_name} --from-file=tls.key=${key_file} --from-file=tls.crt=${crt_file} --from-file=ca.crt=${ca_file}
+    Log    ${command}
```
bdattoma, line 169: this is already logged by Run Process

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+
+Apply ServiceMeshMember Configuration
+    [Documentation]    Apply a ServiceMeshMember configuration using oc.
+    Apply OpenShift Configuration    ${SERVICE_MESH_MEMBER}
+
+Apply Rest API Configuration
+    [Documentation]    Apply a Rest API configuration using oc.
+    Apply OpenShift Configuration    ${ENABLE_REST_API}
+
+Apply OpenShift Configuration
```
bdattoma, line 250: maybe this keyword could be moved in a common resource file to be re-used by others, for example `ods_ci/tests/Resources/OCP.resource`

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    Apply ServiceMeshMember Configuration
+    Run OC Commands And Capture Output
+    Run Update Notebook Script
+    Copy File    ${EXAMPLE_ISTIO_ENV}    ${ISTIO_ENV}
+    Append Key Value To Env File    ${ISTIO_ENV}    DOMAIN    ${DOMAIN}
+    File Should Exist    ${ISTIO_ENV}
+    Log File Content    ${ISTIO_ENV}
+    Generate ModelRegistry Certificates
+    Apply Db Config Samples    namespace=${NAMESPACE_MODEL-REGISTRY}
+    Create Model Registry Secrets
```
bdattoma, line 85: are these sensitive data? if yes, pls add the keyword in RHOSi.resource among the protected keywords

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+${NAMESPACE_ISTIO}=                  istio-system
+${NAMESPACE_MODEL-REGISTRY}=         odh-model-registries
+${SECRET_PART_NAME_1}=               modelregistry-sample-rest
+${SECRET_PART_NAME_2}=               modelregistry-sample-grpc
+${SECRET_PART_NAME_3}=               model-registry-db
+
+
+*** Test Cases ***
+Verify Model Registry Integration With Secured-DB
+    [Documentation]    Verifies the Integartion of Model Registry operator with Jupyter Notebook
```
bdattoma, line 40: is it testing ingration with secure DB or Jupyter Notebook or both?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    Wait Until Workbench Is Started     workbench_title=registry-wb
+    Upload File In The Workbench     filepath=${SAMPLE_ONNX_MODEL}    workbench_title=${WORKBENCH_TITLE}
+    ...         workbench_namespace=${PRJ_TITLE}
+    Upload File In The Workbench     filepath=${JUPYTER_NOTEBOOK_FILEPATH}    workbench_title=${WORKBENCH_TITLE}
+    ...         workbench_namespace=${PRJ_TITLE}
+    Launch And Access Workbench    workbench_title=${WORKBENCH_TITLE}
+    ...    username=${TEST_USER.USERNAME}     password=${TEST_USER.PASSWORD}
+    ...    auth_type=${TEST_USER.AUTH_TYPE}
+    Upload Certificate To Jupyter Notebook    ${CERTS_DIRECTORY}/domain.crt
+    Jupyter Notebook Should Run Successfully     ${JUPYTER_NOTEBOOK}
```
bdattoma, line 63: what this jupyter notebook does? Maybe you could make it clear from the keyword name, for example 
`Jupyter Notebook Can Query Model Registry` or sth similar

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+${NAMESPACE_MODEL-REGISTRY}=         odh-model-registries
+${SECRET_PART_NAME_1}=               modelregistry-sample-rest
+${SECRET_PART_NAME_2}=               modelregistry-sample-grpc
+${SECRET_PART_NAME_3}=               model-registry-db
+
+
+*** Test Cases ***
+Verify Model Registry Integration With Secured-DB
+    [Documentation]    Verifies the Integartion of Model Registry operator with Jupyter Notebook
+    [Tags]    OpenDataHub
```
bdattoma, line 41: is this test for OpenDataHub only?

ods_ci/mysql_link
```diff
@@ -0,0 +1 @@
+tests/Resources/CLI/ModelRegistry/samples/mysql
```
lugi0, line 1: Is this file needed?

ods_ci/tests/Resources/CLI/ModelRegistry/MRMS_TEST_RUNNER.ipynb
```diff
@@ -0,0 +1,186 @@
+{
```
lugi0, line 1: Most (all?) of the jupyter notebooks we use are in this repo: https://github.com/redhat-rhods-qe/ods-ci-notebooks-main - perhaps it'd be better to put this there as well and clone it into your Notebook instead of uploading it from local?
cc @jgarciao @jstourac you might have an opinion on this as well?

ods_ci/tests/Resources/CLI/ModelRegistry/updateIPYNB.py
```diff
@@ -0,0 +1,44 @@
+import json
```
lugi0, line 1: I'm not 100% sure about what the goal of this script is, but it seems like you're fetching some values with `oc` and then substituting them in the notebook file. 
I don't think this is the best way to do this to be honest - if you need to fetch values at runtime from your notebook, wouldn't it be better to run the oc commands from the notebook itself?
If these values are not taken/created at runtime, wouldn't it be better to have the notebook with the values available already?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    Create Workbench    workbench_title=${WORKBENCH_TITLE}    workbench_description=Registry test
+    ...                 prj_title=${PRJ_TITLE}    image_name=Minimal Python  deployment_size=Small
+    ...                 storage=Persistent   pv_existent=${NONE}
+    ...                 pv_name=${NONE}  pv_description=${NONE}  pv_size=${NONE}
+    Workbench Should Be Listed      workbench_title=registry-wb
+    Open Data Science Project Details Page       project_title=${PRJ_TITLE}
```
lugi0, line 47: Aren't you already in this page?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    ${workbenches}=    Create List    ${WORKBENCH_TITLE}
+    Create S3 Data Connection    project_title=${PRJ_TITLE}    dc_name=${DC_S3_NAME}
+    ...            aws_access_key=${S3.AWS_ACCESS_KEY_ID}    aws_secret_access=${S3.AWS_SECRET_ACCESS_KEY}
+    ...            aws_bucket_name=${AWS_BUCKET}    connected_workbench=${workbenches}
+    Data Connection Should Be Listed    name=${DC_S3_NAME}    type=${DC_S3_TYPE}    connected_workbench=${workbenches}
+    Open Data Science Project Details Page       project_title=${prj_title}    tab_id=workbenches
+    Wait Until Workbench Is Started     workbench_title=registry-wb
+    Upload File In The Workbench     filepath=${SAMPLE_ONNX_MODEL}    workbench_title=${WORKBENCH_TITLE}
+    ...         workbench_namespace=${PRJ_TITLE}
+    Upload File In The Workbench     filepath=${JUPYTER_NOTEBOOK_FILEPATH}    workbench_title=${WORKBENCH_TITLE}
```
lugi0, line 57: see comment about jupyter notebook file

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    [Documentation]    Suite setup steps for testing Model Registry.
+    Set Library Search Order    SeleniumLibrary
+    RHOSi Setup
+    Launch Dashboard    ${TEST_USER.USERNAME}    ${TEST_USER.PASSWORD}    ${TEST_USER.AUTH_TYPE}
+    ...    ${ODH_DASHBOARD_URL}    ${BROWSER.NAME}    ${BROWSER.OPTIONS}
+    Open Data Science Projects Home Page
+    Create Data Science Project    title=${PRJ_TITLE}    description=${PRJ_DESCRIPTION}
+    Create Project
+    Apply ServiceMeshMember Configuration
+    Run OC Commands And Capture Output
```
lugi0, line 77: this kw is a bit weird, maybe update the name to specify what it's actually doing with oc?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+    [Documentation]  Logs the Domain and Token capture.
+    ${domain}=    Get Domain
+    ${token}=    Get Token
+    Set Suite Variable    ${DOMAIN}    ${domain}
+    Set Suite Variable    ${TOKEN}    ${token}
+    Log    Domain: ${DOMAIN}
+    Log    Token: ${TOKEN}
+
+Generate ModelRegistry Certificates
+    [Documentation]    Generates OpenSSL certificates for Model-Registry using the generate_certs.sh script.
```
lugi0, line 111: This is strange to me, why do you need to generate all these certificates?
I would assume that RHOAI/ODH takes care of creating what is needed for Model Registry? Or would a end user be expected to create all of this by themselves?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+Get Token
+    [Documentation]    Gets the Token and returns it to 'Run Oc Commands And Capture Output'.
+    ${token_result}=    Run Process    oc    whoami    -t    stdout=YES
+    ${rc}=    Set Variable    ${token_result.rc}
+    IF    ${rc} > 0    Fail    Command 'oc whoami -t' returned non-zero exit code: ${rc}
+    ${token}=    Set Variable    ${token_result.stdout}
+    RETURN    ${token}
+
+Create OpenShift Project
+    [Documentation]    Creates a test namespace and tracks it via ServiceMesh
```
lugi0, line 229: "tracks it via Service Mesh"
How does it do that? this seems to only be creating a new project if it doesn't exist already. Did you want to apply the SMM file as well?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -0,0 +1,277 @@
...
+Teardown Model Registry Test Setup
+    [Documentation]  Teardown Model Registry Suite
+    JupyterLibrary.Close All Browsers
+    ${return_code}    ${output}=    Run And Return Rc And Output
+    ...    oc delete project ${PRJ_TITLE} --force --grace-period=0
+    Should Be Equal As Integers	  ${return_code}	 0
+    Log    ${output}
+    RHOSi Teardown
+
+Create Project
```
lugi0, line 97: Why are you wrapping the call to `Create OpenShift Project` in another keyword?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
lugi0: This seems to only be calling `oc apply`, I'm not really sure we need a keyword for it?
And to be honest we probably already have some keyword(s) that just apply a file, so in that case it might make sense to centralize all of them under one keyword?

lugi0 (commented):
In general you're adding a ton of resource files, and I don't know enough about Model Registry to understand why you're doing it. 
In ODH/RHOAI, I would expect the operator / controller pods to set up everything that is needed in OpenShift, is that not the case?

Also, there are many new keywords defined in `1302_model_registry_model_serving.robot`, but it might be better to have them in a `Resources` file so that they can be reused by other tests for Model Registry?

I'm also not sure of how you're handling the jupyter notebook file needed for your tests

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: Good suggestion, I have done this.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: I have removed that Log.  I wanted to ask you about the linting for this long line. I've tried various different ways of splitting the line, none of which work. Is there a way I should be doing it or is it ok to have such a long line of code?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: I agree with both of you. If something like that exists then I should use it, if not it should go into a common file for others to use. I'll check it out.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: They are certificates that are generated solely for the test and as part of the test. Every time the test runs new certs are created.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: It applies an istio-tls model-registry with a secureDB It integrates with Jupyter Notebook by deploying the model-registry python client which then interacts with model-registry via a rest-api.

So really 3 important parts of the test. 

- Does the secured MR deploy correctly
- Does the MR python client deploy correctly
- Can you interact with model-registry using the python client via a jupyter notebook.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: At the moment as I don't know how to run it against RHOAI as model-registry is not available from RHOAH atm.

ods_ci/mysql_link
tonyxrmdavidson: All of the files for model-registry deployment are needed to deploy the service until it is fully available in ODH, by default model-registry is :removed.   
This will likely change once it gets into ODH and I can refactor the test at that point. For now though they are needed.

ods_ci/tests/Resources/CLI/ModelRegistry/updateIPYNB.py
tonyxrmdavidson: The python client requires the Domain of the hosting cluster, so it's not possible to have the notebook file with pre-done values. That's the main reason I'm keeping it locally, so I can work with it and then upload directly to Jupyter Notebook.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: I'm not sure what you mean?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: Already done that based on Berto's comment.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: That is one of the use cases. The end user might be creating their own certificates.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: Yes, the SMM file applied earlier in the set up

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: hmm, good question. 🤔

tonyxrmdavidson: Removed Create Project and called Create Openshift Project directly

ods_ci/tests/Resources/CLI/ModelRegistry/updateIPYNB.py
tonyxrmdavidson: I had tried to get the values directly in the notebook file at run time, but the permissions are not available to retrieve the Domain.

ods_ci/mysql_link
tonyxrmdavidson: There is also the consideration that it has to run on a disconnected cluster. As I understand, it is not possible to use an external reference from within a disconnected cluster to apply model-registry.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
lugi0: You call the keyword `Open Data Science Project Details Page       project_title=${PRJ_TITLE}`, but given the code executed between the Suite Setup and this Test Case, aren't you already in the Data Science Project details page at this point?

ods_ci/tests/Resources/CLI/ModelRegistry/updateIPYNB.py
lugi0: I'm not 100% sure what you mea by Domain, but would this work?
```
import os
test = os.getenv("NOTEBOOK_ARGS")
dashboard = test.split('"hub_host":')[1].split(",")[0].strip('"')
console = url.replace("rhods-dashboard-redhat-ods-applications","console-openshift-console")
print(f'Env variable: \n{test}')
print('--------------------------')
print('Dashboard url: '+url)
print('--------------------------')
print('Console url: '+console)
```

ods_ci/mysql_link
lugi0: I understand you need most files to set up Model Registry until it is fully integrated, but i was wondering if this `mysql_link` is actually needed, it seems to be a symlink to another directory?

lugi0: >As I understand, it is not possible to use an external reference from within a disconnected cluster to apply model-registry

In general yes, you're correct, in disconnected it's not possible to fetch resources remotely from the cluster itself.
But if you need the ods-ci tests to fetch resources and then apply them from local to the cluster, that is possible in disconnected.
However this seems to be a temporary problem that will go away once ODH/RHOAI can actually set everything up for Model Registry?

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
lugi0: Yeah, given the documentation my question was if you actually wanted to include the call to `Apply ServiceMeshMember Configuration` inside this keyword, rather than calling them separately from the test case.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
bdattoma: if going new line after Set Variable is not enough, maybe just disable the linter check on the line. - add the comment `# robocop: disable:line-too-long`

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: I decided not to make this change as the code will be removed within a week due to changes in model-registry-operator

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: I have put the keyword into OCP.resource and called it from there.  I searched for another function but couldn't see any.  I think it makes sense to have a keyword instead of directly calling oc apply -f for each time it is needed. It has fewer lines of code with the keyword.

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: I have updated based on this suggestion

ods_ci/mysql_link
tonyxrmdavidson: I think I will leave it the way it is for now as it is temporary. I will update appropriately once the changes are actioned in model-registry-operator.

ods_ci/tests/Resources/CLI/ModelRegistry/updateIPYNB.py
tonyxrmdavidson: As I have to get the domain for the setup to be used elsewhere I think it makes more sense to use during setup for both cert generation and the notebook file.  I think this way we will have less code.  For now I'm under a bit of time pressure to get this merged.  I'm really interested in this though, _test = os.getenv("NOTEBOOK_ARGS")_. I wasn't aware this was available.

tonyxrmdavidson: What I mean by Domain is the part of the route urls that is static and does not change. For example apps.tony-modelreg.osp.rh-ods.com

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
tonyxrmdavidson: Ahh yes, that makes sense

ods_ci/tests/Resources/CLI/ModelRegistry/updateIPYNB.py
lugi0: >For now I'm under a bit of time pressure to get this merged

You can definitely get the domain from that environment variable, please refactor this when you update the test

lugi0 (approved):
If this needs to get merged quickly I'm ok approving it, but IMHO there's some refactoring to be done ASAP

bdattoma:
could you also edit a bit the PR title? So from release notes it would be clearer what was added

lugi0 (dismissed):

ods_ci/tests/Tests/1300__model_registry/Model_Registry_Testing.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Create and prepare the test-variables.yml file by using the template file test-variables.yml.example file. The following will need to be set.
+
+  OCP_CONSOLE_URL: "https://console-openshift-console.<YOUR_DOMAIN>.com/"
+  
+  ODH_DASHBOARD_URL: "https://odh-dashboard-opendatahub.<YOUR_DOMAIN>.com/"
+
+- Set the following local files
+  
+  ods_ci/run_robot_test.sh  line 25 SUBFOLDER to be set to true.
```
bdattoma, line 21: what does it mean?

bdattoma: ah, but you should use `--no-output-subfolder true` script argument instead of changing the script code

tonyxrmdavidson: This is a README for the guys in model-registry to help them use the test locally.  So, if I'm getting this I could pass --no-output-subfolder true as an argument in the run command?

bdattoma (approved):

nehachopra27 (approved):

tonyxrmdavidson:
Jira to track the DSCI issue stopping this PR passing on the PR check pipeline.
https://issues.redhat.com/browse/RHOAIENG-11119

PR #1662, opened 2024-07-25
Fix: ODS 2206 wrong xpaths

FedeAlonso:
During the lasts nightlies 2.12 executions a test has been failing due a modification in some of the dashboard paths.

Test Case: Verify User Can Create, Run and Delete A DS Pipeline From DS Project Details Page Using Custom Pip Mirror
ODS-2206    ODS-2226    ODS-2633

Test results:
image

About the 2s wait: 
Sometimes it takes a little bit more to render the row (check the screenshot below). Andwe cannot wait the element, because the element can not be there.

Screenshot from 2024-07-25 00-46-06


jiridanek (approved):

manosnoam (approved):
Thank you for the fix Fede!

PR #1664, opened 2024-07-25
RHOAIENG-10404 - Add ModelRegistry Component Test Cases

mattmahoneyrh:
Test cases to validate DSC component ModelRegistry Controller Managed and Removed states. Tests validate that ModelRegistry resources are properly Managed and Removed.


mattmahoneyrh:
Note: Jenkins test passed against ODH. Will test against RHOAI, when component is supported in DSC.

PR #1666, opened 2024-07-26
Add data science project creation deletion keywords. Modify Elyra test suite to use them

jgarciao:
- Adds keywords  `Create Data Science Project From CLI` and `Delete Data Science Project From CLI By Name`
- Modifies Elyra test suite to use them

Tested successfully in rhods-ci-pr-test/3176

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
@@ -158,16 +158,16 @@ Create Data Science Project
 Create Data Science Project From CLI
     [Documentation]    Creates a Data Science Project, adding the needed labels to be recognized by Dashboard.
     ...    - If ${as_user} is provided, the project is created for that user, and group will be set as
-    ...      system:authenticated:oauth, as it is recommended in https://docs.openshift.com/container-platform/4.16/applications/projects/creating-project-other-user.html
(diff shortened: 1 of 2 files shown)


diegolovison (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
```diff
@@ -180,20 +219,25 @@ Delete Data Science Projects From CLI
...
-        END
+        Delete Data Science Project From CLI By Name    name=${project_k8s_name}
+    END
+
+Delete Data Science Project From CLI By Name
+    [Documentation]    Deletes a single DS project usig CLI command by project resource name
+    [Arguments]     ${name}
+    Oc Delete    kind=Project   name=${name}
+    WHILE    ${TRUE}
+        ${exists}=    Run And Return Rc    oc get project ${name}
```
jstourac, line 238: instead of the while loop, can't we use something like:
```
oc wait --for=delete --timeout=60s project ${name}
```

maybe we can wait already with the first command to delete the project using `--wait=true`

jiridanek (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
jgarciao: I also don't like that part of the code, but it was already like this in the existing keyword. I'll try to fix it in a follow-up PR

PR #1668, opened 2024-07-29
Workaround for Poetry urllib3 connection error

manosnoam:
According to https://stackoverflow.com/questions/78694413 
we can run: `poetry config installer.parallel false`  in order to resolve Poetry urllib3 `Connection broken: IncompleteRead`.

Poetry output shows no error:

image

### DIFF
--- a/ods_ci/run_robot_test.sh
+++ b/ods_ci/run_robot_test.sh
@@ -324,6 +324,8 @@ if [[ ${SKIP_INSTALL} -eq 0 ]]; then
   if [[ -d "${virtenv}" ]]; then
     echo "Using a pre-created virtual environment in '${virtenv}' for poetry to save time."
     poetry config --local virtualenvs.in-project true
+    # Workaround for Poetry urllib3 connection error
+    poetry config --local installer.parallel false
     ln --symbolic "${virtenv}" "${basepath}/../.venv"
   else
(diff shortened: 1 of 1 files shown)


ods_ci/run_robot_test.sh
```diff
@@ -324,6 +324,7 @@ if [[ ${SKIP_INSTALL} -eq 0 ]]; then
   if [[ -d "${virtenv}" ]]; then
     echo "Using a pre-created virtual environment in '${virtenv}' for poetry to save time."
     poetry config --local virtualenvs.in-project true
+    poetry config installer.parallel false
```
jgarciao, line 327: Not sure, but maybe you have to add `--local` too like in the line before

manosnoam: In the original post https://github.com/python-poetry/poetry/issues/3219 I did not see any recommendation to use `--local` too, but I can try for the consistency, thanks!

asanzgom (approved):

CFSNM (approved):

ods_ci/run_robot_test.sh
manosnoam: Works as well, updated.

jiridanek (approved):

PR #1671, opened 2024-07-30
Move enable and disable trainingoperator from parent suite setup

ChughShilpa:
Recently we observered all smoke tests failed because of the parent suite set up for training operator failed.
So, to avoid smoke failures in other components moving enable and disable of training operator from parent suite setup

### DIFF
--- a/ods_ci/tests/Tests/650__distributed_workloads/test-smoke.robot
+++ b/ods_ci/tests/Tests/650__distributed_workloads/test-smoke.robot
@@ -2,8 +2,8 @@
 Documentation     Smoke tests for DistributedWorkloads
 Library           Process
 Resource          ../../../tasks/Resources/RHODS_OLM/install/oc_install.robot
-Suite Setup       Prepare Codeflare E2E Test Suite
-Suite Teardown    Teardown Codeflare E2E Test Suite
+Suite Setup       RHOSi Setup
+Suite Teardown    RHOSi Teardown
 
 
 *** Test Cases ***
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/650__distributed_workloads/test-smoke.robot
```diff
@@ -72,6 +70,9 @@ Training operator smoke test
     [Documentation]    Check that Training operator deployment is up and running
     [Tags]    Smoke
     ...       DistributedWorkloads
+
```
jiripetrlik, line 73: Maybe it would be better to use Setup and Teardown keywords as are mentioned here. It will make sure that tear down is run even if test fails. https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#test-setup-and-teardown

ChughShilpa: Added the suggestion

jiripetrlik (approved):

abhijeet-dhumal (dismissed):

Srihari1192 (commented):
@ChughShilpa Just to confirm don't we need RHOSi Setup and RHOSi Teardown for other smoke tests. Earlier we are running this step for all smoke tests but now its applicable for only Training Operator Tests

ChughShilpa:
> @ChughShilpa Just to confirm don't we need RHOSi Setup and RHOSi Teardown for other smoke tests. Earlier we are running this step for all smoke tests but now its applicable for only Training Operator Tests

We need it for all smoke tests, thanks for this question :)

ChughShilpa:
> > @ChughShilpa Just to confirm don't we need RHOSi Setup and RHOSi Teardown for other smoke tests. Earlier we are running this step for all smoke tests but now its applicable for only Training Operator Tests
> 
> We need it for all smoke tests, thanks for this question :)

Updated the PR. Please review @Srihari1192

Srihari1192 (approved):

PR #1672, opened 2024-07-30
Fix Python selftests on macOS by using portable shell commands only

jiridanek:
* /bin/true and /bin/false don't exist as executable binaries on disk, need to use `true` and `false` shell builtin commands
* echo -e is a nonstandard extension and there are differing implementations, using printf is more reliable


kobihk (approved):

jstourac (approved):
I would deny to use MacOS and world would be much simpler :smile:

PR #1684, opened 2024-08-01
Enable public endpoint by default for kserve models deployed via UI

lugi0:


lugi0:
Verified in rhods-smoke/6444, some failures but they were caused by the test environment somehow breaking and logging our user out. The main issue this aims to fix is confirmed fixed.

FedeAlonso (dismissed):

mwaykole (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -434,6 +435,12 @@ Deploy Kserve Model Via UI  #robocop: disable
     Select Framework    ${model_framework}
     Set Replicas Number With Buttons    ${replicas}
     Set Server Size    ${size}
+    IF    ${public_endpoint}
+        Enable External Serving Route
+        # By default this also enables Token Authentication. Let's disable it and let it get re-enable by the next
+        # IF block if needed.
+        Disable Token Authentication
```
bdattoma, line 442: does this fail if the checkbox is already unselected? 
I was thinking that if the expected behavior is to have it Selected by default but we want to disable it for test purpose, maybe we could still check that it's selected by default before deselecting

lugi0: >does this fail if the checkbox is already unselected?

From the [documentation](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Unselect%20Checkbox): `Does nothing if the checkbox is not selected.`, so it should just continue execution

>I was thinking that if the expected behavior is to have it Selected by default but we want to disable it for test purpose, maybe we could still check that it's selected by default before deselecting

Good idea, will add this check

lugi0: Done

mwaykole (approved):

bdattoma (approved):

PR #1669, opened 2024-07-29
RHOAIENG-10534: look for branch-specific poetry .venv

jiridanek:
* https://gitlab.cee.redhat.com/ods/rhods-qe-tools/-/merge_requests/42
* https://issues.redhat.com/browse/RHOAIENG-10534

PR #1686, opened 2024-08-02
[releases/2.8.0] RHOAIENG-10534: look for branch-specific poetry .venv (#1669)   (cherry picked from commit d09c4d3)

jiridanek:
/job/rhods/job/rhods-ci-pr-test/3194/console

### DIFF
--- a/ods_ci/run_robot_test.sh
+++ b/ods_ci/run_robot_test.sh
@@ -320,7 +320,8 @@ fi
 
 if [[ ${SKIP_INSTALL} -eq 0 ]]; then
   # look for pre-created poetry .venv
-  virtenv="${HOME}/.local/ods-ci/.venv"
+  branch=$(git branch --show-current)
+  virtenv="${HOME}/.local/ods-ci/${branch%,*}/.venv"
   if [[ -d "${virtenv}" ]]; then
     echo "Using a pre-created virtual environment in '${virtenv}' for poetry to save time."
     poetry config --local virtualenvs.in-project true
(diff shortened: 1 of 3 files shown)


ods_ci/run_robot_test.sh
```diff
@@ -319,12 +319,29 @@ if command -v yq &> /dev/null
...
+      break
+    fi
+  done
+  if [[ -d "${basepath}/../.venv" ]]; then
+    echo "We ran already before and settled in venv we linked into '${basepath}/../.venv'."
+  elif [[ -d "${virtenv}" ]]; then
     echo "Using a pre-created virtual environment in '${virtenv}' for poetry to save time."
     poetry config --local virtualenvs.in-project true
+    # Workaround for Poetry urllib3 connection error
+    poetry config --local installer.parallel false
```
jstourac, line 340: I'm wondering whether we still want this workaround, but I suppose that it doesn't matter that much if the venvs will be in place in the most of the time this is executed.

jstourac (approved):

ods_ci/run_robot_test.sh
jiridanek: I think we do, the issue in poetry was not resolved and because the .venvs are in place, installing serially will not slow us much.

jstourac: Yes, I'm aware. My point for removing was because we have the .venvs ready for different branches so it's less probable to hit. But yeah, let's keep it there for the cache-miss situations.

kobihk (approved):

PR #1703, opened 2024-08-08
fix: Component Status kw tolerant to not present components

CFSNM:
Added this default value for jq for components that are not enabled in the DSC. By default, if a component is not specified in the DSC yml file, it is added with an empty map {}, so its the same as setting the component as Removed. 

When we try to gather the status of a component like this, we get a null error due to missing managementState key

Example:

```
spec:
  components:
    codeflare:
      managementState: Managed
    kserve:
      defaultDeploymentMode: Serverless
      managementState: Managed
      serving:
        ingressGateway:
          certificate:
            type: OpenshiftDefaultIngress
        managementState: Managed
        name: knative-serving
    modelregistry: {}
    trustyai:
      managementState: Managed
    ray:
      managementState: Managed
    kueue:
      managementState: Managed
    workbenches:
      managementState: Managed
    dashboard:
      managementState: Managed
    modelmeshserving:
      managementState: Managed
    datasciencepipelines:
      managementState: Managed
    trainingoperator:
      managementState: Removed
```

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -365,7 +365,7 @@ Component Should Not Be Enabled
 Is Component Enabled
     [Documentation]    Returns the enabled status of a single component (true/false)
     [Arguments]    ${component}    ${dsc_name}=${DSC_NAME}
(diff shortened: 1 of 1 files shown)


jstourac (approved):

tonyxrmdavidson (approved):
/lgtm

PR #1705, opened 2024-08-08
Remove default value of gpu taints

aloganat:


bdattoma (approved):

lugi0 (approved):
we could've kept the option to have taints if they are ever needed but this works too IMHO

mwaykole (approved):

PR #1706, opened 2024-08-08
Add script for checking if a build is RC or not

bdattoma:
Part of a larger effort to increase automation in downstream testing. 
Given the build version and date, the script checks the official code freeze date and if the build date is over the code freeze date then the build is considered as RC.

```
python3 utils/scripts/detect_rc.py --smartsheet-token <truncated> --smartsheet-id <truncated> --build-version 2.13.0-Z --build-date YYYY-MM-DDThh:mm:ss.sssssZ

Code freeze for 2.13.0-10 is on <truncated>
The buiild 2.13.0-10 is NOT an RC because created before code freeze date <truncated>
Days remaining from Code Freeze N
```


jstourac:
This is insane... I don't understand why we don't have a flag in UMB message for this kind of thing...

bdattoma:
Closing and moving in another place

PR #1710, opened 2024-08-12
fix: solve issue with flaky test about serverless resources

CFSNM:
The Serverless resources take some time to reconcile, so adding a "Wait Until Keyword Succeeds" to add some retries


jstourac (approved):
I would prefer to try to update this with `oc wait` somehow, but since this is just an addition of an existing check, let's keep it this way then...

manosnoam (approved):

PR #1714, opened 2024-08-13
Exclude drift tests from disconnected

adolfo-ab:
Drift tests are currently pulling images directly from quay, and those images are not present in disconnected clusters. We need to exclude drift tests from disconnected runs until we adapt them.


nehachopra27 (approved):

apodhrad:
Let's wait for a while. It seems that fixing this will be easy

PR #1716, opened 2024-08-13
General fixes for UI Keywords in RHOAI 2.12

manosnoam:
- Fix Keywords `Delete Data Science Project From CLI` and `Delete Data Science Project From CLI By Name` for Tests Setup/Teardown, instead of the UI Keywords.
- New KW `Generate Unique Namespace` in `Create Data Science Project`.
- `Login To RHODS Dashboard` waits for Login page to disappear after login.
- Call `Filter Projects By Name` before `Wait Until Project Is Listed`.
- Add timeout to `Workbench With Description Should Be Listed`.
- Fix `Create/Edit S3 Data Connection`.
- Remove the Obsolete Test `Verify User Can Access Permission Tab`.

image

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
@@ -76,7 +76,7 @@ Authorize rhods-dashboard service account
 Login To RHODS Dashboard
    [Arguments]  ${ocp_user_name}  ${ocp_user_pw}  ${ocp_user_auth_type}
    # Wait until we are in the OpenShift auth page or already in Dashboard
-   ${expected_text_list}=    Create List    Log in with
+   ${expected_text_list}=    Create List    Log in with    Data Science Projects
    Wait Until Page Contains A String In List    ${expected_text_list}
    ${oauth_prompt_visible}=  Is OpenShift OAuth Login Prompt Visible
    IF  ${oauth_prompt_visible}  Click Button  Log in with OpenShift
@@ -84,7 +84,7 @@ Login To RHODS Dashboard
    IF  ${login-required}  Login To Openshift  ${ocp_user_name}  ${ocp_user_pw}  ${ocp_user_auth_type}
    ${authorize_service_account}=  Is rhods-dashboard Service Account Authorization Required
    IF  ${authorize_service_account}  Authorize rhods-dashboard service account
-   Navigate To Page    Applications    Enabled
+   Navigate To Page    Applications    Enabled    timeout=10s
 
 Logout From RHODS Dashboard
     [Documentation]  Logs out from the current user in the RHODS dashboard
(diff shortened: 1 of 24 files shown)


ods_ci/tests/Tests/400__ods_dashboard/410__ods_dashboard_projects/410__ods_dashboard_projects_edit.robot
```diff
@@ -105,17 +105,16 @@ Project Suite Setup
     Set Library Search Order    SeleniumLibrary
     RHOSi Setup
     Launch Data Science Project Main Page
+    Delete Data Science Projects From CLI    ${PROJECTS_TO_DELETE}
     Create Data Science Project    title=${PRJ_TITLE}    description=${PRJ_DESCRIPTION}
-    ...    resource_name=${PRJ_RESOURCE_NAME}
+    ...    resource_name=${PRJ_RESOURCE_NAME}    existing_project=${TRUE}
```
bdattoma, line 110: is it because the project is always already present (e.g., Suite/Test Setup)? or just due to sequential run of the tests?

manosnoam: Due to multiple runs, allowing the reuse of the project saves time

bdattoma: what do you mean by runs? Different jobs?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -76,15 +76,15 @@ Authorize rhods-dashboard service account
 Login To RHODS Dashboard
    [Arguments]  ${ocp_user_name}  ${ocp_user_pw}  ${ocp_user_auth_type}
    # Wait until we are in the OpenShift auth page or already in Dashboard
-   ${expected_text_list}=    Create List    Log in with
+   ${expected_text_list}=    Create List    Log in with    Data Science Projects
```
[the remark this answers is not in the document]
bdattoma: hm probably this whole kw needs a review, including expected behavior.
This "data science projects", refers to the one in the left menu or it means the user is already in the "Data science projects" page? Does it make sense to expect the user to see "data science projects" at login step? maybe yes if the login is not required (expected?)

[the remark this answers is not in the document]
manosnoam: I did not change the logic of this KW, I reverted it back to the original behavior (that was there before my previous PR):
`Create List    Log in with    Data Science Projects`

I don't mind changing it, and it actually works without "Data Science Projects", but keeping "Data Science Projects":
1) Does not affect performance.
2) It might be required when opening the Dashboard with an active login session, where the "Log in with" message does not exist.

ods_ci/tests/Tests/400__ods_dashboard/410__ods_dashboard_projects/410__ods_dashboard_projects_edit.robot
manosnoam: Multiple execution of ODS-CI script, either via Jenkins or locally.
The option `existing_project=${TRUE}` is commonly used in our tests, and I made it to reduce time if a test can reuse an existing project.

manosnoam:
Also passed smoke tests: job/rhods-ci-pr-test/3249

nehachopra27 (approved):

ods_ci/tests/Tests/400__ods_dashboard/410__ods_dashboard_projects/410__ods_dashboard_projects_edit.robot
bdattoma: hm yeah I remember that for re-using in the same run.

I-d argue that it is a good behavior to re-use over multiple ods-ci runs. I think every run should find the cluster clean from previous execution otherwise it would miss covering important pieces like project creation

manosnoam: Our tests do not always use `existing_project=${TRUE}`. However, in some test cases it is more efficient to reuse the project.

FedeAlonso (approved):

PR #1717, opened 2024-08-13
fix: add pod ready conditions when installing dependant operators to solve flaky issues in SNO scenarios

CFSNM:
There are some scenarios like SNO clusters that the Service Mesh operator does not reconcile on time to run the operator tests, and we are getting failures. Adding pod waiting conditions for the dependant operators to avoid this race condition

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -22,6 +22,7 @@ ${DSCI_NAME} =    default-dsci
 ${SERVERLESS_OP_NAME}=     serverless-operator
 ${SERVERLESS_SUB_NAME}=    serverless-operator
 ${SERVERLESS_NS}=    openshift-serverless
+${OPENSHIFT_OPERATORS_NS}=    openshift-operators
 ${SERVICEMESH_OP_NAME}=     servicemeshoperator
 ${SERVICEMESH_SUB_NAME}=    servicemeshoperator
 ${AUTHORINO_OP_NAME}=     authorino-operator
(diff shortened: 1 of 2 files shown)


MarianMacik (approved):
I am glad it worked!

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -407,6 +408,8 @@ Install Authorino Operator Via Cli
           ...    type=CatalogSourcesUnhealthy    status=False
           ...    reason=AllCatalogSourcesHealthy    subcription_name=${AUTHORINO_SUB_NAME}
           ...    retry=150
+    Wait For Pods To Be Ready    label_selector=control-plane=authorino-operator
+          ...    namespace=${OPERATORS_NS}
```
tarukumar, line 412: Add support to validate authernio webhoook pod as well , label selector used for that `authorino-component=authorino-webhooks` or `app=authorino`.

The reason is if webhook pod is not working properly it will cause problem if any resource try to use authernino

CFSNM: done @tarukumar !

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -22,6 +22,7 @@ ${DSCI_NAME} =    default-dsci
 ${SERVERLESS_OP_NAME}=     serverless-operator
 ${SERVERLESS_SUB_NAME}=    serverless-operator
 ${SERVERLESS_NS}=    openshift-serverless
+${OPERATORS_NS}=    openshift-operators
```
tarukumar, line 25: could you change the name of the variable something like `OPENSHIFT_OPERATORS_NS` . This is because to avoid confusion between the rhoai operator ns and other as of today we use `OPERATOR_NAMESPACE` for `redhat-ods-operator`

CFSNM: done @tarukumar !

tarukumar (approved):

MarianMacik (approved):

PR #1719, opened 2024-08-14
fix: solve service mesh state unmanaged test failures

CFSNM:
The Service Mesh Unmanaged test was failing due to be executed after the TrustedCA tests. These cause a reconciliation loop in the operator, so the control plane instance was recreated before running the unmanaged test


manosnoam (approved):
lgtm

zdtsw (approved):

mattmahoneyrh (approved):

PR #1720, opened 2024-08-14
fix: solve must gather test issues

CFSNM:
When moving the must gather script to another location underl ods-ci repo, we lost the permissions to execute it. Adding them again

### DIFF
--- a/ods_ci/tests/Resources/CLI/MustGather/MustGather.resource
+++ b/ods_ci/tests/Resources/CLI/MustGather/MustGather.resource
@@ -10,6 +10,12 @@ Resource         ../../Common.robot
 Get must-gather Logs
     [Documentation]    Runs the must-gather image and obtains the ODH/RHOAI logs
     ${output}=    Run process    tests/Resources/CLI/MustGather/get-must-gather-logs.sh     shell=yes
+    IF  "${output.rc}" == "{0}"
+        Log To Console     ${output.stdout}
+    ELSE
+        Log To Console     ${output.stderr}
+    END
+    Should Be Equal As Integers	  ${output.rc}	 0
     Should Not Contain    ${output.stdout}    FAIL
     ${must-gather-dir}=  Run     ls -d must-gather.local.*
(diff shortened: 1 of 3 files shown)


ods_ci/tests/Resources/CLI/MustGather/get-must-gather-logs.sh
```diff
@@ -3,6 +3,8 @@
 # process buffer and prevents the script from running further.
 oc adm must-gather --image=quay.io/modh/must-gather@sha256:1bd8735d715b624c1eaf484454b0d6d400a334d8cbba47f99883626f36e96657 &> must-gather-results.txt
 
+cat must-gather-results.txt
+
```
MarianMacik, line 7: Do we want to print this every time or was this used just for debugging this?

CFSNM: good catch, I think we can remove it, because the output can be so big depending on the cluster

ods_ci/tests/Resources/CLI/MustGather/MustGather.resource
```diff
@@ -10,6 +10,12 @@ Resource         ../../Common.robot
 Get must-gather Logs
     [Documentation]    Runs the must-gather image and obtains the ODH/RHOAI logs
     ${output}=    Run process    tests/Resources/CLI/MustGather/get-must-gather-logs.sh     shell=yes
+    IF  ${output.rc} == 0
+        Log To Console     ${output.stdout}
+    ELSE
+        Log To Console     ${output.stderr}
+    END
```
bdattoma, line 17: maybe you don't need the IF/ELSE? just try printing the output variable?

CFSNM: changed, thanks @bdattoma

CFSNM: removed the log to console because showing the output:

<result object with rc 0>

and its useless

jstourac (approved):

MarianMacik (approved):

PR #1724, opened 2024-08-15
Fix Selector Dropdown Xpath in `Switch Model Serving Project` Keyword

manosnoam:
Tested both in RHOAI 2.12 and in ODH:

image


tonyxrmdavidson (approved):
/lgtm

jstourac (approved):

PR #1727, opened 2024-08-15
Check if trusty is installed or not

nehachopra27:
Upgrade tests for drift seem to be failing in 2.12 Jenkins. By default trustyai component is in Removed state as it is a tech preview. After upgrade we need to enable it.

### DIFF
--- a/ods_ci/tests/Tests/1200__ai_explainability/1203__trustyai_drift_metrics.robot
+++ b/ods_ci/tests/Tests/1200__ai_explainability/1203__trustyai_drift_metrics.robot
@@ -28,6 +28,8 @@ Run Drift Metrics Tests
 Prepare TrustyAi-tests Test Suite
     [Documentation]    Prepare trustyai-tests E2E Test Suite
     Log To Console    "Prepare Test Suite"
+    ${TRUSTY} =    Is Component Enabled    trustyai    ${DSC_NAME}
+    IF    "${TRUSTY}" == "false"    Enable Component    trustyai
     Drift Setup
     RHOSi Setup
 


apodhrad (approved):

ods_ci/tests/Tests/1200__ai_explainability/1203__trustyai_drift_metrics.robot
```diff
@@ -28,6 +28,8 @@ Run Drift Metrics Tests
 Prepare TrustyAi-tests Test Suite
     [Documentation]    Prepare trustyai-tests E2E Test Suite
     Log To Console    "Prepare Test Suite"
+    ${TRUSTY} =    Is Component Enabled    trustyai    ${DSC_NAME}
+    IF    "${TRUSTY}" == "false"    Enable Component    trustyai
```
jstourac, line 32: Should we put this back to the original state during the TS teardown?

What is the approach here - are we going to ignore the component states that are provided by the Jenkins during the RHOAI installation? Do we still want them at all then?

tonyxrmdavidson (approved):
/lgtm
I think @jstourac's question is valid, but doesn't need to hold this back. Once that descision is made it can actioned via another PR.

ods_ci/tests/Tests/1200__ai_explainability/1203__trustyai_drift_metrics.robot
nehachopra27: During Upgrade testing, the Jenkins components are not used as we do manual upgrades and run smoke test after that.

bdattoma: one danger with this approach is that if the component status changes (mistakenly) as part of the upgrade, we won't notice it

Maybe, I'd remove this patch from master branch after creating the releases/2.12.0 branch. Wdyt?

PR #1728, opened 2024-08-16
Update to README

tonyxrmdavidson:
This PR explains how to use the test to set up for further testing development.

### DIFF
--- a/ods_ci/tests/Tests/1300__model_registry/Model_Registry_Testing.md
+++ b/ods_ci/tests/Tests/1300__model_registry/Model_Registry_Testing.md
@@ -20,6 +20,8 @@ To review and run it locally you will need the following:
   ```bash
   ./odstest --install-identity-providers
   ```
+  Note: once you install the identity providers in might take up to 5 mins to become active.
+
 
 - Create and prepare the test-variables.yml file by using the template file test-variables.yml.example file. The following will need to be set.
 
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/1300__model_registry/Model_Registry_Testing.md
```diff
@@ -20,6 +20,8 @@ To review and run it locally you will need the following:
   ```bash
   ./odstest --install-identity-providers
   ```
+  Note: once you install the identity providers in might take upto 15 mins to become active.
```
jstourac, line 23: Hm, is this truly true? I think that in our Jenkins jobs we wait just 3 minutes :thinking: but can't really argue here as I never ever checked the code.

jiridanek, line 23:
```suggestion
  Note: once you install the identity providers in might take up to 15 mins to become active.
```

tonyxrmdavidson: I don't really know Jan, I had a bit of a wait when I was running locally. It turned out though I had caching issues when trying to log back in so this might be wilding over the mark.  If you are sure it is less than that I'd be happy to go with your suggestion.  Maybe 5 mins at the outside?

jstourac (approved):

jiridanek (approved):

PR #1729, opened 2024-08-16
[fix] Expected string of the final results

jstourac:
After some of the package update, there is now preserved numpy float64 data format in the printed results which in earler version weren't.

----

CI:
image


jiridanek:
I don't know if the new behavior is correct, can't review ;(

jstourac:
> I don't know if the new behavior is correct, can't review ;(

Then let's keep it failing then :smile: ... I don't think it is wrong... it's just the way what data type is returned and printed... I can easily put `float()` into the `.append` command and the output will be without the `np.float64`, but this seems easy as I don't have access to the lugi0's repo...

tonyxrmdavidson (approved):
Not 100% sure I know what it is doing, but since it all seems to be passing, /lgtm.

apodhrad (approved):

PR #1730, opened 2024-08-16
Support disconnected environment for Pipeline SDK tests

diegolovison:
Support disconnected environment for Pipeline SDK tests.

Tested on the disconnected environment by setting `PIP_INDEX_URL`.
Tested on a cluster with access to the internet without `PIP_INDEX_URL`.

### DIFF
--- a/ods_ci/libs/DataSciencePipelinesKfp.py
+++ b/ods_ci/libs/DataSciencePipelinesKfp.py
@@ -2,9 +2,11 @@ import importlib
 import json
 import os
 import sys
+import tempfile
 import time
 
 from DataSciencePipelinesAPI import DataSciencePipelinesAPI
+from kfp import compiler
 from kfp.client import Client
 from robotlibcore import keyword
 
@@ -89,7 +91,17 @@ class DataSciencePipelinesKfp:
 
     @keyword
     def create_run_from_pipeline_func(
-        self, user, pwd, project, source_code, fn, pipeline_params={}, current_path=None, route_name="ds-pipeline-dspa"
+        self,
+        user,
+        pwd,
+        project,
+        source_code,
+        fn,
+        pipeline_params={},
+        current_path=None,
+        route_name="ds-pipeline-dspa",
+        pip_index_url=None,
+        pip_trusted_host=None,
     ):
         print(f"pipeline_params: {pipeline_params}")
         client, api = self.get_client(user, pwd, project, route_name)
@@ -101,7 +113,7 @@ class DataSciencePipelinesKfp:
         if current_path is None:
             current_path = os.getcwd()
         my_source = self.import_souce_code(f"{current_path}/tests/Resources/Files/pipeline-samples/v2/{source_code}")
-        pipeline = getattr(my_source, fn)
+        pipeline_func = getattr(my_source, fn)
 
         # pipeline_params
         # there are some special keys to retrieve argument values dynamically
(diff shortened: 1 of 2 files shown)


jstourac:
@diegolovison just to be sure - do I understand correctly that since we're compiling the pipeline code here ourselves, this change will be still needed even once the https://issues.redhat.com/browse/RHOAIENG-6376 is fixed, correct?

ods_ci/tests/Tests/1100__data_science_pipelines/1100__data-science-pipelines-kfp.robot
```diff
@@ -95,8 +95,19 @@ End To End Pipeline Workflow Using Kfp
     IF    ${ray} == ${TRUE}
         Setup Kueue Resources    ${project}    cluster-queue-user    resource-flavor-user    local-queue-user
     END
-    ${run_id}    Create Run From Pipeline Func    ${username}    ${password}    ${project}
-    ...    ${python_file}    ${method_name}    pipeline_params=${pipeline_params}
+    # the run_robot_test.sh is sending the --variablefile ${TEST_VARIABLES_FILE}
+    # there is no way to double-check if a the key is presented there
+    # the workaround is the following
+    # https://issues.redhat.com/browse/RHOAIENG-6376
```
jstourac, line 101:
```suggestion
    # The run_robot_test.sh is sending the --variablefile ${TEST_VARIABLES_FILE}
    # which may contain the `PIP_INDEX_URL` and `PIP_TRUSTED_HOST` variables,
    # e.g. for disconnected testing. In such case we need to pass them to the
    # `Create Run From Pipeline Func` so that it knows when to amend the pipeline
    # code for the https://issues.redhat.com/browse/RHOAIENG-6376 issue.
```

diegolovison:
> this change will be still needed even once the https://issues.redhat.com/browse/RHOAIENG-6376 is fixed, correct?

Yes.

jstourac (dismissed):

ods_ci/tests/Tests/1100__data_science_pipelines/1100__data-science-pipelines-kfp.robot
```diff
@@ -95,8 +95,19 @@ End To End Pipeline Workflow Using Kfp
     IF    ${ray} == ${TRUE}
         Setup Kueue Resources    ${project}    cluster-queue-user    resource-flavor-user    local-queue-user
     END
-    ${run_id}    Create Run From Pipeline Func    ${username}    ${password}    ${project}
-    ...    ${python_file}    ${method_name}    pipeline_params=${pipeline_params}
+    # The run_robot_test.sh is sending the --variablefile ${TEST_VARIABLES_FILE} which may contain the `PIP_INDEX_URL`
+    # and `PIP_TRUSTED_HOST` variables, e.g. for disconnected testing. In such case there is no way to detect if
+    # `PIP_INDEX_URL` or `PIP_TRUSTED_HOST was definied. The only way is by calling Run Keyword And Return Status
+    # Log If it returns true, means that we are running the test in a disconnected environment.
+    ${pip_index_url_definied}    Run Keyword And Return Status    Log    ${PIP_INDEX_URL}
```
bdattoma, line 102: that is not the only way, but there is a built-in function in RF: https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Get%20Variable%20Value

diegolovison: It will end up in the same situation. Logging or getting will return an error if the variable doesn't exists.
The real way to solve is by checking if the variable exists and I didn't find any function.

bdattoma (approved):

ods_ci/libs/DataSciencePipelinesKfp.py
```diff
@@ -118,7 +130,27 @@ def create_run_from_pipeline_func(
...
+                package_path=pipeline_package_path,
+            )
+
+            if pip_index_url is not None:
+                assert pip_trusted_host is not None
+                with open(pipeline_package_path, "r") as file:
+                    file_content = file.read()
+                file_content = file_content.replace(
+                    "python3 -m pip install",
+                    f"python3 -m pip install --index-url {pip_index_url} --trusted-host {pip_trusted_host}",
```
apodhrad, line 146: Hi @diegolovison , could we remove `--trusted-host {pip_trusted_host}`?
We would like to cover also self-signed certs.

apodhrad (changes requested):

ods_ci/libs/DataSciencePipelinesKfp.py
diegolovison: No. This is related to https://github.com/kubeflow/pipelines/pull/10720

apodhrad (approved):

PR #1731, opened 2024-08-18
Fix Test `Verify Documentation Links HTTP Status Code`

manosnoam:
image


CFSNM (dismissed):

ods_ci/tests/Tests/400__ods_dashboard/401__ods_dashboard.robot
```diff
@@ -129,17 +130,16 @@ Verify CSS Style Of Getting Started Descriptions
...
+    ...    ProductBug: RHOAIENG-11451
     [Tags]    Sanity    Tier1
     ...       ODS-327    ODS-492
-    ...       AutomationBugOnODH
-    # TODO: In ODH the expected Docs links are:
-    # https://opendatahub.io/community
-    # https://opendatahub.io/docs
+    ...       ExcludeOnODH
     ${links}=  Get RHODS Documentation Links From Dashboard
-    Documentation Links Should Be Equal To The Expected Ones   actual_links=${links}  expected_links=${DOC_LINKS_EXP}
```
jstourac, line 142: Is this still used somewhere else? Do we want to remove this keyword?

ods_ci/tests/Tests/400__ods_dashboard/401__ods_dashboard.robot
```diff
@@ -129,17 +130,16 @@ Verify CSS Style Of Getting Started Descriptions
...
+    # Compare Doc Links only by number, since ODH and RHOAI have diffrent URLs (but same count)
+    Lists Size Should Be Equal    ${links}    ${DOC_LINKS_EXP}
```
jstourac, line 142: I think I don't understand this one - this test is now excluded on ODH, so why we're reducing the actual check here from the whole content to the number of elements only now?

ods_ci/tests/Tests/400__ods_dashboard/401__ods_dashboard.robot
manosnoam: It is not used, but I wanted to keep this keyword if it will be required to be used later. 
wdyt, should we remove it ?

ods_ci/tests/Tests/400__ods_dashboard/401__ods_dashboard.robot
manosnoam: The number of "Documentation" links is the same in ODH and in Dashboard. They only lead to other URL addresses, so validating the number of links in UI is as expected.

In the test we also validate that the URL is not 403/404 (this is actually the name of the test: `Verify Documentation Links HTTP Status Code`)

jstourac: My point above was that this test won't be run on ODH, so why is this change done? Based on your comments it's because of the RHOAI vs ODH differences, but I don't understand why it matters since this test won't run on ODH anymore. That is what I don't understand here.

manosnoam: It currently fails on ODH due to bug RHOAIENG-11451 (ODH doc leads to 404), so I added "ExcludeOnODH".
We do not have "ProductBug" for ODH...
Do you think I should change the tag from "ExcludeOnODH" to "ProductBug" ?

jstourac: Hm, okay, so the `ExcludeOnODH` is just a temporary thing... well, the `ProductBug` might be misleading as it's not a product issue... so maybe adding a comment on why we added the `ExcludeOnODH` tag and that it is supposed to be removed in the future once the issue is gone, it would be great.

manosnoam: On second thought maybe it is better to remove the "ExcludeOnODH" tag, otherwise it will never be run on ODH as you noticed, which is why we did not see bug RHOAIENG-11451 up until now (I wonder for how long this bug from).
We better have failures seen with comments of ProductBug, rather than unseen failures.

jstourac (approved):

CFSNM (approved):

PR #1732, opened 2024-08-19
fix the deployment parameter for Model serving

tarukumar:
This is to make consistent with other keyword


CFSNM (approved):

jstourac (approved):
FTR this is the other keyword: `Verify Model Inference` https://github.com/red-hat-data-services/ods-ci/pull/1732/files#diff-3f0da500a83b2db0b933a81882fc35a424e2c53bd4abe276a6dee3faa28e9ae3R325

PR #1738, opened 2024-08-21
Fix "SKIPPED" status in Post upgrade suite

bdattoma:
Currently, some tests in post upgrade suite have "SKIPPED" status because a piece of the teardown gets skipped if RHOAI operator is not Managed. 
This PR replaces the call to SKIP keyword with IF statement so to avoid the SKIPPED status which doesn't reflect the right test status

PR validation:
- pre-upgrade suite + post-upgrade suite (without actual upgrade): `rhods-ci-pr-test/3270` OK - 1 failure expected for automation issue

image


CFSNM (approved):

asanzgom:
Waiting for rhods-ci-pr-test/3270/ to finish successfully to approve

PR #1754, opened 2024-08-28
Fix keyword to watch SM cluster provisioning logs and increase timeout

bdattoma:
Content of the PR:
1. change logic of `Watch Hive Install Log` keyword to avoid RHOAIENG-10829 bug
2. increase timeout to 60m

This won't necessarily fix some of the failures we have seen in CI while provisioning clusters, but would:
1. ensure to provide provisioning logs when available
2. reduce the chance of failure in case provisioning takes more than the fixed timeout

PR validation:
1. PSI provisioning  OK
image

2. IBM Cloud provisioning OK
image

### DIFF
--- a/ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
+++ b/ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
@@ -155,7 +155,7 @@ Watch Hive Install Log
     IF    ${use_cluster_pool}
         ${label_selector} =    Set Variable    hive.openshift.io/clusterpool-name=${pool_name}
     END
-    ${label_selector}=    Catenate    SEPARATOR=    ${label_selector}    ,hive.openshift.io/job-type=provision
+    ${label_selector} =    Catenate    SEPARATOR=    ${label_selector}    ,hive.openshift.io/job-type=provision
     ${logs_cmd} =     Set Variable    oc logs -f -l ${label_selector} -n ${namespace}
     Wait For Pods To Be Ready    label_selector=${label_selector}    namespace=${namespace}    timeout=5m
     TRY
--- a/ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
+++ b/ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
@@ -151,10 +151,11 @@ Create Floating IPs
 
 Watch Hive Install Log
     [Arguments]    ${pool_name}    ${namespace}    ${hive_timeout}=60m
-    ${label_selector}=    Set Variable    hive.openshift.io/cluster-deployment-name=${cluster_name},hive.openshift.io/job-type=provision
+    ${label_selector}=    Set Variable    hive.openshift.io/cluster-deployment-name=${cluster_name}
     IF    ${use_cluster_pool}
         ${label_selector}=    Set Variable    hive.openshift.io/clusterpool-name=${pool_name}        
     END
+    ${label_selector}=    Catenate    ${label_selector}    ,hive.openshift.io/job-type=provision
     ${logs_cmd} =     Set Variable    oc logs -f -l ${label_selector} -n ${namespace}
     TRY
         Wait For Pods To Be Ready    label_selector=${label_selector}    namespace=${namespace}    timeout=5m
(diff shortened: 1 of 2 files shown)


ods_ci/tests/Resources/Common.robot
```diff
@@ -477,7 +477,7 @@ Run And Watch Command
 Check Process Output and Status
   [Documentation]    Helper keyward for 'Run And Watch Command', to tail proccess and check its status
   [Arguments]    ${process_id}    ${process_log}    ${temp_log}
-  Log To Console    .    no_newline=true
+  # Log To Console    .    no_newline=true
```
kobihk, line 480: If we don't want this log to console maybe we can delete it

bdattoma: true, missed it. ty

kobihk (approved):

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -150,43 +150,26 @@ Create Floating IPs
...
+    EXCEPT
+        Log To Console    ERROR: Check Hive Logs if present or you may have hit timeout ${hive_timeout}.
+    END
+    Should Be Equal As Integers    ${return_code}    ${0}
+    ${hive_pods_status} =    Run And Return Rc
+    ...    oc get pod -n ${namespace} --no-headers | awk '{print $3}' | grep -v 'Completed'
+    IF    ${hive_pods_status} != 0
+        Log    All Hive pods in ${namespace} have completed    console=True
     END
-    Should Contain    ${new_log_data}    install completed successfully
```
manosnoam, line 189: @bdattoma If you remove this check, it will not verify that the installation was completed successfully, and it may happen that the pod was completed but the installation failed.

Can you try to use `Run And Watch Command` with the argument `output_should_contain=install completed successfully`

ods_ci/tests/Resources/Common.robot
```diff
@@ -477,7 +477,6 @@ Run And Watch Command
 Check Process Output and Status
   [Documentation]    Helper keyward for 'Run And Watch Command', to tail proccess and check its status
   [Arguments]    ${process_id}    ${process_log}    ${temp_log}
-  Log To Console    .    no_newline=true
```
manosnoam, line 480: Why to remove this line ?
It is a good indication that something is going on ...

manosnoam (changes requested):

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
bdattoma: > pod was completed but the installation failed.


I think the pod fails in case of any install error
I can try adding that arg

bdattoma: done

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -201,14 +185,16 @@ Wait For Cluster To Be Ready
     END
     ${install_log_file} =    Set Variable    ${artifacts_dir}/${cluster_name}_install.log
     Create File    ${install_log_file}
-    Run Keyword And Ignore Error    Watch Hive Install Log    ${pool_namespace}    ${install_log_file}
+    Run Keyword And Continue On Failure    Watch Hive Install Log    ${pool_name}    ${pool_namespace}
```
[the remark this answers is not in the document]
bdattoma: I think it is better to see the failure otherwise any problem with the automation would be hidden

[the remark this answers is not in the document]
manosnoam: ack, if you see false negative failure in the watch (which happened to me a few times), then consider changing back to `Run Keyword And Ignore Error `

manosnoam (approved):

kobihk (approved):

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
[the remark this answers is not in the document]
bdattoma: okay Noam, thanks!

manosnoam:
Hi @bdattoma although you added `output_should_contain=install completed successfully` you should also fail fast if fatal error occurs, as in this job job/rhods/job/rhods-ci-pr-test/3312/console

Please also add `output_should_not_contain=fatal msg`

PR #1758, opened 2024-08-30
Update DW UI tests to accommodate changes introduced in 2.13 and instability tests

Srihari1192:
[RHOAIENG-12042](https://issues.redhat.com/browse/RHOAIENG-12042)

- Updated DW UI tests to accommodate changes introduced in 2.13 for project selection and refresh interval toggle

- Increasing timeout for instability tests


ChughShilpa (approved):

Srihari1192:
Refactored and updated Project selection changes as already related changes merged PR https://github.com/red-hat-data-services/ods-ci/pull/1763

ods_ci/tests/Resources/Page/DistributedWorkloads/WorkloadMetricsUI.resource
```diff
@@ -61,7 +61,7 @@ Select Refresh Interval
     [Arguments]    ${refresh_interval}
     Wait Until Element Is Visible  ${REFRESH_INTERVAL_XP}    timeout=20
     Click Element     ${REFRESH_INTERNAL_MENU_XP}
-    Click Element     xpath=//button[text()="${refresh_interval}"]
+    Click Element     xpath=//li[@class="pf-v5-c-menu__list-item" and @data-testid="${refresh_interval}"]
```
jstourac, line 64: Is it really needed to combine two conditions here to be sure we get the appropriate DOM element? If so, then I suppose that the `data-testid` field value isn't designed properly. Maybe something to come up with dashboard guys whether it can't be improved? It would be definitely better we don't need to rely on this `class` field value.

Srihari1192: @jstourac  Thanks for reviewing. Earlier Refresh interval selection designed with button tag, now its changed to list. It can be refactor without relying on class `//div[@data-ouia-component-type='PF5/Select' and text()="${refresh_interval}"]` wdyt?

jstourac: Hi @Srihari1192, it's a bit of pity that in this alternative we don't use `data-testid` at all then. And yet, `data-testid` should be designed for the use in tests, so it's interesting.

Please, bear in mind that I have no knowledge about this part of the UI, so I am leaving the decision up to you or you other guys who know it better there. I was just trying to point out that it's weird there is some `data-testid` field in the element, yet it's hardly usable for our testing purpose, so maybe it deserves some review and improvement :slightly_smiling_face:

Srihari1192: @jstourac Updated xpath to `//button[@role="option" and contains(., "${refresh_interval}")]`  by avoiding `data-testid`  Thanks

ChughShilpa (approved):
lgtm

sutaakar (approved):

PR #1764, opened 2024-08-30
Fix xpath used to select project in elyra tests

jgarciao:
Signed-off-by: Jorge Garcia Oncins <jgarciao@redhat.com>


jgarciao:
This is already done in 1763

PR #1770, opened 2024-09-02
Fix for Authorino TC 

mwaykole:


mwaykole:
tested : `rhoai-test-flow/51`

mwaykole:
rhoai-test-flow/52/console

nehachopra27 (approved):

aloganat (approved):

lugi0:
@mwaykole I don't think this fix works for all cases. 
You've removed the `protocol=http` line from a bunch of calls, thus forcing the requests to always be made with grpc - I think we should test both protocols to make sure that http works.
You've also updated the token handling to be 
```
IF    "${token}" != "${None}"
        IF    "${protocol}" == "grpc"
            ${header}=   Set Variable   "Authorization: Bearer ${token}" -H ${header}
            ELSE
                Set To Dictionary    ${header}    Authorization    Bearer ${token}
        END
    END
```
The `ELSE` is indented too much, and I *think* this call will fail if the protocol isn't grpc.

Can you please reopen and implement test flows for http calls as well?

mwaykole:
hi @lugi0 

protocol is grpc that was the reason i changed it , in future if we face issue if needed ill fix it 
else it will fail at 
` ${runtime_details}=    Set Variable    ${RUNTIME_FORMATS}[${runtime}][endpoints][${inference_type}][${protocol}]`
as it will not get the protocol 


the o/p of `{RUNTIME_FORMATS}[${runtime}][endpoints][${inference_type}]` is below
```{'grpc': {'endpoint': 'fmaas.GenerationService/Generate', 'header': 'mm-model-id: flan-t5-small-caikit', 'body': "{'requests': [{'text':'At what temperature does water boil?'}]}", 'args': 'proto=text-generation-inference/proto/generation.proto', 'response_fields_map': {'response': 'responses', 'response_tokens': 'generatedTokenCount', 'response_text': 'text'}}}
```

lugi0:
>protocol is grpc that was the reason i changed it 

I'm not sure what you mean? the protocol was explicitly set to `http` in the test cases and you've changed it to use `grpc`.
If we want to test both protocols it's fine, but I don't think you should just remove the `http` steps that were in the test cases.

>else it will fail at
 `${runtime_details}=    Set Variable    ${RUNTIME_FORMATS}[${runtime}][endpoints][${inference_type}][${protocol}]`
as it will not get the protocol

Why does it fail there? you've pasted a response that shows that `grpc` is present as the ${protocol} index when doing a grpc calls, and I expect the same to be true for `http`. If it is not, then this call needs to be fixed for `http` rather than avoiding `http` calls entirely.

I also still think the `IF` block you've added is formatted wrongly and the `ELSE` step could fail when executing this code. Please confirm if that is the case and fix the formatting if needed.

mwaykole:
> > protocol is grpc that was the reason i changed it
> 
> I'm not sure what you mean? the protocol was explicitly set to `http` in the test cases and you've changed it to use `grpc`. If we want to test both protocols it's fine, but I don't think you should just remove the `http` steps that were in the test cases.
> 
> > else it will fail at
> > `${runtime_details}=    Set Variable    ${RUNTIME_FORMATS}[${runtime}][endpoints][${inference_type}][${protocol}]`
> > as it will not get the protocol
> 
> Why does it fail there? you've pasted a response that shows that `grpc` is present as the ${protocol} index when doing a grpc calls, and I expect the same to be true for `http`. If it is not, then this call needs to be fixed for `http` rather than avoiding `http` calls entirely.
> 
> I also still think the `IF` block you've added is formatted wrongly and the `ELSE` step could fail when executing this code. Please confirm if that is the case and fix the formatting if needed.

yes it makes sense ill fix it 
about the formatting its not bcs of that the TC is not failing

PR #1773, opened 2024-09-03
Fix variable evaluation in "Verify Model Inference" and associated keywords

lugi0:
This if block used to fail like this

```
Invalid IF condition: Evaluating expression 'UI == 'UI'' failed: NameError: name 'UI' is not defined nor importable as module

Variables in the original expression '${deployment_mode} == 'UI'' were resolved before the expression was evaluated. Try using '$deployment_mode == 'UI'' syntax to avoid that. See Evaluating Expressions appendix in Robot Framework User Guide for more details.
```

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
@@ -323,9 +323,9 @@ Get Model Inference
 Verify Model Inference
     [Documentation]    Verifies that the inference result of a model is equal to an expected output
     [Arguments]    ${model_name}    ${inference_input}    ${expected_inference_output}    ${token_auth}=${FALSE}
-    ...    ${project_title}=${NONE}    ${deployment_mode}='UI'   ${kserve_mode}=Serverless
+    ...    ${project_title}=${NONE}    ${deployment_mode}=UI    ${kserve_mode}=Serverless
     ...    ${service_port}=${NONE}   ${end_point}=${NONE}
-    IF   ${deployment_mode} == 'UI'
+    IF   $deployment_mode == 'UI'
          Open Model Serving Home Page
          Switch Model Serving Project    ${project_title}
     END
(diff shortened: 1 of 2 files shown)


mwaykole (dismissed):

tarukumar (approved):

lugi0:
Verified in rhoai-test-flow/85 and rhoai-test-flow/87

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -263,12 +263,13 @@ Get Model Inference
     ...    model endpoint. If token authentication is needed for the model, ${token_auth} should be
     ...    set to ${TRUE}.
     [Arguments]    ${model_name}    ${inference_input}    ${token_auth}=${FALSE}   ${project_title}=${NONE}
-    ...    ${kserve_mode}=Serverless   ${deployment_mode}='UI'   ${service_port}=8888   ${end_point}=${NONE}
+    ...    ${kserve_mode}=Serverless   ${deployment_mode}=UI   ${service_port}=8888   ${end_point}=${NONE}
     ${curl_cmd}=     Set Variable    ${NONE}
     ${self_managed}=    Is RHODS Self-Managed
-    IF  ${deployment_mode} == 'UI'
+    IF    $deployment_mode == 'UI'
```
jstourac, line 269: I just wonder whether this change is truly necessary? E.g. on line 299, there is this:
```
IF    '${kserve_mode}' == 'Serverless'
```
So, isn't the change on line 266 ('UI' -> UI) good enough? :thinking:

mwaykole (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
lugi0: It all depends on how we want to treat strings; doing `'${kserve_mode}' == 'Serverless'` is casting the value of `${kserve_mode}` to string, while `${deployment_mode} == 'UI'` is not, which causes a failure when `UI == 'UI'` is evaluated. 
RF provides a special syntax for variables to access their actual values (i.e. $var), see: https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#special-variable-syntax

IMHO this is cleaner than casting each variable to a string before evaluation, but both options work (with the caveat that casting to string might mess with some special values, like `None`, `True`, `False` etc.).

jstourac: Yeah, I missed it was missing the enclosing `'` at the beginning, so as you said in your last sentence.
TBH, I would prefer to have a unified approach here using what is used for `kserve_mode` on the other line since it can then lead to a confusion why these are treated differently.

But I don't insist, I guess.

jstourac (approved):

bdattoma (approved):

PR #1779, opened 2024-09-04
Update to allow installation of the Python Client without external access

tonyxrmdavidson:
This commit will :
add keywords to download the python client binaries. 

- Upload the binaries to a running jupyter notebook. 
- Install Python Client on any cluster, including a disconnected cluster. 
- Delete the binaries when tearing down after test has completed.
- It will not push any binary files to the repository

There are commented out lines of code. This is required as the 'Create S3 Data Connection' keyword is currently broken due to changes in ODH-17. It is subject to another PR.

### DIFF
--- a/ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
+++ b/ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
@@ -22,6 +22,8 @@ ${MODELREGISTRY_BASE_FOLDER}=        tests/Resources/CLI/ModelRegistry
 ${EXAMPLE_ISTIO_ENV}=                ${MODELREGISTRY_BASE_FOLDER}/samples/istio/components/example_istio.env
 ${ISTIO_ENV}=                        ${MODELREGISTRY_BASE_FOLDER}/samples/istio/components/istio.env
 ${SAMPLE_ONNX_MODEL}=                ${MODELREGISTRY_BASE_FOLDER}/mnist.onnx
+${MR_PYTHON_CLIENT_FILES}=           ${MODELREGISTRY_BASE_FOLDER}/Python_Dependencies
+${MR_PYTHON_CLIENT_WHL_VERSION}=     model_registry==0.2.5a1
 ${SERVICE_MESH_MEMBER}=              ${MODELREGISTRY_BASE_FOLDER}/serviceMeshMember.yaml
 ${ENABLE_REST_API}=                  ${MODELREGISTRY_BASE_FOLDER}/enable_rest_api_route.yaml
 ${IPYNB_UPDATE_SCRIPT}=              ${MODELREGISTRY_BASE_FOLDER}/updateIPYNB.py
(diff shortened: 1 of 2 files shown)


tonyxrmdavidson:
Successful PR check here
rhods-ci-pr-test/3330/

CFSNM (approved):

diegolovison (dismissed):
IMHO, this should follow a different approach in airgap environment. 
Similar to what we are doing with Pipelines you provide an index and a trusted host for the pip command.
cc @apodhrad

tonyxrmdavidson:
> IMHO, this should follow a different approach in airgap environment. Similar to what we are doing with Pipelines you provide an index and a trusted host for the pip command. cc @apodhrad

I'm looking to have a test that can work in any environment from local to ODS, RHOAI, connected and disconnected without the need for manual setting any vars.

isinyaaa (approved):

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_model_serving.robot
```diff
@@ -260,9 +267,27 @@ Remove Model Registry
...
+    ${files}=    List Files In Directory    ${directory}
     FOR    ${file}    IN    @{files}
-        Run Process    rm    -f    ${CERTS_DIRECTORY}/${file}
+        Remove Files  ${directory}/${file}
+    END
+
+Download Python Client Dependencies
+    [Documentation]  Download the model-registry package for a specific platform
+    [Arguments]  ${destination}  ${package_version}
+    ${result}=    Run Process    command=pip download --platform=manylinux2014_x86_64 --python-version=3.9 --abi=cp39 --only-binary=:all: --dest=${destination} ${package_version}    # robocop: disable:line-too-long
```
diegolovison, line 281: > pip download

Can you explain how `pip download` will work in an airgap environment without specifying the index URL?

diegolovison: Are you uploading the file to a folder and then "downloading" from there? Is this the way?

diegolovison: Also, why set platform and abi? Why not use the defaults?
* Defaults to the platform of the running system
* If not specified, then the current interpreter abi tag is used

tonyxrmdavidson: This doesn't require a trusted host as there is no need for an external host when installing the Python Client. 

The process is as follows
- The environment running the test downloads the binary and dependencies into it's own environment.
- The test uploads the binary and dependencies to the Jupyter Notebook src directory, once it is running.
- The pip install targets the files already in Jupyter Notebook to install it without the need to goto a trusted host via the mirror.
- On completion of the test the binary and dependencies are deleted from both the Jupyter Notebook and the environment where the test is running.  They are never stored or pushed to any repository.

I also have the issue at the moment that I can't test using a trusted host on a disconnected cluster as Model Registry is not yet on RHOAI. It's also my understanding that you can't test on a disconnected cluster with ODH.  

I'm happy to change this over to what you are suggesting once a fully automated solution is available without the need for Manual intervention, for now though I prefer to have a test that will work anywhere without needing to manually set vars depending on the environment it is run.

CFSNM (approved):

PR #1786, opened 2024-09-05
Add Training operator upgrade tests for RHOAI 2.10

sutaakar:
Using DISTRIBUTED_WORKLOADS_RELEASE_ASSETS from 2.12 as 2.10 version doesn't contain sleep upgrade test.
Using newer release asset is ok as the binary contains just test resources


abhijeet-dhumal (commented):
/lgtm

abhijeet-dhumal (approved):

ChughShilpa (approved):
lgtm

PR #1789, opened 2024-09-06
DataSciencePipelines tag in the test cases

diegolovison:
Introducing the DataSciencePipelines tag because we are doing a migration to the next major version of Argo-Workflow and I need to run all tests that are related to Pipelines

### DIFF
--- a/ods_ci/tests/Tests/1100__data_science_pipelines/1100__data-science-pipelines-api.robot
+++ b/ods_ci/tests/Tests/1100__data_science_pipelines/1100__data-science-pipelines-api.robot
@@ -22,7 +22,7 @@ ${URL_TEST_PIPELINE_RUN_YAML}=                 https://raw.githubusercontent.com
 *** Test Cases ***
 Verify Pipeline Server Creation With S3 Object Storage
     [Documentation]    Creates a pipeline server using S3 object storage and verifies that all components are running
-    [Tags]    Smoke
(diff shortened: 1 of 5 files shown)


jstourac (approved):

ods_ci/tests/Tests/1100__data_science_pipelines/1100__data-science-pipelines-api.robot
```diff
@@ -22,7 +22,7 @@ ${URL_TEST_PIPELINE_RUN_YAML}=                 https://raw.githubusercontent.com
 *** Test Cases ***
 Verify Pipeline Server Creation With S3 Object Storage
     [Documentation]    Creates a pipeline server using S3 object storage and verifies that all components are running
-    [Tags]    Smoke
+    [Tags]    Smoke    DataSciencePipelines
```
jgarciao, line 25: This is not needed: we used to have the DataSciencePipelines tag but recently I spitted it to DataSciencePipelines-Backend, DataSciencePipelines-Dashboard and DataSciencePipelines-IDE. They are usually set at the top of the test suites, in the `Test Tags` section

If you want to run all DataSciencePipelines tests, you can do it with:
 `--extra-robot-args '-i DataSciencePipelines-*  '`

jgarciao (changes requested):

PR #1795, opened 2024-09-10
Bump kfp to 2.9.0 and kfp-kubernetes to 1.3.0

diegolovison:
Bump kfp to 2.9.0 and kfp-kubernetes to 1.3.0

How to review the PR:
* The YAML files were automatically generated by the command described in `pyproject.toml`. The command is calling the python file and the output of that call is the YAML.
* An instruction was added to  `pyproject.toml` to ensure no one will forget to generate the YAML files. As you can see, we forgot from 2.7 to 2.8
* `base_image` is now under each file. You can generate the YAML super easy by just running the `py` file
* Added the `if __name__ == "__main__":` method where it was not presented
* Add the `main` and the `compile` instructions where it was missing

Job used for testing: `job/devops/job/rhoai-test-flow/184`
* **IDE is failing because the server was unavailable. Tested locally**
* * ODS-2197 passed locally running on the same cluster `sh run_robot_test.sh --set-urls-variables true --extra-robot-args '-L DEBUG -i ODS-2197' --skip-oclogin`
* * ODS-2271 passed locally running on the same cluster `sh run_robot_test.sh --set-urls-variables true --extra-robot-args '-L DEBUG -i ODS-2271' --skip-oclogin`
* E2E is expected to be failing - AutomationBug

### DIFF
--- a/ods_ci/tests/Resources/Files/pipeline-samples/v2/pip_index_url/flip_coin_pip_index_url_compiled.yaml
+++ b/ods_ci/tests/Resources/Files/pipeline-samples/v2/pip_index_url/flip_coin_pip_index_url_compiled.yaml
@@ -277,7 +277,7 @@ deploymentSpec:
         - "\nif ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip ||\
           \ python3 -m ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1\
           \ python3 -m pip install --quiet --no-warn-script-location --index-url $PIP_INDEX_URL\
-          \ --trusted-host $PIP_TRUSTED_HOST 'kfp==2.7.0' '--no-deps' 'typing-extensions>=3.7.4,<5;\
+          \ --trusted-host $PIP_TRUSTED_HOST 'kfp==2.9.0' '--no-deps' 'typing-extensions>=3.7.4,<5;\
           \ python_version<\"3.9\"' && \"$0\" \"$@\"\n"
         - sh
         - -ec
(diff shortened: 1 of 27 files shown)


ods_ci/tests/Resources/Files/pipeline-samples/v2/pip_index_url/flip_coin_pip_index_url_compiled.yaml
```diff
@@ -277,7 +277,7 @@ deploymentSpec:
         - "\nif ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip ||\
           \ python3 -m ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1\
           \ python3 -m pip install --quiet --no-warn-script-location --index-url $PIP_INDEX_URL\
-          \ --trusted-host $PIP_TRUSTED_HOST 'kfp==2.7.0' '--no-deps' 'typing-extensions>=3.7.4,<5;\
+          \ --trusted-host $PIP_INDEX_URL 'kfp==2.9.0' '--no-deps' 'typing-extensions>=3.7.4,<5;\
```
jgarciao, line 280: This is wrong. In flip_coin_pip_index_url.py you need to declare pip_trusted host for each component, so when compiling the pipeline the resulting yaml file will have `--trusted-host $PIP_TRUSTED_HOST `

Something like this (for each component):
```
@dsl.component(base_image=common_base_image,
                            pip_index_urls=['$PIP_INDEX_URL'],
                            pip_trusted_host=['$PIP_TRUSTED_HOST'])
def random_num(low: int, high: int) -> int:
    """Generate a random number between low and high."""
    import random
    result = random.randint(low, high)
    print(result)
    return result
```

This change should be done for all pipelines setting pip_index_url

jgarciao (changes requested):

ods_ci/tests/Resources/Files/pipeline-samples/v2/pip_index_url/flip_coin_pip_index_url_compiled.yaml
diegolovison: Done

jgarciao (approved):

CFSNM (approved):

PR #1801, opened 2024-09-11
fix: DSProjects view wrong xpath

FedeAlonso:
There was a wrong xpath in the DSP view. This was discovered during the QGs execution for the 2.13 RC2 release.

Test results:
image


bdattoma (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
```diff
@@ -104,9 +104,8 @@ Project Should Be Listed
 
 Wait Until Project Is Listed
     [Documentation]    Waits until the DS projects appears in the list in DS project Home Page
-    [Arguments]     ${project_title}    ${timeout}=5s
-    Filter Projects By Name    ${project_title}
-    Wait Until Page Contains Element    xpath=//div//a[text()="${project_title}"]
+    [Arguments]     ${project_title}    ${timeout}=30s
+    Wait Until Page Contains Element    xpath=//table[@data-testid="project-view-table"]//td[@data-label="Name"]//a[text()="${project_title}"]
```
bdattoma, line 108: if the projects are spread on multiple pages, I think this will fail. Do you know if this happens for any test?

FedeAlonso: I jsut have tested it in one test case, but It will work the same way as before (there's no pagination option here) but at least the xpath is more readable (is easier to understand that we are looking for the project in a table)..
It's just a fix. If we want pagination, we should create some request for enhancement.

ConorOM1 (approved):

CFSNM (approved):

PR #1804, opened 2024-09-11
Update to reflect changes in model-registry-operator config files.

tonyxrmdavidson:
### This PR makes several changes.
- Updates the service route in 'modelregistry_v1alpha1_modelregistry.yaml' to disabled.
- Updates the 'mysql-db.yaml' with changes to reflect what has changed in model-registry-operator.
- In '1302_model_registry_model_serving.robot' the following changes have been made:
                1. hard coding workbench size
                2. adding tags to include test in ODH-Nightlies
                3. removing code referencing Istio.env as it is no longer needed
                4. removing the code to 'oc delete smm default -n odh-model-registries'

### Succesful pr-check
job/rhods-ci-pr-test/3347/

### DIFF
--- a/ods_ci/tests/Resources/CLI/ModelRegistry/samples/istio/components/istio.env
+++ b/ods_ci/tests/Resources/CLI/ModelRegistry/samples/istio/components/istio.env
@@ -1,0 +1,5 @@
+AUTH_PROVIDER=opendatahub-auth-provider
+DOMAIN=
+ISTIO_INGRESS=ingressgateway
+REST_CREDENTIAL_NAME=modelregistry-sample-rest-credential
+GRPC_CREDENTIAL_NAME=modelregistry-sample-grpc-credential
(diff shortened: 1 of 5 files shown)


ods_ci/tests/Resources/CLI/ModelRegistry/samples/istio/components/istio.env
```diff
@@ -1,4 +1,5 @@
 AUTH_PROVIDER=opendatahub-auth-provider
+DOMAIN=
```
nehachopra27, line 2: Does it works without any value for key Domain?

tonyxrmdavidson: Yes, the operator sets that now.

nehachopra27 (approved):
LGTM

CFSNM (approved):

bdattoma (approved):

PR #1809, opened 2024-09-12
RHOAIENG-12833 - Validate Image Pull Path Is From registry.redhat.io

mattmahoneyrh:
The purpose of this enhancement is to check that the Component Deployment image pull path is from registry.redhat.io, when expected to be so.

Note: kserve-controller-manager is an exception which is pulled from quay.io, thus not validated for pull path registry.redhat.io


CFSNM (approved):

asanzgom (commented):
@mattmahoneyrh correct me if I'm wrong but I think this Test Case should also be running for ODH deployments to check depending on the product, the correct deployment image. Isn't it?

mattmahoneyrh:
@asanzgom Yes, the check will run for both ODH and RHOAI (CPaaS builds).

asanzgom (approved):

PR #1837, opened 2024-09-20
Update button xpath to work with latest ODH

lugi0:


jstourac (dismissed):

CFSNM (dismissed):

jstourac:
Optionally we could put some comment that the latter value should be required once the odh code is pulled downstream.

lugi0:
@jstourac done

PR #1845, opened 2024-09-22
Set DSCInitialization timeout to 3 minutes if TEST_ENV is CRC

manosnoam:
- Deployment of ODH on Openshift Local (CRC) takes longer than deployment on other self-managed clusters.
It was required to increase the waiting time for DSCInitialization to 3 minutes (it is 30 seconds by default).

- Also refactor the Keyword `Wait For DSCInitialization CustomResource To Be Ready` 
to use `oc wait` instead of a `FOR` loop.

image


CFSNM (approved):
LGTM!

FedeAlonso (approved):

PR #1857, opened 2024-09-25
Fix:  xpath=//input[@data-testid="resource-manage-project-modal-name"] failure

ConorOM1:
Failures discovered in ODH Smoke & Sanity

Refactored to identify the element by data-testid only.

Tests fixed:
ODS-1784
ODS-1786
ODS-1775

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
@@ -8,7 +8,7 @@ Resource       ./Workbenches.resource
 ${DS_PROJECT_TITLE}=    Data Science Projects
 ${TITLE_INPUT_XP}=    xpath=//input[@id="manage-project-modal-name"]
 ${DESCR_INPUT_XP}=    xpath=//textarea[@id="manage-project-modal-description"]
-${RESOURCE_INPUT_XP}=    css:[data-testid="resource-manage-project-modal-name"]
+${RESOURCE_INPUT_XP}=    css:[data-testid="manage-project-modal-name"]
 ${GENERIC_CREATE_BTN_XP}=     xpath=//button[text()="Create"]
 ${GENERIC_CANCEL_BTN_XP}=     xpath=//button[text()="Cancel"]
 # TODO: Update to latter option once the change is pulled from ODH into downstream!
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
@@ -8,7 +8,7 @@ Resource       ./Workbenches.resource
 ${DS_PROJECT_TITLE}=    Data Science Projects
 ${TITLE_INPUT_XP}=    xpath=//input[@id="manage-project-modal-name"]
 ${DESCR_INPUT_XP}=    xpath=//textarea[@id="manage-project-modal-description"]
-${RESOURCE_INPUT_XP}=    xpath=//input[@data-testid="resource-manage-project-modal-name"]
+${RESOURCE_INPUT_XP}=    css:data-testid="resource-manage-project-modal-name"
 ${GENERIC_CREATE_BTN_XP}=     xpath=//button[text()="Create"]
 ${GENERIC_CANCEL_BTN_XP}=     xpath=//button[text()="Cancel"]
 # TODO: Update to latter option once the change is pulled from ODH into downstream!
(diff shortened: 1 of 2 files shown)


jiridanek (dismissed):

ods_ci/tests/Tests/0400__ods_dashboard/0410__ods_dashboard_projects/0410__ods_dashboard_projects.robot
```diff
@@ -15,9 +15,9 @@ Test Tags          Dashboard
 
 
 *** Variables ***
-${PRJ_TITLE}=   ODS-CI Common Prj
+${PRJ_TITLE}=   ods-ci-common-project
```
jstourac, line 18: Is this change intentional as part of this fix? It now matches what the `PRJ_RESOURCE_NAME` contains.

jiridanek: Good point, looks I'm not suspecting enough

ConorOM1: yes intentional. was causing the following setup failure `ODS-CI Common Prj != ods-ci-common-project` as the project title used is now ods-ci-common-project

jstourac: I don't get it :slightly_smiling_face: This variable is to define the project title, no? So, what is causing it to be ods-ci-common-project? Isn't there some conflict with what is defined by the other variable I mentioned?

ConorOM1: hmmm yes good point. it looks like somewhere along the line ${PRJ_TITLE} is being changed ? Or not being checked correctly ?

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
```diff
@@ -8,7 +8,7 @@ Resource       ./Workbenches.resource
 ${DS_PROJECT_TITLE}=    Data Science Projects
 ${TITLE_INPUT_XP}=    xpath=//input[@id="manage-project-modal-name"]
 ${DESCR_INPUT_XP}=    xpath=//textarea[@id="manage-project-modal-description"]
-${RESOURCE_INPUT_XP}=    xpath=//input[@data-testid="resource-manage-project-modal-name"]
+${RESOURCE_INPUT_XP}=    css:[data-testid="manage-project-modal-name"]
```
jstourac, line 11:
```suggestion
${RESOURCE_INPUT_XP}=    css:[data-testid="resource-manage-project-modal-name"]
```

ods_ci/tests/Tests/0400__ods_dashboard/0410__ods_dashboard_projects/0410__ods_dashboard_projects.robot
jstourac: I commented a suggestion above what I think is the rootcause.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
jstourac: Also, are you sure that this change works fine? Changing from xpath to css (AFAIK equivalent for the `document.querySelector()` in the browser console) - will this work with all the current places this variable is being used now? :thinking:

jstourac (changes requested):
Commented above; I'm not sure about the current state of these changes.

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
ConorOM1: this is actually the root cause of previous comment. was pointing to the wrong selector, need to clean up

re css: this is the best way going forward according to @manosnoam , by looking for the specific data-testid only as it should be unique

jstourac: I understand the motivation; what I'm not sure is the `css:` prefix there - will it work just fine with the current places this is being used? Shouldn't we have `//` prefix there?

Disclaimer: I didn't test it and I know basically nothing about this and html and stuff around it.

jiridanek: > I understand the motivation; what I'm not sure is the css: prefix there - will it work just fine with the current places this is being used? Shouldn't we have //* prefix there?

This is just SeleniumLibrary syntax for XPaths and css selectors; not really web-stuff related https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#:~:text=can%20be%20implicit.-,Default%20locator%20strategy,-By%20default%2C%20locators

ConorOM1: thanks for your help @jstourac & @jiridanek

jstourac (approved):

manosnoam (approved):

manosnoam:
@ConorOM1 Thanks for the fix. 
BTW, Do you know how will it work without the square brackets around the CSS value ?

ConorOM1:
> @ConorOM1 Thanks for the fix! BTW, Do you know how will it work without the square brackets around the CSS value ?

@manosnoam TY ! without brackets throws an invalid selector exception

jiridanek (approved):

PR #1859, opened 2024-09-25
Introduce support for the Jupyter/Elyra 4.

jstourac:
These changes are required to support the Jupyter/Elyra 4.x IDE version
bump which is expected in the OOTB Workbench images 2024.2+. All changes
are done so that the backward compatibility with the current
Jupyter/Elyra 3.x version is preserved.

Note: since the 'default' version of the workbench images still points
to 2024a with Jupyter 3, the line with the global env variable update to
work with Jupyter 4 is commented out now with a TODO comment.

Also, the reason why we are adding some code from the JupyterLibrary is
that we're trying to use SeleniumLibrary for the browser opening. This
doesn't match with some part of the code of the JupyterLibrary where it
is expected that some keywords that are present in both (SeleniumLibrary
and JupyterLibrary) are used from the JupyterLibrary. Since to narrow
down this discrepancy (using JupyterLibrary explicitly and completely
for the JupyterHub testing) would require more changes, I did it via
this ughly hack now having part of the source from the JupyterLibrary
directly in our codebase.
Truth is that JupyterLibrary project seems to be abandonded so we may
also think about fork it and handle it by ourselves in case we'll need
more changes there.

---

Tested locally so far, Jenkins env is WIP but feel free to overview if you are interested.

CI:
1. `/job/devops/job/rhoai-test-flow/501` - run against 2024a - only 1 failed test due to the expected Jupyter 4 - but the image isn't there yet, updated and that one test re-run in 502 :white_check_mark: other failed tests are probably unrelated to changes in this PR
2. Sanity on the older image - `/job/devops/job/rhoai-test-flow/503/` :white_check_mark: no regression by these changes seem to be introduced


jiridanek (approved):

ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
```diff
@@ -354,6 +354,12 @@ Spawn Notebook With Arguments  # robocop: disable
             Run Keyword And Warn On Failure   Login To Openshift  ${username}  ${password}  ${auth_type}
             ${authorization_required} =  Is Service Account Authorization Required
             IF  ${authorization_required}  Authorize jupyterhub service account
+
+            # For Jupyter 4, we need to update global default variable values (images 2024b and newer)
+            # This calls method from JupyterLibrary Version.resource module
+            # TODO - shall be uncommented once the 2024b images will land into the product
+            # IF  "${version}"=="default"  Update Globals For JupyterLab 4
```
jgarciao, line 361: If the workbench pod is already running and you want to know the  jupyterlab version you could use keyword `Run Command In Container` and run this command `oc exec -n ${namespace}  ${pod_name} -- jupyter --version | grep jupyterlab | awk '{print $3}'`

jgarciao (approved):

ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterHubSpawner.robot
jstourac: Thank you Jorge. That is correct. I was a bit lazy to do it that way. I presume that having the check for `default` value should be sufficient for now and eventually won't be needed once we move further with another upgrade in 6 months. :slightly_smiling_face:

PR #1871, opened 2024-09-30
Fix check for nvidia-device-plugin-daemonset when deploying NVIDIA operator stack

bdattoma:
As of now, when deploying the NVIDIA GPU operator stack, the nvidia-device-plugin-daemonset gets restarted after being in "init" status. Our script has already started waiting for the pod to be in Ready status but it waits for a pod which won't never be up and running. In addition, it has a 20 minutes timeout which makes CI spending more time than needed

Solution:
- instead of waiting 20 minutes after fetching the pod name, it waits only 10seconds and if the pod is not ready, it continues the loop (i.e., fetching the pod name again, wait 10 s, etc)
- wait for the GPU node to be ready before deploying the NVIDIA Stack, it reduces the risk of hitting the timeouts

PR validation:
1. provision smaller GPU on IBM cloud: `rhods-ci-pr-test/3368` PASS - the job failure at the end is not related to this PR. It took 15 minutes from provisioning to stack deployment
2. provision bigger GPU on GCP: `rhods-ci-pr-test/3371` - approx 15 min for e2e flow (provisioning + operator installtion)
4. test provisioning without this PR and compare timing: `rhods-ci-pr-test/3372` PASS - it took 28 minutes for e2e flow (provisioning + operator installtion) - the job failure at the end is not related to this PR

### DIFF
--- a/ods_ci/tasks/Resources/Provisioning/GPU/provision-gpu.sh
+++ b/ods_ci/tasks/Resources/Provisioning/GPU/provision-gpu.sh
@@ -40,11 +40,9 @@ sed -i'' -e "s/INSTANCE_TYPE/$INSTANCE_TYPE/g" $PROVIDER_OVERLAY_DIR/gpu.yaml
 oc apply --kustomize $PROVIDER_OVERLAY_DIR
 # Add GPU label to the new machine-set
 oc patch machinesets -n openshift-machine-api "$NEW_MACHINESET_NAME" -p '{"metadata":{"labels":{"gpu-machineset":"true"}}}' --type=merge
-oc wait --timeout=$MACHINE_WAIT_TIMEOUT --for jsonpath='{.status.phase}'=Running machine $NEW_MACHINESET_NAME -n openshift-machine-api
-if [ $? -ne 0 ]; then
-  echo ">> Machine $NEW_MACHINESET_NAME is not in Running state after timeout $MACHINE_WAIT_TIMEOUT. Check the cluster please"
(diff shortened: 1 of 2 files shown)


ods_ci/tasks/Resources/Provisioning/GPU/provision-gpu.sh
```diff
@@ -39,3 +40,10 @@ sed -i'' -e "s/INSTANCE_TYPE/$INSTANCE_TYPE/g" $PROVIDER_OVERLAY_DIR/gpu.yaml
 oc apply --kustomize $PROVIDER_OVERLAY_DIR
 # Add GPU label to the new machine-set
 oc patch machinesets -n openshift-machine-api "$NEW_MACHINESET_NAME" -p '{"metadata":{"labels":{"gpu-machineset":"true"}}}' --type=merge
+# wait for the machine to be Ready
+echo "Waiting for GPU Node to be Ready"
+oc wait --timeout=$MACHINE_WAIT_TIMEOUT --for jsonpath='{.status.readyReplicas}'=1 machineset $NEW_MACHINESET_NAME -n openshift-machine-api
+ if [ $? -ne 0 ]; then
+  echo "Machine Set $NEW_MACHINESET_NAME does not have its Machines in Running status after $MACHINE_WAIT_TIMEOUT timeout"
+  echo "Please check the cluster"
```
kobihk, line 48: maybe in this case we want to add `exit 1` to raise this problem
```suggestion
  echo "Please check the cluster"
  exit 1
```

bdattoma: done

kobihk (approved):

tarukumar (approved):

PR #1867, opened 2024-09-29
New KW `Wait For Deployment Replica To Be Ready` for oc_install.robot

manosnoam:
This is to verify if the number of running pods is as desired by replica-set, instead of directly failing if the pods number is not as defined in ODS-CI (hard-coded).

For example, 3 pods are currently expected for odh-model-controller in ODS-CI, but in the latest ODH build the replica-set desired number is 1, not 3. This should fix this false failure.

PR #1872, opened 2024-10-01
Revert "New KW `Wait For Deployment Replica To Be Ready` for oc_install.robot" because discussion was not done 

tarukumar:
Reverts red-hat-data-services/ods-ci#1867


tarukumar:
why you guys will merge the pr when discussion was still on

tarukumar:
what if i go ahead and get two approval and merge the pr. Kindly follow the process and do not decide on the fly until all the discussion is done

jstourac (approved):
Approving this in case we decided we want to revert this before the true and proper tests are implemented elsewhere.

It would be also nice to enrich this with the extra commit amending the number of expected pods for the `odh-model-controller` from 3 to 1 on those two places where it is checked here. But can be also as a separate PR.

PR #1873, opened 2024-10-01
enhancement: include model registry namespace as part of the uninstalling

CFSNM:
there is a new namespace created downstream for the model registry named rhoai-model-registries. Deleting it as part of the uninstall process

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
@@ -110,7 +110,10 @@ Uninstall RHODS V2
     ${return_code}    ${output}    Run And Return Rc And Output
     ...    oc delete operatorgroup --all -n ${OPERATOR_NAMESPACE} --ignore-not-found
     Should Be Equal As Integers  ${return_code}   0   msg=Error deleting operatorgroup
-    Log To Console    message=Deleting Operator and it's associate namepsace
+    Log To Console    message=Deleting model registry associated namespace ${MODEL_REGISTRY_NAMESPACE}
+    ${return_code}    ${output}    Run And Return Rc And Output    oc delete ns ${MODEL_REGISTRY_NAMESPACE} --ignore-not-found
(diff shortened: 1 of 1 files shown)


aloganat (approved):

ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
```diff
@@ -110,7 +110,10 @@ Uninstall RHODS V2
     ${return_code}    ${output}    Run And Return Rc And Output
     ...    oc delete operatorgroup --all -n ${OPERATOR_NAMESPACE} --ignore-not-found
     Should Be Equal As Integers  ${return_code}   0   msg=Error deleting operatorgroup
-    Log To Console    message=Deleting Operator and it's associate namepsace
+    Log To Console    message=Deleting model registry associated namespace ${MODEL_REGISTRY_NAMESPACE}
+    ${return_code}    ${output}    Run And Return Rc And Output    oc delete ns ${MODEL_REGISTRY_NAMESPACE} --ignore-not-found
```
jstourac, line 114: Why is this needed here and the other namespaces don't have to be deleted separately here?

CFSNM: the model registry namespace does not include the label opendatahub.io/generated-namespace so its not deleted by the command in line 117

jstourac: And is it by design? :thinking:

CFSNM: let me ask the platform devs

CFSNM: @jstourac I am going to close this PR and revert the other one. This namespace is not intended to be deleted when uninstalling the operator, because there can be some model registry instances that need to be kept even if we delete the operator

PR #1878, opened 2024-10-01
Removing hard coded auth_provider

tonyxrmdavidson:
Removes the hardcoded auth_provider value as it is no longer needed.


nehachopra27 (approved):
LGTM

lugi0 (approved):

PR #1888, opened 2024-10-03
Patch kustomize for GPUs to provision in regions without zones on Azure

bdattoma:
In order to support provisioning of GPUs in some regions of Azure where availability zone are not present we need to skip removal of "zone" key

Notes:
- it is forcing the replicas to 1 because in case there are no zones, the machine set will get created with n > 1 nodes
- atm among the regions we use, `west us` is the only one without zones

PR validation (on local CI):
1. provision in westus region: PASS
2. regression: provision in eastus2 region: PASS


jstourac (dismissed):
/lgtm

Don't have thorough knowledge here, but seems good.

Don't like those missing new-lines warnings, though :grin:

lugi0 (approved):

jstourac (approved):

PR #1901, opened 2024-10-08
Fix; 11846 user can start stop launch delete workbench

FedeAlonso:
During the Pre RC efforts for 2.14 we found that this test was failing due to some wrong xpaths.
Here's the fix
image


asanzgom (approved):

jiridanek (approved):

ods_ci/tests/Resources/Page/ODH/JupyterHub/JupyterLabLauncher.robot
```diff
@@ -357,15 +357,13 @@ Add And Run JupyterLab Code Cell 5 In Active Notebook
...
-    ${add icon} =    Get JupyterLab Icon XPath Custom    add
-
     ${nb} =    Get WebElement    xpath://div${JLAB XP NB FRAG}\[${n}]
     ${nbid} =    Get Element Attribute    ${nb}    id
 
     ${active-nb-tab} =    Get WebElement    xpath:${JL_TABBAR_SELECTED_XPATH}
     ${tab-id} =    Get Element Attribute    ${active-nb-tab}    id
 
-    Click Element    xpath://div[@aria-labelledby="${tab-id}"]/div[1]//${add icon}
+    Click Element    xpath://div[@aria-labelledby="${tab-id}"]//div[@data-jp-item-name="insert"]
```
lugi0, line 366: @jstourac doesn't this overlap with your fix?

jstourac: Hm, yeah, very well spotted, Luca. These two changes in this file should be reverted as we can use the approach before with the update of the global values for the Jupyter based on the version in use.

jstourac: @FedeAlonso do you mind reverting it just to be sure I'm not missing anything that could break on your side?

jstourac: ... well, I went on and raised #1905 then.

PR #1903, opened 2024-10-09
fix: ODS-1813 wrong span

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-14144
During 2.14 nightly execution this test was failing due to a modification of an xpath.
Test results (using nightly cluster):
image


jiridanek (approved):

jstourac (approved):
I see there is no testid assigned to any of these elements, maybe it's worth to ask dashboard guys to add some.

PR #1913, opened 2024-10-10
Fix user permission tests

jstourac:
1. [Fix] the Jupyter app user custom permission settings
   * fixes the failing test caued by recent changes in the ODHDashboard
  resource which we use for the dashboard page loading
   * minor refactoring of the code and some comments added for easier
  understandability of the test steps
2. [Fix] improve the way we wait for the particular Dashboard page to load
For some dashboard pages, it takes time to load properly and they can
change the content during that time. E.g. there is a "Loading" text with
an animated gif provided until the actual page is present.
In case of accessing the Applications -> Enabled -> Jupyter app tile,
there may be also present a page title with "Loading..." text for a
brief amount of time. Such thing breaks our detection technique as then
our expected page title text doesn't match with what was present on the
page and fails with:
   ```
   BuiltIn.Should Be Equal:
   Loading... != Start a notebook server
   ```
   Adding a couple of retries should help to avoid such situation.

---

CI:
image

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
@@ -103,7 +103,8 @@ Wait For RHODS Dashboard To Load
     IF    "${expected_page}" == "${NONE}"
         Wait Until Page Contains Element    //div[@data-testid="home-page"]    timeout=${timeout}
     ELSE
-        Wait For Dashboard Page Title    ${expected_page}    timeout=${timeout}
+        Wait Until Keyword Succeeds    3x    5s
+        ...    Wait For Dashboard Page Title    ${expected_page}    timeout=${timeout}
     END
     IF    ${wait_for_cards} == ${TRUE}
         Wait Until Keyword Succeeds    3 times   5 seconds    Wait Until Cards Are Loaded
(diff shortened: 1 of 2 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
```diff
@@ -103,7 +103,8 @@ Wait For RHODS Dashboard To Load
     IF    "${expected_page}" == "${NONE}"
         Wait Until Page Contains Element    //div[@data-testid="home-page"]    timeout=${timeout}
     ELSE
-        Wait For Dashboard Page Title    ${expected_page}    timeout=${timeout}
+        Wait Until Keyword Succeeds    3x    5s
+        ...    Wait For Dashboard Page Title    ${expected_page}    timeout=${timeout}
```
lugi0, line 107: I'm not sure this is going to help for the problem you're describing. If the issue is that the page takes longer than expect to get from "loading" state to "loaded" state, we either need to improve the checks in the `Wait...` keyword or increase the timeout. 
Adding retries here is likely to result in multiple failed checks for the same reason, no?

lugi0 (approved):
one comment but overall lgtm

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
lugi0: Nevermind, the keyword does not reload the page in between retries, this should be good enough

jiridanek (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDashboard.robot
jstourac: Yeah, there is no reload of the page. It's just restart of the check that expects some text value in the xpath element. This seemed like the easiest solution for this problem to me.

PR #1920, opened 2024-10-10
[Fix] the tests for custom image import

jstourac:
https://issues.redhat.com/browse/RHOAIENG-1192 has been resolved, so we can update that test accordingly.

Plust this fixes another one test in the custom image import and fixes some common keywords from the ODHDashboard.robot file.

---

CI:

image

### DIFF
--- a/ods_ci/tests/Tests/0500__ide/0501__ide_jupyterhub/custom-image.robot
+++ b/ods_ci/tests/Tests/0500__ide/0501__ide_jupyterhub/custom-image.robot
@@ -18,7 +18,7 @@ Test Tags        JupyterHub
 *** Variables ***
 ${YAML} =         tests/Resources/Files/custom_image.yaml
 ${IMG_NAME} =            custom-test-image
-${IMG_URL} =             quay.io/opendatahub-contrib/workbench-images:jupyter-datascience-c9s-py311_2023c_latest
+${IMG_URL} =             quay.io/opendatahub/workbench-images:jupyter-minimal-ubi9-python-3.11-2024b-20241004
 ${IMG_DESCRIPTION} =     Testing Only This image is only for illustration purposes, and comes with no support. Do not use.
 &{IMG_SOFTWARE} =        Software1=x.y.z
 &{IMG_PACKAGES} =        elyra=2.2.4    foo-pkg=a.b.c
@@ -62,7 +62,7 @@ Verify Custom Image Can Be Added
 
 Test Duplicate Image
     [Documentation]  Test adding two images with the same name (should fail)
-    ...       ProductBug - https://issues.redhat.com/browse/RHOAIENG-1192
+    ...       There was a bug related https://issues.redhat.com/browse/RHOAIENG-1192
     [Tags]    Tier1    ExcludeOnDisconnected
     ...       ODS-1368
     Sleep  1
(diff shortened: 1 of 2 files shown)


FedeAlonso (approved):

lugi0 (approved):
I think you could set some of the xpath variables as globals and reuse them across keywords, but this works too

ods_ci/tests/Tests/0500__ide/0501__ide_jupyterhub/custom-image.robot
```diff
@@ -94,9 +93,11 @@ Test Bad Image URL
 
 Test Image From Local registry
     [Documentation]  Try creating a custom image using a local registry URL (i.e. OOTB image)
-    ...       Issue reported for this test in the past - https://github.com/opendatahub-io/odh-dashboard/issues/2185
     [Tags]    Tier1
     ...       ODS-2470
+    ...       ExcludeOnDisconnected    # Since we don't have internal image registry enabled there usually
+    ...       InternalImageRegistry    # Requires internal image registry enabled
+    ...       ProductBug               # https://issues.redhat.com/browse/RHOAIENG-1193
```
jgarciao, line 100: If the test has `ProductBug`, it will be run only in the known-issues quality gate. Beware RHOAIENG-1193 is closed as "Won't Do"

jgarciao (approved):

jiridanek (approved):

CFSNM (approved):

ods_ci/tests/Tests/0500__ide/0501__ide_jupyterhub/custom-image.robot
jstourac: Yeah, TBH, I was wondering whether we shouldn't remove the test at all... will consider it and maybe raise a separate issue.

ods_ci/tests/Tests/0500__ide/0501__ide_jupyterhub/custom-image.robot
```diff
@@ -94,9 +93,11 @@ Test Bad Image URL
 
 Test Image From Local registry
     [Documentation]  Try creating a custom image using a local registry URL (i.e. OOTB image)
-    ...       Issue reported for this test in the past - https://github.com/opendatahub-io/odh-dashboard/issues/2185
     [Tags]    Tier1
     ...       ODS-2470
+    ...       ExcludeOnDisconnected    # Since we don't have internal image registry enabled there usually
```
apodhrad, line 98: Hi @jstourac this is the matter of an option in our automation where the internal registry is disabled. Moreover, the image registry can be enabled/disabled outside of disconnected automation (using `oc` command).

jstourac:
> I think you could set some of the xpath variables as globals and reuse them across keywords, but this works too

You are right, Luca - I was considering it for the `custom_image_name_record` but it needs the image name to be resolved properly. At the same time - shortening it so it doesn't contain the image name looked a bit ugly for me. Thus I left it defined in each function... maybe we can restructure it in the future

ods_ci/tests/Tests/0500__ide/0501__ide_jupyterhub/custom-image.robot
jstourac: I know, Andrej. But the test isn't passing now anyway. Also, since the linked issue won't be fixed probably, the test will probably go away at all.

PR #1923, opened 2024-10-13
[Fix] the test for the package versions on images

jstourac:
In the second part of this test, we have a selected list of packages so that we expect these packages to be same version amongst all the tested images. We have to reduce this list as now there are more packages that are expected not to be in the same version as was in past.

---

CI:

image

The failed tests are expected and will be fixed once this fix opendatahub-io/notebooks#742 is propagated into the product.


jstourac:
Just rebased so that we can see the GHA are truly green.

jiridanek (approved):

CFSNM (approved):

PR #1925, opened 2024-10-14
fix: skip disruptive operator tests in disconnected environments

CFSNM:
The Operator disruptive tests reinstall the dependant operator and check some conditions. As per a conversation with @apodhrad, these tests can cause flaky results as per the environment nature. Adding the tag to skip them


asanzgom (approved):

bdattoma (approved):

bdattoma:
@CFSNM @asanzgom @apodhrad Instead of skipping, could you run them separately from the rest of the Disconnected suites? So we don't lose coverage.
Also, would it be possible to make these tests more stable on disconnected as well?

CFSNM:
@bdattoma the main issue with disconnected environment is that the approval plan is always setted to manual, so the dependant operator reinstall is stuck due to the need to wait for the install plan to be approved.

I just wanted to disable them in the disconnected environments until we find a proper way to handle them (adding an additional step only in disconnected to approve the install plan, but not sure how feasible this change is)

CFSNM:
@bdattoma @asanzgom raised https://issues.redhat.com/browse/RHOAIENG-14362 to track this development.

PR #1934, opened 2024-10-16
fix for single serving test

ConorOM1:
This is to add UI fixes for ODS-2626 & ODS-2552. Previous test updates were for multi-model scenario, the original UI bugs were in the single model server test.

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
@@ -19,13 +19,14 @@ ${S3_REGION_DC_INPUT_XP}=          xpath=//input[@aria-label="Field list AWS_DEF
 ${S3_BUCKET_DC_INPUT_XP}=     xpath=//input[@aria-label="Field list AWS_S3_BUCKET"]
 ${S3_DEFAULT_BUCKET}=    ods-ci-s3
 ${MS_TABLE_PROJECT}=    /../../td[@data-label="Project"]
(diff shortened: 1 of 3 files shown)


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHModelServing.resource
```diff
@@ -19,13 +19,14 @@ ${S3_REGION_DC_INPUT_XP}=          xpath=//input[@aria-label="Field list AWS_DEF
 ${S3_BUCKET_DC_INPUT_XP}=     xpath=//input[@aria-label="Field list AWS_S3_BUCKET"]
 ${S3_DEFAULT_BUCKET}=    ods-ci-s3
 ${MS_TABLE_PROJECT}=    /../../td[@data-label="Project"]
-${MS_TABLE_ENDPOINT_INPUT}=    /ancestor::tr//td[@data-label="Inference endpoint"]//button[@data-testid='internal-service-button']
+${MS_TABLE_ENDPOINT_INPUT}=    /ancestor::tr//td[@data-label="Inference endpoint"]//button[@id='copy-button-4']
```
jstourac, line 22: Hm, interesting - truly moving away from the `data-testid` to a simple `id`? Was it removed from the element so we can't use it anymore?

ConorOM1: yeah there is only a data-testid for the internal endpoint option at the moment. If the external route option is enabled there is no data-testid

lugi0 (approved):

FedeAlonso (approved):

CFSNM (approved):

PR #1935, opened 2024-10-16
Increase timeout for Iris Pipeline

diegolovison:
* Increased the timeout because it is taking more time to run on airgap cluster
* `Delete Project` is under the `[Teardown]`

Tested on ods-dis-psi-04


jstourac:
Also, please set the labels for this one.

diegolovison:
> Do we know the reason why it takes longer to disconnect?

No. It just take times to finish :)

apodhrad (approved):

apodhrad:
Hi @jstourac there might be several reasons
* since iris pipeline downloads python deps, these deps are first mirrored via devpi on bastion and then sent to the python execution
* I have observed that pulling some big images from image registry on bastion is slower than in online cluster from quay.io/registry.redhat.com

I do not have any stats / metrics. It is just my random observation.

PR #1955, opened 2024-10-22
Remove generation validation from Pipelines Test

diegolovison:
Remove generation validation from Pipelines Test

That step is not needed

Tested with: rhoai-test-flow/967


CFSNM (approved):

ods_ci/tests/Resources/CLI/DataSciencePipelines/DataSciencePipelinesBackend.resource
```diff
@@ -63,10 +63,6 @@ Create PipelineServer Using Custom DSPA
     ...    ${assert_install}=${TRUE}    ${configure_pip_index}=${TRUE}
```
bdattoma, line 63: maybe this argument is not needed anymore now? could you remove it in case?

diegolovison: https://github.com/red-hat-data-services/ods-ci/issues/1959

PR #1957, opened 2024-10-23
Adds a new tests to check the notebook controllers deployments

jstourac:
This creates a dedicated test for the notebook controller deployments, their replica sets and containers. Just a basic metadata are checked.

Should resolve the:
* https://issues.redhat.com/browse/RHOAIENG-13756
* https://issues.redhat.com/browse/RHOAIENG-13757

---

CI: job/devops/job/rhoai-test-flow/985/

image

Since I touched also the code that is used during the product installation on the cluster, I ran the job linked above together with the RHOAI 2.15 installation and it passed just fine.

Also, the execution above was run without the [latest change](https://github.com/red-hat-data-services/ods-ci/compare/0fcc02897ea78829399de19c0752f4b661b80a7a..f0b9279aa5eae0f7192b14c3f8192342c4050a62) - I checked locally that no `true` is printed in the log anymore.

### DIFF
--- a/ods_ci/tests/Resources/OCP.resource
+++ b/ods_ci/tests/Resources/OCP.resource
@@ -107,16 +107,19 @@ Get Events And Pod Logs
     RETURN    ${events}    ${podlogs}
 
 Wait For Pods To Be Ready
+    [Documentation]    Waits until the pods with given label selector in the given namespace have `condition=ready`.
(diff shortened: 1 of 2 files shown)


jiridanek (approved):

CFSNM (approved):
LGTM!

PR #1964, opened 2024-10-25
RHOAIENG-11849 Refactor is required for edit DSP form

ConorOM1:
This fixes an automation bug due to UI changes for Sanity test ODS-2112

Tested on: /devops/job/rhoai-test-flow/1018

image

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
+++ b/ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
@@ -456,8 +456,7 @@ Check Resource Name Should Be Immutable
     [Documentation]    Checks if the Resource Name is not editable
     [Arguments]     ${project_title}
     ODHDashboard.Click Action From Actions Menu    item_title=${project_title}    item_type=project   action=Edit
-    Wait Until Page Contains Element    ${RESOURCE_INPUT_XP}
-    Element Should Be Disabled    ${RESOURCE_INPUT_XP}
+    Page Should Not Contain Element    ${RESOURCE_INPUT_XP}
     Click Button    Cancel
 
 Update Data Science Project Name


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Projects.resource
```diff
@@ -456,8 +456,7 @@ Check Resource Name Should Be Immutable
     [Documentation]    Checks if the Resource Name is not editable
     [Arguments]     ${project_title}
     ODHDashboard.Click Action From Actions Menu    item_title=${project_title}    item_type=project   action=Edit
-    Wait Until Page Contains Element    ${RESOURCE_INPUT_XP}
-    Element Should Be Disabled    ${RESOURCE_INPUT_XP}
+    Page Should Not Contain Element    ${RESOURCE_INPUT_XP}
```
jstourac, line 459: I still see the `Resource name` title and an immutable field below it. So, shouldn't we update the xpath of the RESOURCE_INPUT_XP? Otherwise this check doesn't make sense. What am I missing?
image

ConorOM1: so the xpath for ${RESOURCE_INPUT_XP} is used for creating a project test, and the changes for ODS-2112 test is to check that its not there when editing a project. WYT ?

jstourac: Okay, got it. Since we were testing presence of the Element before, I would expect that apart from this Should Not Contain Element, we would also have another Should Contain Element with the xpath to the expected field that is there now.

But I'm not gonna block this.

jstourac (approved):

manosnoam (approved):

PR #1979, opened 2024-10-28
Add checks for RHOAIENG-15033 bug in post upgrade scenario

lugi0:
This change checks for the presence of bug RHOAIENG-15033 in the post upgrade scenario for Model Registry.
Since the bug has not yet been fixed, the test case will fail in 2.15 and likely in 2.16 as well

### DIFF
--- a/ods_ci/tests/Resources/Page/ModelRegistry/ModelRegistry.resource
+++ b/ods_ci/tests/Resources/Page/ModelRegistry/ModelRegistry.resource
@@ -372,6 +372,13 @@ Model Registry Post Upgrade Scenario
     Set Library Search Order    SeleniumLibrary
     RHOSi Setup
     Component Should Be Enabled    modelregistry
+    # Check for bug RHOAIENG-15033, will fail in 2.15
+    ${operator_rest}=    Get Model Registry Operator REST Image
+    ${operator_grpc}=    Get Model Registry Operator GRPC Image
+    ${instance_rest}=    Get Model Registry Instance rest Image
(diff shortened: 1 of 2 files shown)


lugi0:
Verified in rhoai-test-flow/1069

tonyxrmdavidson (approved):
I can see the pr-check run passed, so /lgtm for me.

ods_ci/tests/Resources/Page/ModelRegistry/ModelRegistry.resource
```diff
@@ -404,3 +411,37 @@ Post Upgrade Scenario Teardown
...
+    [Documentation]    Returns the name of the operator pod for model registry
+    ${rc}    ${name} =    Run And Return Rc And Output
+    ...    oc get pod -l component.opendatahub.io/name=model-registry-operator -n ${APPLICATIONS_NAMESPACE} -o yaml | yq '.items[0].metadata.name'
+    Should Be Equal As Integers    ${rc}    0
+    RETURN    ${name}
+
+Get Model Registry Operator ${type} Image
+    [Documentation]    Returns the GRPC image defined in the operator, type should be either GRPC or REST (case 
+    ...    sensitive)
+    IF    "${type}"!="GRPC" and "${type}"!="REST"
```
bdattoma, line 425: it should be an OR condition

maybe you need to lower case them in the check at least? the keyword calls use both upper and lowercase

bdattoma (approved):

PR #1983, opened 2024-10-28
This PR adds a sleep to allow the DSC model-registry flag setting to Managed to be implemented

tonyxrmdavidson:
This PR adds a sleep to allow the DSC model-registry flag setting to Managed to be implemented


lugi0 (approved):
Probably a bit excessive but we can find a better signal later

CFSNM:
Suggestion: move this keyword to a common resource file and use it: https://github.com/red-hat-data-services/ods-ci/blob/7c7efd904cfec39c5693d0f9eaed34a40fad6d81/ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0113__dsc_components.robot#L327

It switches the state of a component to managed and wait for it to be completed

bdattoma (approved):

PR #1986, opened 2024-10-29
Update codeflare-sdk tests to run against Python 3.11 and 3.9 version

Srihari1192:


Srihari1192:
Triggered the tests here  `job/rhoai-test-flow/1088/`

ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource
```diff
@@ -53,9 +54,9 @@ Prepare Codeflare-SDK Test Setup
```
ChughShilpa, line 54: `${result} =    Run Process  virtualenv -p python3.11 ${VIRTUAL_ENV_NAME}`
    It should also be updated with ${PYTHON_VERSION}

Srihari1192: Thanks for pointing this @ChughShilpa.. I think we created this earlier virtual env setup for 3.9 as by default venv 3.11 available in our jenkins.. will update it accordingly

Srihari1192: Removed virtual env creation steps as we can switch poetry env with existing virtual environment in jenkins , test successfully ran here `job/rhoai-test-flow/1109/`

ChughShilpa (approved):
/lgtm

sutaakar (approved):

jiripetrlik (approved):

PR #1987, opened 2024-10-29
Add workbench runtimes for RHOAI 2.15 & script to download them from running workbench

jgarciao:
We have some pipeline tests that use the workbench runtimes (e.g. the pytorch-romc runtime). 

In order to help finding the latest workbench images shipped for each version of RHOAI, we would like to add them here in ods-ci, with a script to obtain them semi-automatically.

In a follow-up PR we'll update the images used in our pipeline samples.

### DIFF
--- a/ods_ci/tests/Resources/Files/workbench-runtimes/update-runtimes.sh
+++ b/ods_ci/tests/Resources/Files/workbench-runtimes/update-runtimes.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+#######################################################################################
+# Follow these steps to update the workbench runtimes in this folder                  #
+#  - Open a browser and connect to RHOAI Dashboard                                    #
+#  - Create a data science project my-project                                         #
+#  - Start a Workbench using the Standard Data Science Image. Name should be wb-sds   #
+#  - Open a terminal:                                                                 #
+#    - oc login ...                                                                   #
+#    - ./update-runtimes.sh my-project                                                #
+#  - The workbench runtimes will be downloaded, overwritting existing ones            #
+#######################################################################################
+
+RUNTIME_IMAGES_PATH=/opt/app-root/share/jupyter/metadata/runtime-images/
+
+if [ $# -ne 1 ]; then
+    echo "Wrong number of parameters: missing project name" >&2
+    script_name=$(basename "$0")
+    echo "Usage: $script_name my-project" >&2
+    exit 2
+fi
+export MY_PROJECT=${1}
+
+if ! oc get project "$MY_PROJECT" > /dev/null 2>&1
+then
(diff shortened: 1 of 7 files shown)


ods_ci/tests/Resources/Files/workbench-runtimes/update-runtimes.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#######################################################################################
+# Follow these steps to update the workbench runtimes in this folder                  #
+#  - Open a browser and connect to RHOAI Dashboard                                    #
+#  - Create a data science project my-project                                         #
+#  - Start a Workbench using the Standard Data Science Image. Name should be wb-sds   #
```
diegolovison, line 6: Why the dashboard team can't have a tag that matches the RHOAI releases?

diegolovison: @jstourac what are your thoughts about ^ ?

jstourac: > Why the dashboard team can't have a tag that matches the RHOAI releases?
I presume it should be IDE team, not dashboard team.

I guess that the reason we don't use RHOAI version tags for the images is that we don't necessarily update our images with every single rhoai release - as such, these aren't exactly bounded to one and only release now but are being updated based on the changes being done in the repository.

jstourac:
I'm not sure whether it was intentional to get it from a RHOAI build and as such for running this images update means having a running OpenShift cluster, RHOAI build installed there and also running a workbench. This has advantage that it can be used for any nightly build for sure.

In case we are okay with only not too frequent updates (e.g. only in CR builds), we can download this information directly from the [notebooks repository here](https://github.com/red-hat-data-services/notebooks/tree/main/jupyter/datascience/ubi9-python-3.11/runtime-images).

ods_ci/tests/Resources/Files/workbench-runtimes/update-runtimes.sh
jgarciao: Thanks for the explanation. It makes sense

jgarciao:
Thanks @jstourac , I didn't know we could get the info from there. Can we assume the files on each branch (e.g. rhoai-2.15) contains the runtimes in that release? If yes, then the script is not needed

jstourac:
Yes, release branch is where you want to look into. Just beware - as of today, the updates to the runtime images are usually propagated to the release branch after the code freeze (as at this point the proper notebooks release is performed). This will change in the future and will be updated more frequently hopefully.

So, if the above isn't a concern, nudging the data from the repository is probably better than necessity to rely on a running cluster with an actual RHOAI installation on it.

PR #1991, opened 2024-10-30
Modify available OCP versions logic

omrirh:
Check if available OCP versions is neither None nor empty str.
The existing logic fails for old OCP versions (e.g. 4.12-latest), from a [failing job](https://opendatascience-jenkins-csb-rhods.apps.ocp-c1.prod.psi.redhat.com/job/devops/job/rhoai-test-flow/1118/console) with Versions status logging:
`Versions length before failure: 0 and its type is: <class 'str'>`

### DIFF
--- a/ods_ci/utils/scripts/ocm/ocm.py
+++ b/ods_ci/utils/scripts/ocm/ocm.py
@@ -147,7 +147,7 @@ class OpenshiftClusterManager:
 
                 version_cmd = f'ocm list versions {chan_grp} | grep -w "{re.escape(version)}*"'
                 versions = execute_command(version_cmd)
-                if versions is not None:
+                if versions:
                     version = [ver for ver in versions.split("\n") if ver][-1]
                 self.openshift_version = version
             else:


kobihk (dismissed):

ods_ci/utils/scripts/ocm/ocm.py
```diff
@@ -147,7 +147,7 @@ def osd_cluster_create(self):
 
                 version_cmd = f'ocm list versions {chan_grp} | grep -w "{re.escape(version)}*"'
                 versions = execute_command(version_cmd)
-                if versions is not None:
+                if versions:
```
jstourac, line 150: Hmm, shouldn't we `log.error()` and return `sys.exit(1)` now, since this case is properly handled? :thinking:

jstourac: Yeah, IIUC, this is about the case when the required version isn't offered by `ocm` command anymore. So we should log some clear error message and return non zero code, isn't so? What happened up to now before your fix? I suppose that the code failed non-clean so we saw this right error right-away. But now it will continue and the empty `version` is provided further on to the `ocm` command which will not make it easier to debug what the hell went wrong, IMO.

Unless I'm missing something. I don't know this part of the code.

omrirh: oh yeah you're right, we definitely should fail the job when there are no corresponding versions in OCM.
I added this logic now.
Thanks for catching that

jstourac: Thanks, LGTM now :)

jstourac (approved):

PR #1992, opened 2024-10-30
Re-organise the Model Registry Tests

tonyxrmdavidson:
This PR provides a fix for the **_Wait For Model Registry Containers To Be Ready_** keyword.

It also re-organises the Model Registry Tests by moving duplicated keywords into ModelRegistry.resource.


bdattoma:
> It also re-organises the Model Registry Tests by moving duplicated keywords into ModelRegistry.resource.

it is not moving actually, just deleting. Are the deleted keyword already in the destination path maybe?

bdattoma (approved):

PR #1996, opened 2024-10-30
Add wait for deployment ready

asanzgom:
Added check for deployment ready since execution together with other Test Cases that remove model mesh deployment can cause the recreation of deployment to not be ready at the moment of execution

### DIFF
--- a/ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0114__component_images.robot
+++ b/ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0114__component_images.robot
@@ -2,9 +2,24 @@
 Documentation       Test Suite to check for components
 
 Resource            ../../../../../../tasks/Resources/RHODS_OLM/install/oc_install.robot
-
+Resource            ../../../../../../tests/Resources/Page/Components/Components.resource
 Suite Setup         Suite Setup
 Suite Teardown      Suite Teardown
+
+
+*** Variables ***
+${APPLICATIONS_NS}                              ${APPLICATIONS_NAMESPACE}
+${KUEUE_LABEL_SELECTOR}                         app.kubernetes.io/name=kueue
+${CODEFLARE_LABEL_SELECTOR}                     app.kubernetes.io/name=codeflare-operator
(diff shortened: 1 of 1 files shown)


mattmahoneyrh (approved):

asanzgom:
Running rhods-ci-pr-test/3423/ together with all Operator Test Cases

ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0114__component_images.robot
```diff
@@ -2,28 +2,43 @@
...
+${MODELMESH_CONTROLLER_LABEL_SELECTOR}          app.kubernetes.io/instance=modelmesh-controller
+${ODH_MODEL_CONTROLLER_LABEL_SELECTOR}          app=odh-model-controller
+${MODELREGISTRY_CONTROLLER_LABEL_SELECTOR}      control-plane=model-registry-operator
+${KSERVE_CONTROLLER_MANAGER_LABEL_SELECTOR}     control-plane=kserve-controller-manager
+${TRUSTYAI_CONTROLLER_MANAGER_LABEL_SELECTOR}   app.kubernetes.io/part-of=trustyai
+${WORKBENCHES_LABEL_SELECTOR}                   app.kubernetes.io/part-of=workbenches
+
+
 *** Test Cases ***
 Check For Correct Component Images
```
jgarciao, line 26: As now we are waiting a few minutes for each component, could you add a [test timeout](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#timeouts) to prevent waiting too much?  

Example  (not sure what will be the appropriate timeout for this test)
`[Timeout]    15`

jgarciao (changes requested):

ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0114__component_images.robot
asanzgom: well, as we keep adding more components, it becomes more slowly and sometimes prone to errors, if we limit the Test Case to a fixed Timeout it is going to fail as a false positive IMO

MarianMacik (approved):
Looks good to me!

PR #1963, opened 2024-10-24
fix: add statement to approve the installplan when a new operator is being installed

CFSNM:
Related ticket: https://issues.redhat.com/browse/RHOAIENG-14362

Also enabling again the disruptive tests that were disabled in disconnected environments. This PR fixes that

Checks:

- Selfmanaged cluster OK
- Managed cluster OK
- Disconnected cluster OK

PR #1997, opened 2024-10-30
Added While Loop to wait until Kw succeeds to get installplan name

asanzgom:


asanzgom:
running rhoai-test-flow/1151/

bdattoma (dismissed):

ods_ci/tests/Resources/Page/Operators/ISVs.resource
```diff
@@ -37,7 +37,8 @@ Install ISV Operator From OperatorHub Via CLI    # robocop: disable
     Wait Until Keyword Succeeds    1 min    0 sec
     ...    Is Resource Present    Subscription    ${subscription_name}     ${namespace}    ${IS_PRESENT}
     Sleep    30s
-    ${installplan_name}=     Get Resource Attribute      ${namespace}
+    ${installplan_name}=       Wait Until Keyword Succeeds    1 min    0 sec
```
jstourac, line 40: Isn't 0 sec between the attempts too eager? Shouldn't we give some time between the attempts?

I can see that this is a second instance of such a thing in this file - 3 lines above it was added just recently too [in this change](https://github.com/red-hat-data-services/ods-ci/pull/1963/files#diff-f7e0124fd66571233d7c43584d02497c8e74d0b6e66d77d4876cc1cf0005a23fR37).

So, maybe there is some reason for it and we don't bother, but I truly don't understand why. Personally, I would use something like this instead:

```suggestion
    ${installplan_name}=       Wait Until Keyword Succeeds    1 min    5 sec
```
And same for the waiting for the Subscription thing on the line 37 :shrug: 

Anyway, this is just a proposal - also, I just saw that we have this 0s retry interval on multiple places in other files there too... sooo... whatever.... we're ods-ci is deprecated anyway, right? :rofl:

jstourac (dismissed):

ods_ci/tests/Resources/Page/Operators/ISVs.resource
```diff
@@ -37,7 +37,8 @@ Install ISV Operator From OperatorHub Via CLI    # robocop: disable
     Wait Until Keyword Succeeds    1 min    0 sec
     ...    Is Resource Present    Subscription    ${subscription_name}     ${namespace}    ${IS_PRESENT}
     Sleep    30s
```
CFSNM, line 39: if we are waiting for the keyword to be successful, we dont need this sleep line

CFSNM (changes requested):

ods_ci/tests/Resources/Page/Operators/ISVs.resource
CFSNM, line 40: @asanzgom I think we need to refactor the keyword in the PR.
Wait Until Keyword Succeeds expects no failures in the keyword to check, but when finding an empty name for a install plan, this is not an error indeed, so the keyword will assume success in the first iteration.

CFSNM (changes requested):

ods_ci/tests/Resources/Page/Operators/ISVs.resource
```diff
@@ -37,8 +38,10 @@ Install ISV Operator From OperatorHub Via CLI    # robocop: disable
...
-    ${installplan_name}=     Get Resource Attribute      ${namespace}
-    ...      Subscription       ${subscription_name}        .status.installPlanRef.name
+    WHILE   $installplan_name == $EMPTY
```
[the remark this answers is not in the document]
asanzgom: Added Keyword Timeout

ods_ci/tests/Resources/Page/Operators/ISVs.resource
asanzgom: @jstourac this approach did not work so I resolve thread

ods_ci/tests/Resources/Page/Operators/ISVs.resource
asanzgom: removed sentence

ods_ci/tests/Resources/Page/Operators/ISVs.resource
asanzgom: yes @CFSNM that is the final approach I took

CFSNM (approved):
LGTM. Lets test this PR in a fresh managed cluster, to see if the dependant operators are installed properly when enabling the option to install RHOAI

asanzgom:
rhoai-test-flow/1221/ succeded

CFSNM:
@asanzgom can u try it on a managed cluster? I have seen recurrent errors there. You need to run a job to create a managed cluster and installing the latest RHOAI nightlies. The dependant operators should be installed properly

bdattoma (approved):

PR #1998, opened 2024-10-31
Update to Model Registry Smoke Test to remove the need for the Sleep 90s keyword

tonyxrmdavidson:
This PR will remove the Sleep 90s keyword and add a check to ensure the model-registry-controller-manager is in a Running state after modelregistry is set to Managed.

A successful test run can be seen on the following jenkins pipeline
- job/rhods-ci-pr-test/3429/console


tonyxrmdavidson:
Successful pr-check  for all commits job/rhods-ci-pr-test/3431

bdattoma (approved):

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_smoke.robot
```diff
@@ -28,17 +33,20 @@ Deploy Model Registry
...
+    ...    control-plane=model-registry-operator
     Component Should Be Enabled    modelregistry
-    Apply Db Config Samples    namespace=${NAMESPACE_MODEL_REGISTRY}
+    Apply Db Config Samples    namespace=${NAMESPACE_MODEL_REGISTRY}    samples=${MODEL_REGISTRY_DB_SAMPLES}
 
 Registering A Model In The Registry
     [Documentation]    Registers a model in the model registry
     [Tags]    Smoke    MR1302    ModelRegistry
     Register A Model    ${URL}
+    Sleep    5s
```
jstourac, line 46: This sleep here looks a bit fragile.

jstourac (dismissed):

lugi0 (approved):

ods_ci/tests/Tests/1300__model_registry/1302_model_registry_smoke.robot
```diff
@@ -68,7 +78,6 @@ Remove Model Registry Non UI
 
 Run Curl Command And Verify Response
     [Documentation]    Runs a curl command to verify response from server
-    [Arguments]    ${URL}
```
jstourac, line 71: I'm not sure why this change is part of this PR meaning that we'll prefer the global variable instead of passing it here. I'm not gonna block as I can see that in this file there is quite chaotic approach - some of the keywords use argument for this, some use global value... this isn't a good approach from my point of view.

PR #2005, opened 2024-11-06
Exclude must gather test case on disconnected

asanzgom:
Exclude must gather test case on disconnected envs


MarianMacik (approved):
Looks good to me.

CFSNM (approved):

PR #2017, opened 2024-11-11
Fix: Dashboard ODS-2206 and ODS-1814

FedeAlonso:
On nightly /rhoai-test-flow/1324 
Tests:
- ODS-2206: Verify User Can Create, Run and Delete A DS Pipeline From DS Project Details Page Using Custom Pip Mirror
- ODS-1814:  Verify User Can Create And Start A Workbench With Existent PV Storage
Were failing due to different Dashboard related issues

This PR fixes it:
image


ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
```diff
@@ -85,8 +85,6 @@ Create Pipeline Run
     ...            ${end_time}=${NONE}    ${cron_expr}=${NONE}    ${press_cancel}=${FALSE}    &{model_param}
     Projects.Move To Tab    Pipelines
     Pipelines.Click Action From Pipeline Actions Menu    pipeline_name=${pipeline_name}    action=Create run
-    Wait For RHODS Dashboard To Load    expected_page=Create run
```
jiridanek, line 88: Removing a wait is suspicious, but if it works...

jiridanek (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
jstourac: Looks like the wait was actually moved from here to the `Fill In Run Creation Form` keyword, so it seems safe.

But I agree that it would deserve some brief explanation in the commit message or PR description at least.

jstourac (approved):

ods_ci/tests/Resources/Page/ODH/ODHDashboard/ODHDataScienceProject/Pipelines.resource
bdattoma: I guess it works because of the Wait added in the next keyword https://github.com/red-hat-data-services/ods-ci/pull/2017/files#diff-82e844de82ae4101df6d15b896b02ea96f801b4a27c3a102c4e120da077518d1R46

bdattoma: Jan was faster :)

jgarciao:
Thanks for fixing this, Fede!

PR #2024, opened 2024-11-13
Hotfix/validate service mesh control plane already created

asanzgom:
Fixed Validate Service Mesh Control Plane Already Created Test Case to iterate on while loop until expected error message is printed on logs with a 5 min timeout.


asanzgom:
rhoai-test-flow/1409 passed succesfully

mattmahoneyrh (approved):

jstourac (approved):

PR #2034, opened 2024-11-15
fix: new Connection Types feature issues

FedeAlonso:
Due to the new "Connection types" feature, nigthly tests were failing.
This PR fixes it. 
It has been tested using the nightly cluster and running ODS-1932 (In this test there is a creation and edition of a connection)
image


jiridanek (approved):

jiridanek (approved):

jiridanek (approved):

ConorOM1 (dismissed):

lugi0 (dismissed):

FedeAlonso:
cc @jgarciao  I've fixed the issue with the creation of the data connection in the setup of pipelines tests 
image

(The error is not related with data connections)

jgarciao (approved):
Thanks a lot, Fede!

lugi0:
I think this broke the ability to create a Data Connection in a Data Science Project? I've tried running against 2.14/2.15 and the latest nightly I could find for 2.16 and it always failed for me

PR #2036, opened 2024-11-15
Chore: Added sleep and timeout for while loop (#2029)

CFSNM:
* Added sleep and timeout for while loop

* Added If else structure for Gettting Resource Attribute Kw for when value comes empty


bdattoma (approved):

MarianMacik (approved):
LGTM.

PR #2039, opened 2024-11-15
Add negative test for dashboard RBAC on Model Registry

lugi0:


lugi0:
Verified in rhoai-test-flow/1567

ods_ci/tests/Tests/1300__model_registry/1301_model_registry_model_serving.robot
```diff
@@ -76,3 +71,14 @@ Verify Model Registry Integration With Secured-DB
...
+
+Verify Unallowed User Cannot See Model Registry From The Dashboard
+    [Documentation]    Negative path test for dashboard RBAC on the Model Registry. User not part of the group that is
+    ...    allowed to use a Model Registry instace should not be able to see it from the dashboard.
+    [Tags]    Smoke    MRMS1301    ModelRegistry
+    Depends On Test    Verify Model Registry Integration With Secured-DB
+    Launch Dashboard    ${TEST_USER_3.USERNAME}    ${TEST_USER_3.PASSWORD}    ${TEST_USER_3.AUTH_TYPE}
+    ...    ${ODH_DASHBOARD_URL}    ${BROWSER.NAME}    ${BROWSER.OPTIONS}
+    Open Model Registry Dashboard Page    allowed_user=${FALSE}
+    SeleniumLibrary.Page Should Contain    Request access to model registries
```
bdattoma, line 84: isn't this already checked here https://github.com/red-hat-data-services/ods-ci/pull/2039/files#diff-d120a1bd56f9d1fb5afc47239d24ae89b2ae6da94f25d270134f304a412bea40R313

lugi0: It is, but I wanted to surface it as an explicit step in this TC just for clarity when going through the logs

bdattoma (approved):

ods_ci/tests/Tests/1300__model_registry/1301_model_registry_model_serving.robot
lugi0: I think it makes more sense to have it defined this way because if the user is allowed we need to refresh the page in case the permissions have not been reconciled yet, and so the keyword will still need to be passed some variable to define if the user is allowed or not. 
Once you have that information in the keyword what makes the most sense IMHO is to have that logic defined in the keyword

tarukumar (approved):

PR #2057, opened 2024-11-19
Fix model serving issues in 2.15

FedeAlonso:
https://issues.redhat.com/browse/RHOAIENG-16046

Fix model serving issues found in the upgrade

image


manosnoam (commented):
Thanks Fede. Can you check it against ODH-Nightly or 2.16 RC ?
If it does not work for 2.16, maybe we should explicitly say that it is only for 2.15.

mwaykole (approved):

kobihk (approved):

PR #2065, opened 2024-11-22
Update fms-hf-tuning image to version v2.1.2

sutaakar:


ChughShilpa (approved):

ChughShilpa:
Thinking, we can also add a note to mention the release version of fms image so that anyone is aware which tag is used.

sutaakar:
hmm, that can be added for all the images

abhijeet-dhumal (approved):

PR #2068, opened 2024-11-23
Update the way how to remove environment variables for Jupyter ntb

jstourac:
This change slightly simplifies the way how the variables are removed from the Jupyter spawner page for the notebook - now it doesn't depend on particular variable values, but simply clicks until the remove button is present on the page.

Another change here is that this variable removal is now run also in case that the given set of variables to the `Spawn Notebook With Arguments` keyword is empty. This will ensure that any eventual leftovers from previous test are cleaned up properly not affecting this current test.

CI: tested on my cluster with empty and non-empty variables pre-set to the jupyter spawner page


jiridanek:
This will assure that any eventual leftovers from previous test are cleaned up properly not affecting this current test.

> What does assure mean?
> Assure is something you would do to a person or an animal. These beings can have doubts or insecurities you can remove by assuring them, by giving them assurance:

https://www.grammarly.com/blog/commonly-confused-words/assure-ensure-insure/

jiridanek (approved):

PR #2071, opened 2024-11-25
fix: use cleanup.sh script to uninstall in sm clusters

CFSNM:
As discussed with @bdattoma and @aloganat, we were not using olminstall scripts to uninstall ODH/RHOAI. This PR replaces the "Uninstall RHODS V2" Kw by the one for selfmanaged clusters, using cleanup script


aloganat (approved):

bdattoma:
I'm adding @lugi0 and @tarukumar among the reviewers because from the history of the file it seems they were the main maintainer of that uninstall keyword.
To me, the new changes look ok, however I'd also ask them to have a look in case we're missing something. Thanks!!

CFSNM:
thanks @bdattoma

asanzgom (approved):

bdattoma (approved):

PR #2074, opened 2024-11-26
Add keywords to skip tests on Self managed cluster

ChughShilpa:
Closes [RHOAIENG-16219](https://issues.redhat.com/browse/RHOAIENG-16219)

This PR 

- adds a keyword `Is Cluster Type Self-Managed` to get cluster type 
- adds a keyword `Skip If Cluster Type Is Self-Managed` and `Skip If Cluster Type Is Managed` to skip tests on self managed/managed cluster

### DIFF
--- a/ods_ci/tests/Resources/OCP.resource
+++ b/ods_ci/tests/Resources/OCP.resource
@@ -332,3 +332,18 @@ Wait For Namespace To Be Active
     Log    ${rc}
     Log    ${value}
     Should Be Equal As Integers    ${rc}    0
+
+Is Cluster Type Self-Managed
+    [Documentation]    Get the value of cluster type depending on the console URL domain
+    ...    Returns ${TRUE} if cluster type is Self-Managed
+    ...    Returns ${FALSE} if cluster type is Managed
+    ${matches}=    Get Regexp Matches    ${OCP_CONSOLE_URL}    rh-ods
+    ${size}=    Get Length    ${matches}
+    IF    ${size}>0
+        ${domain}=    Get From List    ${matches}    0
+        IF    "${domain}" == "rh-ods"
+            RETURN    ${TRUE}
+        END
+    ELSE
+        RETURN    ${FALSE}
+    END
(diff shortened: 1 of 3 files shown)


CFSNM (approved):

sutaakar (dismissed):

ods_ci/tests/Resources/OCP.resource
```diff
@@ -332,3 +332,18 @@ Wait For Namespace To Be Active
     Log    ${rc}
     Log    ${value}
     Should Be Equal As Integers    ${rc}    0
+
+Get Cluster Type By Domain
```
jgarciao, line 336: As this keyword is returning a boolean it should be called **Is Self Managed Cluster Domain** or similar to follow the current naming in this file

jgarciao (commented):
I don't think you do it in this PR but please don't remove the existing keyword `Skip If RHODS Is Self-Managed`. It is being used in some tests

ods_ci/tests/Resources/OCP.resource
ChughShilpa: Renamed to `Is Cluster Type Self-Managed` in resemblance to related keyword `Skip If Cluster Type Is Self-Managed`

ChughShilpa: `I don't think you do it in this PR but please don't remove the existing keyword Skip If RHODS Is Self-Managed. It is being used in some tests`
@jgarciao I didn't remove the `Skip If RHODS Is Self-Managed` keyword in this PR. Also, not planning to remove it.
I got it statement regarding this keyword led this interpretation. 
Modifying the statement in description now

sutaakar (approved):

CFSNM (approved):

jgarciao (approved):

ods_ci/tests/Resources/OCP.resource
jgarciao: No problem. Thanks a lot!

PR #2078, opened 2024-11-27
Remove Suite variables to avoid overriding values provided by tests in pre/post upgrade tests

lugi0:
I have noticed that the model registry upgrade tests were using the project name `model-serving-upgrade` instead of the expected `model-registry-project-e2e`.
This change fixes the problem.


bdattoma (approved):
maybe you need to backport this to releases/2.16.0 branch as well and check if this is needed in other supported releases branch? https://access.redhat.com/support/policy/updates/rhoai-sm/lifecycle

lugi0:
@bdattoma the test still passes as expected, it's just using a namespace I did not expect. I think we can avoid backporting unless we see actual failures caused by this.

bdattoma:
> @bdattoma the test still passes as expected, it's just using a namespace I did not expect. I think we can avoid backporting unless we see actual failures caused by this.

Ah ok! I assumed this was to fix something broken

jgarciao (approved):

PR #2079, opened 2024-11-27
Add tag to disruptive tests so that they can be exclution from Operator  suite runs

mattmahoneyrh:
There are Operator tests that are disruptive in nature, which at times will need to exclude from being run with the Operator suite. Thus, adding tag so that desired tests can be excluded from run.

### DIFF
--- a/ods_ci/tests/Tests/2001__disruptive_tests/2002__dsc_negative_dependant_operators_not_installed.robot
+++ b/ods_ci/tests/Tests/2001__disruptive_tests/2002__dsc_negative_dependant_operators_not_installed.robot
@@ -28,7 +28,7 @@ ${IS_NOT_PRESENT}                           1
 Validate DSC and DSCI Created With Errors When Service Mesh Operator Is Not Installed    #robocop:disable
     [Documentation]    The purpose of this Test Case is to validate that DSC and DSCI are created
     ...                without Service Mesh Operator installed, but with errors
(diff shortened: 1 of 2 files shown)


jgarciao:
We currently use the Tier3 tag to mark the disruptive tests. 

Could you consider this?
- Non disruptive operator tests: add tag Operator and Tier1 or Tier2 (depending on time required)
- Disruptive operator tests: add Operator and Tier3 tag. If they are so disruptive that they prevent other tests to run, then remove the Tier3 tag and add OperatorDisruptiveTest, as César proposed
 
There is more info about the tags usage at the [Style Guide](https://docs.google.com/document/d/11ZJOPI1uq-0Wl6a2V8fkAv_TQhfzp9t_IjXAheaJxmQ/edit?tab=t.0#heading=h.s819p3c5ud7p). If you end up using OperatorDisruptiveTest, could you add this tag to the guide? Thanks!

mattmahoneyrh:
Using the style guild by using Tier3 tag to exclude disruptive tests would work, which for this case to run the Operator suite while excluding destructive tests  would be something like "--include Operator --exclude Tier3".

@CFSNM wdyt?

CFSNM:
@mattmahoneyrh I am okay about using -e Tier3. We are adding this tag to all of the Operator disruptive tests

asanzgom:
I do agree on using Tier3 tag as well

asanzgom (approved):

PR #2089, opened 2024-11-29
Add image references to the Distributed Workload image digests

sutaakar:


sutaakar:
I think those images could be all moved to the `DistributedWorkloads.resource`, WDYT?

ChughShilpa:
> I think those images could be all moved to the `DistributedWorkloads.resource`, WDYT?

Yes we can move it

ChughShilpa (approved):
Thanks

Srihari1192 (approved):

abhijeet-dhumal (approved):

PR #2100, opened 2024-12-02
Increase Codeflare SDK tests timeout

jiripetrlik:


Srihari1192:
@jiripetrlik I already opened PR for this https://github.com/red-hat-data-services/ods-ci/pull/2093 . As Tier 1 tests execution should be less than 10 minutes, i have updated timeout to 9 minutes and pre-requisite setup may take less than minute

PR #2128, opened 2024-12-10
Add AutomationBug Tag to Verify RHODS User Groups

asanzgom:
This is a temporary tag as it fails on cascade meanwhile Verify RHODS Accept Multiple Admin Groups And CRD Gets Updates is being fixed


kobihk (approved):

FedeAlonso (approved):
Wow! Great job!

PR #2129, opened 2024-12-10
RHOAIENG-12192 - Extend DSP e2e tests

jiripetrlik:
Move pipeline tests related to Distributed Workloads to Workloads Orchestration folder and add a new test "Verify Ods Users Can Create And Run A Data Science Pipeline With Ray Job Using The kfp Python Package".

### DIFF
--- a/ods_ci/tests/Resources/Files/pipeline-samples/v2/cache-disabled/ray_integration.py
+++ b/ods_ci/tests/Resources/Files/pipeline-samples/v2/cache-disabled/ray_integration.py
@@ -6,11 +6,11 @@ common_base_image = (
 
 
 # image and the sdk has a fixed value because the version matters
-@dsl.component(packages_to_install=["codeflare-sdk==0.21.1"], base_image=common_base_image)
+@dsl.component(packages_to_install=["codeflare-sdk==v0.24.0"], base_image=common_base_image)
 def ray_fn() -> int:
     import ray  # noqa: PLC0415
     from codeflare_sdk import generate_cert  # noqa: PLC0415
-    from codeflare_sdk.cluster.cluster import Cluster, ClusterConfiguration  # noqa: PLC0415
+    from codeflare_sdk.ray.cluster import Cluster, ClusterConfiguration  # noqa: PLC0415
 
     cluster = Cluster(
         ClusterConfiguration(
@@ -24,7 +24,7 @@ def ray_fn() -> int:
             worker_cpu_limits=1,
             worker_memory_requests=1,
             worker_memory_limits=2,
-            image="quay.io/modh/ray:2.35.0-py39-cu121",
+            image="quay.io/modh/ray@sha256:0d715f92570a2997381b7cafc0e224cfa25323f18b9545acfd23bc2b71576d06",
             verify_tls=False
         )
     )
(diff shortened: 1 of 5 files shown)


ods_ci/tests/Resources/Files/pipeline-samples/v2/cache-disabled/ray_job_integration.py
```diff
@@ -0,0 +1,382 @@
...
+)
+
+# Train the model
+trainer.fit(model)
+"""
+
+    pip_requirements = """
+pytorch_lightning==1.9.5
+torchmetrics==0.9.1
+torchvision==0.20.1
```
sutaakar, line 258: I have migrated the CFO tests to use latest versions and Python 3.11 - https://github.com/project-codeflare/codeflare-operator/pull/641
Would it have sense to upgrade also here?

ods_ci/tests/Tests/0600__distributed_workloads/0601__workloads_orchestration/test-pipelines-integration.robot
```diff
@@ -0,0 +1,131 @@
...
+    [Documentation]    Creates, runs pipelines with regular user. Double check the pipeline result and clean
+    ...    the pipeline resources.
+    ...    AutomationBugOnDisconnected: RHOAIENG-12514
+    [Tags]      Tier1
+    ...         AutomationBugOnDisconnected
+    ...         DistributedWorkloads
+    ...         WorkloadsOrchestration
+    ...         DataSciencePipelines-DistributedWorkloads
+    Skip If Component Is Not Enabled    ray
+    Skip If Component Is Not Enabled    codeflare
```
sutaakar, line 37: Those components are enabled by default.
IMHO we should remove these checks as disabled component should fail the test to indicate that something is wrong.

ods_ci/tests/Tests/0600__distributed_workloads/0601__workloads_orchestration/test-pipelines-integration.robot
```diff
@@ -0,0 +1,131 @@
...
```
sutaakar, line 67: Same here

sutaakar (approved):
ok for initial approach
would be good to adjust the code to replace versions/images with values from https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource#L7

ods_ci/tests/Resources/Files/pipeline-samples/v2/cache-disabled/ray_integration.py
```diff
@@ -24,7 +24,7 @@ def ray_fn() -> int:
             worker_cpu_limits=1,
             worker_memory_requests=1,
             worker_memory_limits=2,
-            image="quay.io/modh/ray:2.35.0-py39-cu121",
+            image="quay.io/modh/ray@sha256:0d715f92570a2997381b7cafc0e224cfa25323f18b9545acfd23bc2b71576d06",
```
ChughShilpa, line 27: This sha belongs to `quay.io/modh/ray:2.35.0-py39-cu121`. We should use the sha for `quay.io/modh/ray:2.35.0-py311-cu121"`

jiripetrlik: Unfortunately, it is necessary to use Python 3.9 because in interactive mode both server and client must have the same minor Python version and pipelines are currently tested with Python 3.9 images.

ChughShilpa (approved):

PR #2131, opened 2024-12-10
Build ODS-CI test image for release 2.16

liswang89:


bdattoma (changes requested):
pls don't remove the code for OCM. This script is supposed to work with both options

ods_ci/build/install_idp.sh
```diff
@@ -446,10 +430,35 @@ function validate_user_config_file(){
...
 }
 
-if [ "${USE_OCM_IDP}" -eq 1 ]
-      then
-          perform_ocm_login
-fi
+function update_test_config(){
+    echo "Update test config file..."
+
+    AWS_SHARED_CREDENTIALS_FILE="${CLUSTER_PROFILE_DIR}/.awscred"
```
bdattoma, line 436: this would work only for the interop pipeline case, while this script is generic. You can add the feature if you want, but it'd be better to put it under an option of the script/env variable

ods_ci/build/install_idp.sh
```diff
@@ -446,10 +430,35 @@ function validate_user_config_file(){
...
+    yq -i '.S3.AWS_ACCESS_KEY_ID=env(AWS_ACCESS_KEY_ID)' test-variables.yml
+    yq -i '.S3.AWS_SECRET_ACCESS_KEY=env(AWS_SECRET_ACCESS_KEY)' test-variables.yml
+}
+
 validate_user_config_file
 check_installation
 install_identity_provider
+update_test_config
+
+echo "Performing oc login using TEST username and password"
```
bdattoma, line 455: the login is performed earlier in this script with htp user. Why do we need to log in with a different one?

ods_ci/run_interop.sh
```diff
@@ -0,0 +1,48 @@
+!/bin/bash
```
omrirh, line 1:
```suggestion
#!/bin/bash
```

omrirh: following this error:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/58481/rehearse-58481-periodic-ci-red-hat-data-services-ods-ci-master-rhoai2.16-ocp4.18-lp-interop-rhoai-interop-aws/1867044695671050240#1:build-log.txt%3A96

bdattoma:
> Should we set `USE_OCM_IDP=0` ? See error message in job: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/58481/rehearse-58481-periodic-ci-red-hat-data-services-ods-ci-master-rhoai2.16-ocp4.18-lp-interop-rhoai-interop-aws/1867044695671050240#1:build-log.txt%3A98


yes, you should set that env variable to zero in your job.


> And how it is consumed by install_idp.sh:


it's an Environment variable, so that's how install_idp.sh consumes it

ods_ci/run_interop.sh
liswang89: Updated.

ods_ci/run_interop.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+export SET_ENVIRONMENT=1
+export USE_OCM_IDP=0
+export RUN_SCRIPT_ARGS="skip-oclogin true --set-urls-variables true"
+export ROBOT_EXTRA_ARGS="-i Smoke --dryrun"
```
omrirh, line 6:
```suggestion
export ROBOT_EXTRA_ARGS="-i Smoke -e Resources-GPU -e Resources-2GPUS --dryrun"
```

omrirh: Resolving this as Smoke tests does not include GPU tests

PR #2139, opened 2024-12-18
Update AMD Operator and NFD install scripts

bdattoma:
This PR is addressing 2 issues:
1. Certified AMD GPU operator is available only from OCP versions >= 4.16
2. The fixed NFD CR image we use is not compatible with OCP versions > 4.16

Proposed solutions:
1. create a custom catalog source pointing to OCP 4.16 certified operators CS
2. use the NFD CR image corresponding to OCP version

A note about 2: we could fetch the image dynamically using the vx.y label, however this wouldn't be compatible with our disconnected cluster automation. Hence, for now, we are going with fixed image digests

MR validation:
- OCP 4.17 with AMD GPU: `rhoai-test-flow/2157` PASS
- OCP 4.15 with NVIDIA GPU: `rhoai-test-flow/2158` PASS


ods_ci/tasks/Resources/Provisioning/GPU/NFD/install_nfd.sh
```diff
@@ -0,0 +1,47 @@
...
+    ["4.17"]="registry.redhat.io\/openshift4\/ose-node-feature-discovery-rhel9@sha256:154cf3f1ddaf895d7ecd04947bd455a930132f72acc6e8bde8c26bc123184ace"
+    # 4.18 is a pre-release image. We need to update it later
+    ["4.18"]="registry.redhat.io\/openshift4\/ose-node-feature-discovery-rhel9@sha256:510cb4351253492455664b6c323f54dc2f6f2f8791c5e92ba6b7e60b8adb357c"
+)
+if [ "${images[$xyVersion]}" ]; then
+    imageUrl="${images[$xyVersion]}"
+    echo "Using image SHA for $xyVersion: $imageUrl"
+else
+    imageUrl="${images["4.17"]}"
+    echo "I don't know the sha for $xyVersion. Re-using default 4.17 $imageUrl. It might now work!"
```
jstourac, line 44: I would like to see some word here that might come up in some search/grep in case we investigate some problem:

```suggestion
    echo "WARNING: I don't know the sha for $xyVersion. Re-using default 4.17 $imageUrl. It might now work!"
```

jstourac (dismissed):
one comment, but LGTM in general

ods_ci/tasks/Resources/Provisioning/GPU/NFD/install_nfd.sh
bdattoma: agree! done

jstourac (dismissed):

apodhrad (approved):

jstourac (approved):

PR #2154, opened 2025-01-03
Add keyword to find cluster type based on cluster infrastructure

ChughShilpa:
This PR

- Removes keyword "Is Cluster Type Self-Managed" as it was suggested not to rely on console url to find cluster type
- In all self-managed clusters, there is no common key in "oc get infrastructure cluster" output which can be used to 
  identify the cluster-type 
- There is one common key in all managed clusters and this is used to find the cluster-type
   ```
   {
                        "key": "red-hat-managed",
                        "value": "true"
                    }
```

- Adds a keyword "Is Cluster Type Managed" to find the cluster type based on cluster infrastructure

### DIFF
--- a/ods_ci/tests/Resources/OCP.resource
+++ b/ods_ci/tests/Resources/OCP.resource
@@ -333,17 +333,15 @@ Wait For Namespace To Be Active
     Log    ${value}
     Should Be Equal As Integers    ${rc}    0
 
-Is Cluster Type Self-Managed
-    [Documentation]    Get the value of cluster type depending on the console URL domain
-    ...    Returns ${TRUE} if cluster type is Self-Managed
-    ...    Returns ${FALSE} if cluster type is Managed
-    ${matches}=    Get Regexp Matches    ${OCP_CONSOLE_URL}    rh-ods
-    ${size}=    Get Length    ${matches}
-    IF    ${size}>0
-        ${domain}=    Get From List    ${matches}    0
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by sutaakar — APPROVED** [review]

File: ods_ci/tests/Resources/OCP.resource [hunk 1]
```diff
@@ -333,17 +333,15 @@ Wait For Namespace To Be Active
     Log    ${value}
     Should Be Equal As Integers    ${rc}    0
 
-Is Cluster Type Self-Managed
```
jgarciao, line 336: Could you maintain this keyword returning a negation of `Is Cluster Type Managed` ?

Something like:
```
    ${is_managed}=    Is Cluster Type Managed
    RETURN    not ${is_nanaged}
```

ChughShilpa: "Is Cluster Type Managed" keyword covers both the cases for cluster type (managed and self-managed) in one single keyword, which is sufficient in my opinion.

ChughShilpa: I can keep it `Is Cluster Type Self-Managed` if you have a strong preference

jgarciao: No, no need. You are right. Thanks @ChughShilpa !

jgarciao (approved):

PR #2157, opened 2025-01-06
fix the selfmanaged uninstall of rhoai operator

kobihk:
Since we moved to using Konflux, we don't have the same CATALOG_SOURCE_NAME and OPERATOR_NAME for Stage and Red Hat.

Stage:
Operator name: rhoai-operator-dev, CatalogSource: rhoai-catalog-dev

Red Hat:
Operator name: rhods-operator, CatalogSource: redhat-operators

So the uninstall didn't work as expected.

Related to Jira ticket: RHOAIENG-2720

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
@@ -5,8 +5,7 @@ Is RHODS Installed
       IF  "${INSTALL_TYPE}" == "Cli"
           ${result}=  Run Keyword And Return Status
           ...  Run Keywords
-          ...  Check A RHODS Family Operator Is Installed  namespace=${OPERATOR_NAMESPACE}
-          ...                                              subscription=rhoai-operator-dev  AND
+          ...  Check A RHODS Family Operator Is Installed  AND
           ...  Oc Get  kind=Namespace  field_selector=metadata.name=${MONITORING_NAMESPACE}  AND
           ...  Oc Get  kind=Namespace  field_selector=metadata.name=${APPLICATIONS_NAMESPACE}  AND
           ...  Oc Get  kind=Namespace  field_selector=metadata.name=${OPERATOR_NAMESPACE}  AND
@@ -15,8 +14,7 @@ Is RHODS Installed
       ELSE IF  "${INSTALL_TYPE}" == "OperatorHub"
           ${result}=  Run Keyword And Return Status
           ...  Run Keywords
-          ...  Check A RHODS Family Operator Is Installed  namespace=${OPERATOR_NAMESPACE}
-          ...                                              subscription=${OPERATOR_NAME}  AND
+          ...  Check A RHODS Family Operator Is Installed  AND
           ...  Oc Get  kind=Namespace  field_selector=metadata.name=${MONITORING_NAMESPACE}  AND
           ...  Oc Get  kind=Namespace  field_selector=metadata.name=${APPLICATIONS_NAMESPACE}  AND
           ...  Oc Get  kind=Namespace  field_selector=metadata.name=${OPERATOR_NAMESPACE}
@@ -34,8 +32,7 @@ Is RHODS Installed
       END
       ${result}=  Run Keyword And Return Status
       ...  Run Keywords
-      ...  Check A RHODS Family Operator Is Installed  namespace=${OPERATOR_NAMESPACE}
-      ...                                              subscription=${SUB_NAME}  AND
+      ...  Check A RHODS Family Operator Is Installed  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=${MONITORING_NAMESPACE}  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=${APPLICATIONS_NAMESPACE}  AND
       ...  Oc Get  kind=Namespace  field_selector=metadata.name=${OPERATOR_NAMESPACE}  AND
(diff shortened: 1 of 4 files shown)


apodhrad:
Hi @kobihk , in the description you have mentioned `rhoai-operator-dev`, what is that?
My understanding is that the package name is still the same `rhods-operator` and only CatalogSource varies.

ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
```diff
@@ -49,17 +46,40 @@ Is RHODS Installed
...
   END
-  Log   Operator with sub ${subscription} is installed result: ${result}      console=yes
+  Log   Operator with sub '${subscription}' is installed result: ${result}      console=yes
   IF  not ${result}     FAIL    The operator with sub ${subscription} is not installed.
+
+Get RHODS Subscription Name
+    [Documentation]    Returns the subscription name of RHOAI/ODH operator
+    Log   Get the RHODS subscription name by package name: '${RHODS_PACKAGE_NAME}'    console=yes
+    ${rc}    ${out}=    Run And Return RC And Output
+    ...    oc get subscriptions -A | grep ${RHODS_PACKAGE_NAME} | awk '{print $2}'
```
apodhrad, line 73: Using `grep` sometimes leads to multi-line results, for example
```
$ printf "rhods-operator\nrhods-operator-dev\n" | grep rhods-operator
rhods-operator
rhods-operator-dev
```
That's why I recommend using something like this
```
$ oc get sub -A -o json | jq --arg pkgName "${RHODS_PACKAGE_NAME}" -r '.items[] | select(.spec.name==$pkgName) | .metadata.name'
```

ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
```diff
@@ -49,17 +46,40 @@ Is RHODS Installed
...
+    Should Be Equal As Integers    ${rc}    0
+    Log   The RHODS Subscription Name is: '${out}'    console=yes
+    RETURN    ${out}
+
+Get RHODS Namespace
+    [Documentation]    Returns the namespace of RHOAI/ODH operator
+    Log   Get the RHODS namespace by package name: '${RHODS_PACKAGE_NAME}'    console=yes
+    ${rc}    ${out}=    Run And Return RC And Output
+    ...    oc get subscriptions -A | grep ${RHODS_PACKAGE_NAME} | awk '{print $1}'
```
apodhrad, line 82: Using `grep` sometimes leads to multi-line results, for example
```
$ printf "rhods-operator\nrhods-operator-dev\n" | grep rhods-operator
rhods-operator
rhods-operator-dev
```
That's why I recommend using something like this
```
$ oc get sub -A -o json | jq --arg pkgName "${RHODS_PACKAGE_NAME}" -r '.items[] | select(.spec.name==$pkgName) | .metadata.namespace'
```

kobihk:
> Hi @kobihk , in the description you have mentioned `rhoai-operator-dev`, what is that? My understanding is that the package name is still the same `rhods-operator` and only CatalogSource varies.

Thanks @apodhrad !!
I updated the description with `Operator name` to make it clear what is this name

ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
kobihk: Updated!!

ods_ci/tasks/Resources/RHODS_OLM/pre-tasks/oc_is_operator_installed.robot
kobihk: Updated!!

apodhrad (approved):

kobihk:
> /lgtm. Maybe uninstall should be verified on managed cluster as well.

The uninstall of rhoai addon on managed cluster - failed 

AFAICS this is something that we need to fix in: olminstall/cleanup.sh script file so probably it will be in another patch

aloganat:
Hi @kobihk , Cli way of uninstallation using olminstall is common across self managed/managed and odh-nightlies. It varies only in the argument that we pass.

Also, rhoai-operator-dev - Could you confirm if this is the operator name that you are observing in stage ?

ods_ci/tests/Resources/RHOSi.resource
```diff
@@ -84,6 +84,7 @@ Initialize Global Variables
     Set Global Variable   ${RHODS_VERSION}
     Set Prometheus Variables
     Set Global Variable    ${DASHBOARD_APP_NAME}    ${PRODUCT.lower()}-dashboard
+    Set Global Variable    ${RHODS_PACKAGE_NAME}    rhods-operator
```
bdattoma, line 87: are you working on passing this variable from CI/CD job? otherwise ODH uninstall won't work

kobihk: as far as I can see we have the same package name also for ODH:

```
➜  oc get sub -A                                                                                                             
NAMESPACE               NAME                  PACKAGE               SOURCE                CHANNEL
konveyor-forklift       forklift-operator     forklift-operator     community-operators   stable-v2.6
opendatahub-operators   rhoai-operator-dev    rhods-operator        rhoai-catalog-dev     odh-nightlies
...
```

bdattoma (approved):

kobihk:
> Hi @kobihk , Cli way of uninstallation using olminstall is common across self managed/managed and odh-nightlies. It varies only in the argument that we pass.
> 
> Also, rhoai-operator-dev - Could you confirm if this is the operator name that you are observing in stage ?

Here you can see an example of rhoai-operator-dev

image

bdattoma:
> The uninstall of rhoai addon on managed cluster - failed
> 
> AFAICS this is something that we need to fix in: olminstall/cleanup.sh script file so probably it will be in another patch

is it currently working the uninstall for Managed clusters?  I mean, without this PR. 
Also, I  would involve platform team because they also contribute to operator install/uninstall scripts: @asanzgom @CFSNM

CFSNM:
@bdattoma @kobihk hey, we have not seen any issues in our jobs, both selfmanaged and managed when uninstalling ODH/RHOAI

CFSNM:
cc @MarianMacik FYI

PR #2158, opened 2025-01-06
Skip FMS Training operator tests for version less than 2.12.0

sutaakar:
Purpose is to skip tests when testing upgrade path from 2.8 EUS to latest.

Also using raw form to retrieve `startingCSV` to get rid of quotes messing up comparison between versions.


ChughShilpa (approved):
/lgtm

abhijeet-dhumal (approved):

PR #2173, opened 2025-01-13
[fix] the Verify Notebook Has Not Restarted upgrade test

jstourac:
This refactores this test - the first test part is moved to "pre_upgrade" phase, because this phase is run on the pre-upgrade version with the pre-upgrade code-base (branch). This is crucial since the further phases will be run with the code-base of the upgraded product and we may hit some issues in the differences in the product otherwise.

There are done other changes and fixes to the test to make it more robust including the notebook deletion in the post upgrade phase.

Apart from that, to propagate the timestamp value between the phases, this introduces a configmap in the "upgrade" namespace, where we can put some context that should be passed to the followup phase.

https://issues.redhat.com/browse/RHOAIENG-14840

---

At the moment I don't delete the created Config Map anywhere. We can discuss whether and when to do so eventually.

---

CI: tested locally on my machine with following steps:
1. Installed RHOAI 2.16.0
2. Run the pre_upgrade test (solely the `Long Running Jupyter Notebook` test only) - using releases/2.16.0 with this backported - see #2174
3. Performed the upgrade to 2.17 nightly build (manually)
4. Run the post_upgrade test (solely the `Verify Notebook Has Not Restarted` test only) - using this branch which is based on master

Jenkins runs:
* rhoai-test-flow/2352/ - PreUpgrade
* rhoai-test-flow/2353/ - PostUpgrade

### DIFF
--- a/ods_ci/tests/Tests/0200__rhoai_upgrade/0201__pre_upgrade.robot
+++ b/ods_ci/tests/Tests/0200__rhoai_upgrade/0201__pre_upgrade.robot
@@ -24,7 +24,7 @@ Resource            ../../Resources/Page/DistributedWorkloads/DistributedWorkloa
 Resource            ../../Resources/Page/DistributedWorkloads/WorkloadMetricsUI.resource
 Resource            ../../Resources/Page/ModelRegistry/ModelRegistry.resource
 
-Suite Setup         Dashboard Suite Setup
+Suite Setup         Upgrade Suite Setup
 Suite Teardown      RHOSi Teardown
 
 Test Tags           PreUpgrade
@@ -34,6 +34,9 @@ Test Tags           PreUpgrade
 ${CUSTOM_CULLER_TIMEOUT}    60000
 ${S_SIZE}                   25
 ${DW_PROJECT_CREATED}       False
+${CODE}     while True: import time ; time.sleep(10); print ("Hello")
+${UPGRADE_NS}    upgrade
+${UPGRADE_CONFIG_MAP}    upgrade-config-map
 
 
 *** Test Cases ***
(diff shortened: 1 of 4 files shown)


jiridanek (approved):

jiridanek (approved):

kobihk (approved):

ods_ci/tests/Tests/0200__rhoai_upgrade/0201__pre_upgrade.robot
```diff
@@ -246,8 +249,44 @@ Model Registry Pre Upgrade Set Up
...
     Model Registry Pre Upgrade Scenario
 
+Long Running Jupyter Notebook
+    [Documentation]    Launch a long running notebook before the upgrade
+    [Tags]      Upgrade
+    Launch Notebook
+    Add And Run JupyterLab Code Cell In Active Notebook     ${CODE}
+
+    # Get the notebook pod creation timestamp
+    ${notebook_pod_name}=    Set Variable    jupyter-nb-ldap-2dadmin2-0
```
bdattoma, line 259: why not getting the nb name using either `Get User CR Notebook Name` or `Get User Notebook Pod Name`
?

jstourac: good point, it was just copy-paste thing; I'll update that as you proposed :+1:

jstourac: rebased and implemented, thanks!

jiridanek (approved):

kobihk (approved):

PR #2176, opened 2025-01-13
Migration of Onnx  model kserve Rest testcase UI -> API

Raghul-M:


Raghul-M:
Verified with Jenkins Build 2318

bdattoma (approved):

PR #2213, opened 2025-02-04
Add RHOAI Upgrade Testing on OpenShift CI

liswang89:
Update run script to run RHOAI upgrade testing.

### DIFF
--- a/ods_ci/run_interop.sh
+++ b/ods_ci/run_interop.sh
@@ -8,6 +8,25 @@ export ROBOT_EXTRA_ARGS="-i Smoke --dryrun"
 TEST_CASE_FILE="tests/Tests"
 TEST_VARIABLES_FILE="test-variables.yml"
 
+if [[ -z "${TEST_SUITE}" ]]; then
+  echo "Error: TEST_SUITE not set. Please define it. Exiting.."
+  exit 1
+fi
+
+if [[ -z "${ARTIFACT_DIR}" ]]; then
+  echo "ARTIFACT_DIR is empty. Using  default value /tmp."
+  ARTIFACT_DIR="/tmp"
+fi
+
+if [[ ${TEST_SUITE} == "PostUpgrade" ]]; then
+  echo "Retrive test config file..."
+  cp ${SHARED_DIR}/${TEST_VARIABLES_FILE} ${TEST_VARIABLES_FILE}
+
+  echo "Running post-upgrade testing"
+  poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE}
+  exit $?
+fi
+
 echo "Install IDP users and map them to test config file"
 ./build/install_idp.sh
 
@@ -65,14 +84,11 @@ if [ $retVal -ne 0 ]; then
     exit "$retVal"
 fi
 
-if [[ -z "${TEST_SUITE}" ]]; then
-  echo "Define TEST_SUITE"
-  exit 1
+if [[ ${TEST_SUITE} == "PreUpgrade" ]]; then
+  echo "Save test config file..."
+  cp ${TEST_VARIABLES_FILE} ${SHARED_DIR}/${TEST_VARIABLES_FILE}
+
+  echo "Running pre-upgrade testing"
 fi
 
-if [[ -z "${ARTIFACT_DIR}" ]]; then
-  echo "Define ARTIFACT_DIR"
-  ARTIFACT_DIR="/tmp"
-fi
-
-poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE}
+poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE}
--- a/ods_ci/build/Dockerfile_interop
+++ b/ods_ci/build/Dockerfile_interop
@@ -12,6 +12,7 @@ ARG PYTHON_VERSION=3.11
 
 ENV ODS_VENV="/ods_venv" \
     HOME="/ods_venv" \
+    WORKSPACE="/ods_venv/ods_ci" \
     PATH="ods_venv/bin:${PATH}" \
     AM_I_IN_CONTAINER="Yes" \
     PYTHONUNBUFFERED="True"


bdattoma (dismissed):

ods_ci/run_interop.sh
```diff
@@ -8,6 +8,26 @@ export ROBOT_EXTRA_ARGS="-i Smoke --dryrun"
 TEST_CASE_FILE="tests/Tests"
 TEST_VARIABLES_FILE="test-variables.yml"
 
+if [[ -z "${TEST_SUITE}" ]]; then
+  echo "Define TEST_SUITE"
```
bdattoma, line 12: maybe for easier debugging, you could print something like " TEST_SUITE not set. Please define it. Exiting.."

This would help distinguishing from lines 16-17-18-19, where the error message is the same but it doesn't exit

ods_ci/run_interop.sh
```diff
@@ -65,14 +85,13 @@ if [ $retVal -ne 0 ]; then
...
+if [[ ${TEST_SUITE} == "Pre-Upgrade" ]]; then
+  echo "Save test config file..."
+  cp test-variables.yml ${SHARED_DIR}/test-variables.yml
 
-if [[ -z "${ARTIFACT_DIR}" ]]; then
-  echo "Define ARTIFACT_DIR"
-  ARTIFACT_DIR="/tmp"
+  echo "Running pre-upgrade testing"
+  TEST_CASE_FILE="tests/Tests/0200__rhoai_upgrade/0201__pre_upgrade.robot"
+  poetry run robot -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE} || true
```
bdattoma, line 94: why you want it to always return exit code 0?
and why don't you want the same for the `else` part?

ods_ci/run_interop.sh
apodhrad: Moreover, is it also good to state that this is an error, for example
```
echo "Error: ..."
echo "[ERROR] ..."
```

This is useful when we are checking logs (we usually search for `error` string)

ods_ci/run_interop.sh
```diff
@@ -8,6 +8,26 @@ export ROBOT_EXTRA_ARGS="-i Smoke --dryrun"
 TEST_CASE_FILE="tests/Tests"
 TEST_VARIABLES_FILE="test-variables.yml"
 
+if [[ -z "${TEST_SUITE}" ]]; then
+  echo "Define TEST_SUITE"
+  exit 1
+fi
+
+if [[ -z "${ARTIFACT_DIR}" ]]; then
+  echo "Define ARTIFACT_DIR"
```
apodhrad, line 17: Please consider something like
```
ARTIFACT_DIR is empty. Using  default value /tmp.
```

ods_ci/run_interop.sh
liswang89: Updated.

ods_ci/run_interop.sh
liswang89: updated.

ods_ci/run_interop.sh
liswang89: yes, any failure will stop the pipeline and we want to run post-upgrade tests even if there are failures in pre-upgrade.

omrirh (approved):

omrirh (dismissed):

ods_ci/run_interop.sh
```diff
@@ -65,14 +85,13 @@ if [ $retVal -ne 0 ]; then
...
-fi
+if [[ ${TEST_SUITE} == "Pre-Upgrade" ]]; then
+  echo "Save test config file..."
+  cp test-variables.yml ${SHARED_DIR}/test-variables.yml
 
-if [[ -z "${ARTIFACT_DIR}" ]]; then
-  echo "Define ARTIFACT_DIR"
-  ARTIFACT_DIR="/tmp"
+  echo "Running pre-upgrade testing"
+  TEST_CASE_FILE="tests/Tests/0200__rhoai_upgrade/0201__pre_upgrade.robot"
```
kobihk, line 93: Since we changed the path of these files in 2.12 if you run it with 2.8 you need to use a different path like:
tests/Tests/100__deploy/120__upgrades/120__pre_upgrade.robot
So it will be better to include the Tag `PreUpgrade` which is good for all branches.

ods_ci/run_interop.sh
```diff
@@ -65,14 +85,13 @@ if [ $retVal -ne 0 ]; then
...
+  poetry run robot -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE} || true
+else
+  poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE}
```
kobihk, line 96: Maybe it will be better to add also 
`--exclude "ProductBug"`

ods_ci/run_interop.sh
```diff
@@ -8,6 +8,26 @@ export ROBOT_EXTRA_ARGS="-i Smoke --dryrun"
...
+  echo "ARTIFACT_DIR is empty. Using  default value /tmp."
+  ARTIFACT_DIR="/tmp"
+fi
+
+if [[ ${TEST_SUITE} == "Post-Upgrade" ]]; then
+  echo "Retrive test config file..."
+  cp ${SHARED_DIR}/test-variables.yml test-variables.yml
+
+  echo "Running post-upgrade testing"
+  TEST_CASE_FILE="tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot"
```
kobihk, line 26: I think it will be better to use the tag `PostUpgrade` 
for more details, you can see the comment below https://github.com/red-hat-data-services/ods-ci/pull/2213/files#diff-2d80a284294fa1f65798ca0f1558ac396aabd2075ea029de95b2c23a40d961cdR93

ods_ci/run_interop.sh
```diff
@@ -8,6 +8,26 @@ export ROBOT_EXTRA_ARGS="-i Smoke --dryrun"
...
+  poetry run robot -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE}
```
kobihk, line 27: Maybe it will be better to add also:
` --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug"`

ods_ci/run_interop.sh
kobihk, line 94: Maybe it will be better to add also:
 --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug"

ods_ci/run_interop.sh
liswang89: Without the "ProductBug" exclusion, we now have 69 test cases from Smoke, which is the same as with the exclusion. I will keep it as is for now and add it back if needed.

kobihk: These tags are generic and could be added from time to time. 
so it is better to add `--exclude` to ensure we will not run them when someone will add it to a specific test

ods_ci/run_interop.sh
liswang89: updated

ods_ci/run_interop.sh
liswang89: updated

ods_ci/run_interop.sh
```diff
@@ -65,14 +84,11 @@ if [ $retVal -ne 0 ]; then
     exit "$retVal"
 fi
 
-if [[ -z "${TEST_SUITE}" ]]; then
-  echo "Define TEST_SUITE"
-  exit 1
-fi
+if [[ ${TEST_SUITE} == "PreUpgrade" ]]; then
+  echo "Save test config file..."
+  cp test-variables.yml ${SHARED_DIR}/test-variables.yml
```
kobihk, line 89: I think it will be better to use the defined variable TEST_VARIABLES_FILE instead of writing the file name like:
```suggestion
  cp ${TEST_VARIABLES_FILE} ${SHARED_DIR}/${TEST_VARIABLES_FILE}
```

omrirh (dismissed):

ods_ci/run_interop.sh
liswang89: updated

ods_ci/run_interop.sh
```diff
@@ -8,6 +8,25 @@ export ROBOT_EXTRA_ARGS="-i Smoke --dryrun"
...
+fi
+
+if [[ -z "${ARTIFACT_DIR}" ]]; then
+  echo "ARTIFACT_DIR is empty. Using  default value /tmp."
+  ARTIFACT_DIR="/tmp"
+fi
+
+if [[ ${TEST_SUITE} == "PostUpgrade" ]]; then
+  echo "Retrive test config file..."
+  cp ${SHARED_DIR}/test-variables.yml test-variables.yml
```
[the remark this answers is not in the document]
liswang89: updated

kobihk (approved):

omrirh (approved):

PR #2217, opened 2025-02-06
Disbale kueue to run ray e2e tests

ChughShilpa:
Disbale kueue to run ray e2e tests to avoid ValidatingAdmissionPolicy check


ChughShilpa:
Test passed in `rhoai-test-flow/2740`

abhijeet-dhumal:
/lgtm
Thank you!

jiripetrlik:
Hello @ChughShilpa why do we need to disable Kueue? Is there any issue related to it?

sutaakar (approved):
Thanks

sutaakar:
@jiripetrlik The reason is that KubeRay upstream e2e tests don't use Kueue, so they don't contain local queue label.
As a result Kueue ValidatingAdmissionPolicy blocks RayClusters created by e2e tests.

Disabling Kueue avoids ValidatingAdmissionPolicy issue.

jiripetrlik:
@sutaakar Thank you, makes sense to me.

jiripetrlik (approved):

PR #2222, opened 2025-02-11
Update Verify RHODS User Groups Test Case to new Auth CRD

asanzgom:
rhoai-test-flow/2786/


CFSNM (dismissed):
LGTM! Lets backport it to the 2.17 branch

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
```diff
@@ -55,11 +56,12 @@ Verify Pod Toleration
 
 Verify RHODS User Groups
     [Documentation]    Verify User Configuration after the upgrade
-    [Tags]      Upgrade
-    ${admin}        Set Variable        ${payload[0]['spec']['groupsConfig']['adminGroups']}
-    ${user}     Set Variable        ${payload[0]['spec']['groupsConfig']['allowedGroups']}
-    Should Be Equal As Strings      '${admin}'      'rhods-admins,rhods-users'
```
MarianMacik, line 61: What about check for the rhods-users? This group is still assigned as admins in the pre-upgrade test suite.
https://github.com/red-hat-data-services/ods-ci/blob/cb949633d30b38a7a524219aa5f611770fcd52a9/ods_ci/tests/Tests/0200__rhoai_upgrade/0201__pre_upgrade.robot#L78

MarianMacik (commented):
FYI there are other tests which look at OdhDashboardConfig, but they currently fail on UI elements not visible, so they don't reach the assertion phase.
Just an example of such test:
https://github.com/red-hat-data-services/ods-ci/blob/59f14dbc50e9d85f30ad4540a9dffa59129e7327/ods_ci/tests/Tests/0400__ods_dashboard/0405__ods_dashboard_user_mgmt.robot#L25-L29

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
asanzgom: that group is not used by default https://github.com/opendatahub-io/opendatahub-operator/blob/main/controllers/components/dashboard/dashboard_support.go#L34 @MarianMacik 
I can update the pre-upgrade test suite on another PR

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
```diff
@@ -55,11 +56,12 @@ Verify Pod Toleration
...
-    Should Be Equal As Strings      '${user}'       'system:authenticated'
+    [Tags]      Upgrade     Operator
+    Get Auth Cr Config Data
+    ${admin}                Set Variable            ${AUTH_PAYLOAD[0]['spec']['adminGroups']}
+    ${user}                 Set Variable            ${AUTH_PAYLOAD[0]['spec']['allowedGroups']}
```
bdattoma, line 62: I think instead of reading the expected values from the cluster itself, we should impost the expected truth for the test.
So, if in the pre-upgrade tests we set 2 specific groups as admins, we should check in post upgrade that those groups are still set (in either UI and CRD or both)

cc: @jstourac @manosnoam

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
MarianMacik: @asanzgom but that group is explicitly added in the pre-upgrade test suite, so will it actually work? We are explicitly setting these 2 groups as administrators in pre-upgrade test suite, so in post-upgrade we should check both groups are present, right?

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
```diff
@@ -55,11 +56,12 @@ Verify Pod Toleration
 
 Verify RHODS User Groups
     [Documentation]    Verify User Configuration after the upgrade
-    [Tags]      Upgrade
-    ${admin}        Set Variable        ${payload[0]['spec']['groupsConfig']['adminGroups']}
-    ${user}     Set Variable        ${payload[0]['spec']['groupsConfig']['allowedGroups']}
-    Should Be Equal As Strings      '${admin}'      'rhods-admins,rhods-users'
-    Should Be Equal As Strings      '${user}'       'system:authenticated'
+    [Tags]      Upgrade     Operator
```
MarianMacik, line 59: I noticed the added tag `Operator` but this test won't work without running pre-upgrade test suite as it expects different values than default. We are setting these values in the pre-upgrade test suite here:
https://github.com/red-hat-data-services/ods-ci/blob/cb949633d30b38a7a524219aa5f611770fcd52a9/ods_ci/tests/Tests/0200__rhoai_upgrade/0201__pre_upgrade.robot#L78

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
bdattoma: and those are different from the ones set during pre-upgrade https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tests/Tests/0200__rhoai_upgrade/0201__pre_upgrade.robot#L78 

The test should check the current user group set in the Auth CR matches the ones we set during pre-upgrade no?


>I think instead of reading the expected values from the cluster itself,


yeah this was wrong sorry

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
asanzgom: I will substitute by RHOAING tag

asanzgom: I will replace by RHOAIENG tag

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
asanzgom: yes, it should check basically what is set on Auth CR

ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot
asanzgom: You are right @MarianMacik I will add it too

asanzgom:
rhoai-test-flow/3220 passed successfully

MarianMacik (approved):

PR #2223, opened 2025-02-11
KubeRay: Make sure that ValidatingAdmissionPolicy is removed

sutaakar:
There is small delay between disabling Kueue component and removing ValidatingAdmissionPolicy. As a result the KubeRay tests can occasionally crash as ValidatingAdmissionPolicy is enforced.

Waiting for ValidatingAdmissionPolicy removal makes sure that the validation is gone and KubeRay tests may run.


sutaakar:
ValidatingAdmissionPolicy is compatible with latest OCP only, closing the PR, will use a different approach, probably based on waiting.

PR #2227, opened 2025-02-12
KubeRay: Sleep for 3 seconds after disabling Kueue

sutaakar:
ValidatingAdmissionPolicy is supported just on newer OCP, using Sleep as a workaround to make sure that ValidatingAdmissionPolicy is removed if it exists.

The sleep is preventing race condition which would be caused by running KubeRay tests while ValidatingAdmissionPolicy is still available.


ChughShilpa (approved):
Thanks Karel

abhijeet-dhumal (approved):
/lgtm
Thank you!

PR #2230, opened 2025-02-13
Use OCP_API_URL parameter for Login To OCP function

sutaakar:
That will make sure that OCP URL from `test-variables.yml` is used regardless of current `oc` session.
Also printing the output to console in case of logging error to simplify analysis of possible errors.

### DIFF
--- a/ods_ci/tests/Resources/Page/OCPLogin/OCPLogin.robot
+++ b/ods_ci/tests/Resources/Page/OCPLogin/OCPLogin.robot
@@ -13,11 +13,13 @@ Login To OCP
 Login To OCP Using API
   [Documentation]   Login to openshitf using username and password
   [Arguments]    ${username}      ${password}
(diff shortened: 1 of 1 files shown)


Srihari1192 (approved):

ChughShilpa:
I think jenkins always use OCP URL from test-variables.yml to login, in local execution it may differ how we run the tests.
What's the need for this change, did you encounter any discrepancy with execution ?

sutaakar:
@ChughShilpa There is an issue in upgrade tests ran in OpenShift CI.
One possible explanation is that OpenShift URL in `oc` CLI context differ from `OCP_API_URL`. This change should make sure that `OCP_API_URL` value is really used.

ChughShilpa (approved):
Thanks @sutaakar for the explanation, that could be a reason

PR #2234, opened 2025-02-14
Add ProductBug Tag to Validate SMCP Already created and negative TC

asanzgom:
Adding `ProductBug` Tag to both Test Cases until fix is delivered so Pipelines stay green


CFSNM (approved):
@asanzgom lets backport it to 2.17 branch

MarianMacik (approved):

PR #2242, opened 2025-02-24
Fix locators on Verify RHODS Accept Multiple Admin Groups And CRD Get…

asanzgom:
Fix outdated locators on Verify RHODS Accept Multiple Admin Groups And CRD Gets Updates Test Case
rhoai-test-flow/3050/


asanzgom:
rhoai-test-flow/3050/ passed successfully

MarianMacik (approved):

PR #2244, opened 2025-02-24
Add Interop files

omrirh:
required for https://github.com/openshift/release/pull/61959

PR #2251, opened 2025-02-26
Fix PreUpgrade testing return code to always return 0

liswang89:
When PreUpgrade testing fails, it returns 1 and the pipeline fails as well. We want the PreUpgrade testing to always return 0 so we can continue the PostUpgrade testing.

### DIFF
--- a/ods_ci/run_interop.sh
+++ b/ods_ci/run_interop.sh
@@ -24,7 +24,7 @@ if [[ ${TEST_SUITE} == "PostUpgrade" ]]; then
 
   echo "Running post-upgrade testing"
   poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE}
-  exit $?
+  exit 0
 fi
 
 echo "Install IDP users and map them to test config file"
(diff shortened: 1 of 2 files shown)


omrirh (approved):

ods_ci/run_interop.sh
```diff
@@ -89,6 +89,7 @@ if [[ ${TEST_SUITE} == "PreUpgrade" ]]; then
   cp ${TEST_VARIABLES_FILE} ${SHARED_DIR}/${TEST_VARIABLES_FILE}
 
   echo "Running pre-upgrade testing"
+  poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE} || true
```
kobihk, line 92: maybe it will be better to add the same changes @omrirh did in #2244 to this file

liswang89: @kobihk If we want to continue to run PostUpgrade and Smoke tests, we cannot return $? instead, use " || true " to always return 0.

liswang89: @kobihk for PostUpgrade testing, There is no need to run the create IDP users again. It has been saved from PreUpgrade testing.

kobihk: I see.
So maybe we need to fix it in #2244 or to add 
```
if [[ ${TEST_SUITE} == "PreUpgrade" ]]; then
```
around create IDP

cc: @omrirh

kobihk (approved):

ods_ci/run_interop.sh
liswang89: @kobihk If the test suite is not PostUpgrade, it will create IDP anyway, because we need to share the code base with Smoke or other tests. We only need to exclude it for PostUpgrade test for now.

liswang89: @kobihk If the test suite is not PostUpgrade, it will create IDP anyway, because we need to share the code base with Smoke or other tests. We only need to exclude it for PostUpgrade test for now.

bdattoma (approved):

PR #2254, opened 2025-03-03
Fix RayCluster upgrade tests permission issue for OCP

Srihari1192:


sutaakar (approved):
Thanks

ChughShilpa:
Isn't these upgrade tests should be run by test user instead of admin user ? 
I understand that these tests are failing in Openshift CI but that's not an ideal thing to do in my opinion

Srihari1192:
> Isn't these upgrade tests should be run by test user instead of admin user ? I understand that these tests are failing in Openshift CI but that's not an ideal thing to do in my opinion

@ChughShilpa SDK Upgrade tests authentication using cluster admin login with CI login [here](https://github.com/project-codeflare/codeflare-sdk/blob/main/tests/upgrade/raycluster_sdk_upgrade_test.py#L39) , so the admin login credentials to be used for UI login too , thats why this change. As Ray Clusters creation should be supported by different users including admins, i feel its ok to have here admin login

If we want to change this completely to test user auth login while running should be done by specific test user and same user needs to be handled in UI, not sure how we can handle here. Any thoughts or solution about this?

Srihari1192:
> I think that if we want to use test user then we would need to pass its credentials i.e. by using env variables, not to rely on `oc whoami`.

Yes One possible way that, we can pass Test user name and password as an environment variable and perform OC login in sdk tests itself with `oc login ${OCP_API_URL} -u ${username} -p ${password} --kubeconfig=${kubeconfig} --insecure-skip-tls-verify=true`

ChughShilpa:
We could try running sdk-upgrade tests with test user as like we run [mnist_ray tests](https://github.com/opendatahub-io/distributed-workloads/blob/main/tests/odh/mnist_ray_test.go#L123-L124) and use the same user to login to UI as well

Srihari1192:
@ChughShilpa @sutaakar The error is not caused by Ray cluster authentication creation but due to access permission for project creation user and the logged-in user. SDK tests use by default kube config login in CI to admin user permissions and in UI we are approaching to different user..

One possible approach is performing oc login within this process of[ running SDK tests ](https://github.com/red-hat-data-services/ods-ci/blob/master/ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource#L74) , but i tried several ways its not allowing to run oc login inside the RUN process of same shell session 

Error `Keyword 'Process.Run Process' expected at least 1 non-named argument, got 0.`

sutaakar:
@Srihari1192 What about using token authentication within the SDK, using the token passed from ODS CI?
https://project-codeflare.github.io/codeflare-sdk/user-docs/authentication.html#method-1-token-authentication

Srihari1192:
@sutaakar @ChughShilpa Update the tests to ensure TestUser logs in before running SDK tests. This synchronization guarantees that project creation and UI test user authentication align, preventing access permission errors. Please review

sutaakar (approved):

Srihari1192:
> @Srihari1192 just to be sure, does test user have rights to create cluster wide resources like ResourceFlavor and ClusterQueue?

Yes able to create the kueue resources successfully with TEST_USER in jenkins

Bobbins228 (commented):
@sutaakar @Srihari1192 I thought that regular users or in this case the "test user" would normally not have permission to create cluster resources such as Resource Flavors and ClusterQueues?

ChughShilpa:
@Srihari1192 
QE-jenkins don't have"TEST_USER" context stored in kubeconfig at present so `Login To OCP Using API And Kubeconfig` doesn't work as expected.
This is an ongoing discussion for this, so as temporary fix, you can try generating kubeconfig for TEST_USER in the tests

Srihari1192:
> @sutaakar @Srihari1192 I thought that regular users or in this case the "test user" would normally not have permission to create cluster resources such as Resource Flavors and ClusterQueues?

@Bobbins228 Test user which we are referring here group.user.openshift.io/rhods-admins added: "ldap-admin1

> @Srihari1192 QE-jenkins don't have"TEST_USER" context stored in kubeconfig at present so `Login To OCP Using API And Kubeconfig` doesn't work as expected. There is an ongoing discussion for this, so as temporary fix, you can try generating kubeconfig for TEST_USER in the tests

@ChughShilpa We can use `Login To OCP Using API    ${TEST_USER.USERNAME}    ${TEST_USER.PASSWORD}` method instead it worked fine in CI `rhoai-test-flow/3216`

Srihari1192:
> @sutaakar @Srihari1192 I thought that regular users or in this case the "test user" would normally not have permission to create cluster resources such as Resource Flavors and ClusterQueues?

@Bobbins228 The Test variables we are referring as multiple Test user with different access such as `TEST_USER, TEST_USER_2 and so on` The Test user TEST_USER which we are using here as enough permission to create kueue resources

ChughShilpa (approved):
/lgtm

abhijeet-dhumal (approved):
/lgtm

sutaakar (approved):

PR #2260, opened 2025-03-04
Adjust fms-hf-tuning tests to use Granite model

sutaakar:
Needed due to changes in fms-hf-tuning 2.6.0 image.


ChughShilpa (approved):
/lgtm

PR #2261, opened 2025-03-04
Add Smoke tests to Upgrade

liswang89:
Add Smoke tests after the upgrade completes, ensure they are executed even if the upgrade testing fails.

### DIFF
--- a/ods_ci/run_interop.sh
+++ b/ods_ci/run_interop.sh
@@ -5,9 +5,6 @@ export USE_OCM_IDP=0
 export RUN_SCRIPT_ARGS="skip-oclogin true --set-urls-variables true"
 export ROBOT_EXTRA_ARGS="-i Smoke --dryrun"
 
-TEST_CASE_FILE="tests/Tests"
-TEST_VARIABLES_FILE="test-variables.yml"
-
 if [[ -z "${TEST_SUITE}" ]]; then
   echo "Error: TEST_SUITE not set. Please define it. Exiting.."
   exit 1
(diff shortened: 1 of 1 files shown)


ods_ci/run_interop.sh
```diff
@@ -23,7 +23,11 @@ if [[ ${TEST_SUITE} == "PostUpgrade" ]]; then
   cp ${SHARED_DIR}/${TEST_VARIABLES_FILE} ${TEST_VARIABLES_FILE}
 
   echo "Running post-upgrade testing"
-  poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE}
+  poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR}/${TEST_SUITE} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE} || true
+  
+  echo "Running Smoke testing after upgrade"
+  TEST_SUITE="Smoke"
+  poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR}/${TEST_SUITE} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE} || true
```
kobihk, line 30: Just a suggestion for enhancment:
Instead of repeating almost the same command a couple of times, you can create a method like:

```suggestion
run_tests() {
  TEST_SUITE=$1
  BEFORE_AFTER=$2
  echo "Running $1 testing $2 upgrade"
  poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR}/${TEST_SUITE} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE} || true
}

run_tests "Smoke" "after"
```

kobihk (dismissed):

kobihk (approved):

ods_ci/run_interop.sh
```diff
@@ -18,12 +15,23 @@ if [[ -z "${ARTIFACT_DIR}" ]]; then
...
 fi
 
+run_tests() {
+  echo "Running $1 testing"
+  
+  TEST_CASE_FILE="tests/Tests"
+  TEST_VARIABLES_FILE="test-variables.yml"
+  TEST_SUITE=$1
+
+  poetry run robot --include ${TEST_SUITE} --exclude "ExcludeOnRHOAI" --exclude "AutomationBug" --exclude "ProductBug" -d ${ARTIFACT_DIR}/${TEST_SUITE} -x xunit_test_result.xml -r test_report.html --variablefile ${TEST_VARIABLES_FILE} ${TEST_CASE_FILE} || true
```
bdattoma, line 25: to be more clear, ${TEST_SUITE} could be named ${TEST_SUITE_TAG}, because we're passing the a Tag not the suite name nor the suite path

bdattoma (approved):

ods_ci/run_interop.sh
liswang89: Yes, it is a tag, it is also a test suite that stands for different test categories.

PR #2293, opened 2025-03-12
fix: adapt disruptive tests to ODH/RHOAI 2.19

CFSNM:


ods_ci/tests/Tests/2001__disruptive_tests/2003__smcp_already_created.robot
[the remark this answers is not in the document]
CFSNM: yes

ods_ci/tests/Tests/2001__disruptive_tests/2002__dsc_negative_dependant_operators_not_installed.robot
```diff
@@ -220,3 +257,15 @@ DataScienceCluster Should Fail Because Serverless Operator Is Not Installed
     ...    oc get DataScienceCluster ${DSC_NAME} -n ${OPERATOR_NS} -o json | jq -r '.status.conditions | map(.message) | join(",")'    #robocop:disable
     Should Be Equal As Integers  ${return_code}   0   msg=Error retrieved DSC conditions
     Should Contain    ${output}    Serverless operator must be installed for this component's configuration    #robocop:disable
+
+DataScienceCluster Should Fail Because Service Mesh Is Not Present And Model Registry Is Managed
+    [Documentation]   Keyword to check the DSC conditions when Service Mesh instance is not present.
+    ...           One condition should appear saying this instance is needed to make ModelRegistry work.
+    ${modelregistry} =    Is Component Enabled    modelregistry    ${DSC_NAME}
+    IF    "${modelregistry}" == "true"
```
MarianMacik, line 265: Maybe this if is not needed as we run this keyword only if registry is managed (it is in the name of the keyword).

ods_ci/tests/Tests/2001__disruptive_tests/2002__dsc_negative_dependant_operators_not_installed.robot
```diff
@@ -117,12 +120,46 @@ Validate DSC and DSCI Created With No Errors When Kserve Serving Is Unmanaged An
...
     [Teardown]    Reinstall Service Mesh And Serverless Operators And Recreate DSC And DSCI
 
+Validate DSC and DSCI Created With Errors When Kserve Serving Is Unmanaged And ModelRegistry Is Managed And Service Mesh And Serverless Operators Are Not Installed    #robocop:disable
+    [Documentation]    The purpose of this Test Case is to validate that DSC and DSCI are created
+    ...                without dependant operators ((servicemesh, serverless) installed and with errors
+    ...                because the Kserve component serving is unmanaged and modelregistry component is managed
+    [Tags]    Operator    Tier3     RHOAIENG-3472
+
+    ${modelregistry} =    Is Component Enabled    modelregistry    ${DSC_NAME}
+    Skip If     "${modelregistry}" == "false"     msg=This test is skipped if Model Registry component is Removed
```
MarianMacik, line 130: Maybe it would be better to skip if product is ODH and in the other test RHOAI?

asanzgom (approved):

ods_ci/tests/Resources/ODS.robot
```diff
@@ -404,10 +404,10 @@ Wait For DSCI Ready State
     ...    oc wait --timeout=${wait_time} --for jsonpath='{.status.conditions[].reason}'=ReconcileCompleted -n ${namespace} dsci ${dsci}
     Should Be Equal    "${rc}"    "0"     msg=${output}
 
-Wait For DSC Conditions Reconciled
+Wait For DSC Ready State
     [Documentation]    Checks all DSC conditions to be successfully reconciled
     [Arguments]    ${namespace}    ${dsc_name}    ${wait_time}=20m
     ${rc}    ${out}=    Run And Return Rc And Output
-    ...    oc wait --timeout=${wait_time} --for jsonpath='{.status.conditions[].reason}'=ReconcileCompleted -n ${namespace} dsc ${dsc_name}    # robocop: disable
+    ...    oc wait --timeout=${wait_time} --for jsonpath='{.status.conditions[].type}'=Ready -n ${namespace} dsc ${dsc_name}    # robocop: disable
```
MarianMacik, line 411: When I tried this locally, this returns just the type of the first condition (apparently conditions[] is the same as conditions[0]), which is expected to be Ready. But the condition has to have `status: 'True'`. Previously this was basically waiting for the reason of the first condition to be `ReconcileCompleted` as the first condition was before of type `Available`.
Now if the first condition is of type `Ready` the wait will succeed no matter what is the real status, it might be `False`.

ods_ci/tests/Tests/2001__disruptive_tests/2002__dsc_negative_dependant_operators_not_installed.robot
CFSNM: Its not about ODH or RHOAI, but the status of modelregistry.

E.g. when MR becomes GA in RHOAI, then the same test will be executed in both products

ods_ci/tests/Resources/ODS.robot
CFSNM: we can revisit that on Monday to find a more robust condition

ods_ci/tests/Tests/2001__disruptive_tests/2002__dsc_negative_dependant_operators_not_installed.robot
CFSNM: I was just adding this double condition to avoid a misuse of the kw, but yeah, it can be removed

MarianMacik (approved):
I'll approve to unblock the nightly. We can revisit next week.

PR #2301, opened 2025-03-18
Add fix for poetry path

omrirh:


aloganat (approved):
lgtm

kobihk (approved):
For the next time:
we shouldn't use the 2.10 anymore since it is EOL
Red Hat OpenShift AI Self-Managed Life Cycle Dates
https://access.redhat.com/support/policy/updates/rhoai-sm/lifecycle
or you can see it also in our Jenkins repo in:
https://gitlab.cee.redhat.com/ods/jenkins/-/blob/master/resources/configs/rhoai-releases.yaml?ref_type=heads

PR #2309, opened 2025-03-20
Run KFTO upgrade tests for RHOAI 2.19 onwards

sutaakar:
Enabling Training operator upgrade testing from version when operator is GA.
This will make sure that upgrade tests won't mess up operator disabling/enabling.


ChughShilpa (approved):
Thanks

PR #2311, opened 2025-03-20
fix the 503 service unavailable

kobihk:


ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -251,7 +252,7 @@ Login To Cluster
 Get Cluster Pool Namespace
     [Arguments]    ${hive_pool_name}
     Log    Cluster pool name is: ${hive_pool_name}     console=True
-    ${namespace} =    Wait Until Keyword Succeeds    2 min    2 s
+    ${namespace} =    Wait Until Keyword Succeeds    2 min    10 s
```
bdattoma, line 255: if the suggestion we got to change [lines 8-9](https://github.com/red-hat-data-services/ods-ci/pull/2311/files#diff-069081f68a5b54944843728104321c2a7f2d596f186734b832ddf7d8265185b7R8-R9) works, i think we can reduce the time between attempts to 2s again, or 5 tops.

bdattoma (approved):

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
kobihk: I agree that we can reduce it in the future patch

PR #2313, opened 2025-03-21
fix: delete auth when deleting dsci

CFSNM:
added a new condition for the tests that delete the DSCI to also delete the auth CR


MarianMacik (approved):
LGTM!

PR #2350, opened 2025-04-01
RHOAIENG-13079 - ODS-514: Verify that no false alerts are fired while…

MarianMacik:
… installing RHOAI

This PR:
* Makes `Alert Should Not Be Firing` fail if the alert is not present. Previously if an alert was not found the keyword would succeed anyway.
* Included a check for number of non-firing alerts (currently 47) and firing alerts (just 1 - DeadManSnitch). This way the test for alerts will be kept up to date if new alerts come, forcing us to update the test.
* [Not related to this test but noticed during the development] Added missing recording and alert groups.

Tested using /job/devops/job/rhoai-test-flow/3723/


CFSNM (commented):
just to know

CFSNM (commented):
is this new test being skipped for selfmanaged clusters?

MarianMacik:
Yes, it is here: https://github.com/red-hat-data-services/ods-ci/blob/cb7b4a61f78a324cf8c14697030724f85d0995f6/ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot#L257

asanzgom (approved):

CFSNM (approved):

CFSNM (approved):

CFSNM (approved):

asanzgom (approved):

PR #2351, opened 2025-04-02
Skip Training operator smoke check for RHOAI < 2.19.0

sutaakar:
Check is skipped for older versions to prevent smoke tests after upgrade to fail, as before 2.19.0 Training operator was disabled by default.


ChughShilpa (approved):
Thanks

abhijeet-dhumal (approved):

PR #2352, opened 2025-04-02
Install Authorino from stable channel for both RHOAI (>= 2.20) and ODH

aloganat:


bdattoma:
@aloganat valid starting from 2.19?

bdattoma (approved):

aloganat:
@bdattoma It will be valid from 2.20. But its not confirmed yet.

lugi0:
Thanks for setting it up @aloganat, let's hold this change until we get the go ahead from management

kobihk (approved):

lugi0 (approved):

PR #2367, opened 2025-04-07
Improve Component Should Be Enabled robustness

asanzgom:
Improve the 'Component Should /Not Be Enabled' Keyword robustness through a WHILE loop to allow time and several tries for the DSC to leave the component in the final Managed/Removed state

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -436,13 +436,19 @@ Wait For DataScienceCluster CustomResource To Be Ready
 
 Component Should Be Enabled
     [Arguments]    ${component}    ${dsc_name}=${DSC_NAME}
-    ${status} =    Is Component Enabled    ${component}    ${dsc_name}
-    IF    '${status}' != 'true'    Fail
+    ${status} =   Set Variable   False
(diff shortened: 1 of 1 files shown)


asanzgom:
rhoai-test-flow/3838 succeeded

CFSNM (approved):

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -436,13 +436,19 @@ Wait For DataScienceCluster CustomResource To Be Ready
 
 Component Should Be Enabled
     [Arguments]    ${component}    ${dsc_name}=${DSC_NAME}
-    ${status} =    Is Component Enabled    ${component}    ${dsc_name}
-    IF    '${status}' != 'true'    Fail
+    ${status} =   Set Variable   False
+    WHILE   '${status}' != 'true'    limit=60 seconds
```
asmigala, line 440: One concern I have is that doing this in a loop will hammer the cluster with a lot of requests for 1 minute in case the component is not enabled, should we maybe add a short sleep to the loop body?

asanzgom: mmm we have not had issues with hammering the cluster AFAIK

asmigala (approved):

PR #2373, opened 2025-04-08
Add debugging kws to randomly failing trusted ca bundles managed tc

asanzgom:
These are changes to improve the verbosity and visibility when Trusted CA Bundles Managed TC fails randomly to better diagnose how to make the TC more robust.

### DIFF
--- a/ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0111__trusted_ca_bundles.robot
+++ b/ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0111__trusted_ca_bundles.robot
@@ -37,7 +37,13 @@ Validate Trusted CA Bundles State Managed
     ...    Check ConfigMap Contains CA Bundle Key    ${TRUSTED_CA_BUNDLE_CONFIGMAP}    ca-bundle.crt    ${TEST_NS}
 
     Set Custom CA Bundle Value In DSCI    ${DSCI_NAME}   ${CUSTOM_CA_BUNDLE}    ${OPERATOR_NS}
-    Wait Until Keyword Succeeds    10 min    0 sec
+    ${rc}   ${output}=    Run And Return Rc And Output
+    ...    oc get project test-trustedcabundle -o yaml
+    Log To Console          ${output}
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0111__trusted_ca_bundles.robot
```diff
@@ -37,7 +37,13 @@ Validate Trusted CA Bundles State Managed
...
 
     Set Custom CA Bundle Value In DSCI    ${DSCI_NAME}   ${CUSTOM_CA_BUNDLE}    ${OPERATOR_NS}
-    Wait Until Keyword Succeeds    10 min    0 sec
+    ${rc}   ${output}=    Run And Return Rc And Output
+    ...    oc get project test-trustedcabundle -o yaml
+    Log To Console          ${output}
+    ${rc}   ${output}=    Run And Return Rc And Output
+    ...    oc get configmap ${TRUSTED_CA_BUNDLE_CONFIGMAP} -n ${TEST_NS} -o yaml | grep ${CUSTOM_CA_BUNDLE}
+    Log To Console          ${output}
+    Wait Until Keyword Succeeds    5 min    0 sec
```
bdattoma, line 46: these logs are containing dummy values right?

asanzgom: No, what they are ment for is to display the value the failing kw is expecting, but a few moments before the configmap disappears, because it disappears due to some annotation set to false on the namespace yaml and I want to get the snapshot whenever it fails again because it is happening in parallel

MarianMacik (approved):
Makes sense to me.

CFSNM (approved):

ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0111__trusted_ca_bundles.robot
bdattoma: yeah I get it. I mean, the output would contain a dummy CA_BUNDLE?

PR #2386, opened 2025-04-10
[RHOAIENG-12693] Validate DSC Component Names and Versions

asmigala:
The test goes through the prefetched manifests in the operator repo and verifies that each component_metadata.yaml file is reflected in the DSC's status, while skipping over unmanaged components.

### DIFF
--- a/ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
+++ b/ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
@@ -17,6 +17,11 @@ Resource            ../../../../Resources/Common.robot
 Suite Setup         RHOSi Setup
 Suite Teardown      RHOSi Teardown
 
+*** Variables ***
+${RHODS_OPERATOR_GIT_REPO}       %{RHODS_OPERATOR_GIT_REPO=https://github.com/red-hat-data-services/rhods-operator}
+${RHODS_OPERATOR_GIT_DIR}        ${OUTPUT DIR}/rhods-operator
+
+
 *** Test Cases ***
 Verify Dashbord has no message with NO Component Found
     [Tags]  Tier3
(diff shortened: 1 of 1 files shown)


asmigala:
Latest version passed in rhoai-test-flow/3988

ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
```diff
@@ -376,6 +381,37 @@ Verify No Alerts Are Firing After Installation Except For DeadManSnitch    # rob
     # Model Registry not GA yet (Removed state), so its metrics are not enabled by default
     # Verify "Model Registry Operator Probe Success Burn Rate" Alerts Are Not Firing And Continue On Failure
 
+Verify DSC Contains Correct Component Versions  # robocop: disable:too-long-test-case
+    [Tags]    RHOAIENG-12693
+    ...       ExcludeOnODH
```
CFSNM, line 386: wdyt about including both Operator and Smoke tags?

CFSNM (commented):
Some suggestions added

ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
```diff
@@ -376,6 +381,37 @@ Verify No Alerts Are Firing After Installation Except For DeadManSnitch    # rob
     # Model Registry not GA yet (Removed state), so its metrics are not enabled by default
     # Verify "Model Registry Operator Probe Success Burn Rate" Alerts Are Not Firing And Continue On Failure
 
+Verify DSC Contains Correct Component Versions  # robocop: disable:too-long-test-case
```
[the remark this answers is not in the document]
asmigala: AFAICT there's no `component_metadata.yaml` equivalent in ODH, so there's no way for us to know what the expected versions are

ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
CFSNM: Smoke + Operator should be enough

ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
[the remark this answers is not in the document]
CFSNM: okay okay, thats true

ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
```diff
@@ -376,6 +381,37 @@ Verify No Alerts Are Firing After Installation Except For DeadManSnitch    # rob
...
+    [Tags]    RHOAIENG-12693
+    ...       ExcludeOnODH
+    [Documentation]   Verify that component versions are present in DSC status and match the release repo
+    Gather Release Attributes From DSC And DSCI
+    ${rhods_operator_branch} =  Remove String Using Regexp  ${DSC_RELEASE_VERSION}  \\.[0-9]+\$
+    Common.Clone Git Repository  ${RHODS_OPERATOR_GIT_REPO}  rhoai-${rhods_operator_branch}  ${RHODS_OPERATOR_GIT_DIR}
```
asmigala, line 390: One problem with this approach is that if the versions get updated after a downstream build is created but before we run the test, this will fail. Not sure if that could ever happen, as it would mean work on the next z-release started before the current one is tested.

I don't see any z-release tags in the downstream repo, but one way of getting the correct commit would be to get the actual operator image running in the cluster and check its `git.commit` label. Not sure that's necessary though.

WDYT @asanzgom @CFSNM @MarianMacik

CFSNM: the DSC components info and the manifests info should be consistent across all the builds, so I am not sure if we could face this race condition. We can assume atm this situation is not possible and revisit the tests in the next versions just in case we need it

asmigala:
Latest version passed in rhoai-test-flow/4007

CFSNM (approved):

ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
MarianMacik: @asmigala this could happen only when e.g. 2.16.1 is built on the latest nightly build but then, before the next nightly/CI build DevOps update versions to be aligned with 2.16.2. Then the latest build is still from 2.16.1 sources, so the tests will be failing, but that will happen only within a certain window (just a few hours) AND only if versions were updated in z-stream release, which is rarely the case. Z-stream releases should be just about CVE fixes.

ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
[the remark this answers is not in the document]
MarianMacik: The equivalent would be to fetch each ODH repo separately and look there, e.g. for modelmesh-serving: https://github.com/red-hat-data-services/modelmesh-serving/blob/main/config/component_metadata.yaml

But that is too cumbersome I guess.

PR #2401, opened 2025-04-16
Use 4.16 catalog source for installing AMD Certified Operator on yet uncertified OCP versions

bdattoma:
AMD certified their GPU operator on 4.16 and 4.17 only. We were already handling versions < 4.16 but not > 4.18.

Certified versions: https://github.com/redhat-openshift-ecosystem/certified-operators/blob/main/operators/amd-gpu-operator/v1.1.1/metadata/annotations.yaml#L18

UPDATE
v1.2.1 has been released as certified yday https://github.com/redhat-openshift-ecosystem/certified-operators/blob/main/operators/amd-gpu-operator/v1.2.1/metadata/annotations.yaml and it is certified on 4.18.

### DIFF
--- a/ods_ci/tasks/Resources/Provisioning/GPU/AMD/amd_operator.sh
+++ b/ods_ci/tasks/Resources/Provisioning/GPU/AMD/amd_operator.sh
@@ -176,14 +176,16 @@ EOF
   fi 
 }
 
-function applyWorkaroundForOlderOCPVersions () {
-  # workaround for OCP versions less than 4.16
+function applyWorkaroundForUncertifiedOCPVersions () {
+  # workaround for OCP versions less than 4.16 and greater than 4.17
   # AMD certified operator is published starting from OCP v4.16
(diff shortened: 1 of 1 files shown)


ods_ci/tasks/Resources/Provisioning/GPU/AMD/amd_operator.sh
```diff
@@ -196,12 +198,13 @@ spec:
   publisher: RHOAI QE
   sourceType: grpc
 EOF
+    sleep 5s
```
bdattoma, line 201: adding a sleep because I noticed the catalog source pod being created twice: first time it's deleted immediately

bdattoma:
new version (1.2.1) has been just released as certified https://github.com/redhat-openshift-ecosystem/certified-operators/blob/main/operators/amd-gpu-operator/v1.2.1/metadata/annotations.yaml

apodhrad (approved):

PR #2420, opened 2025-04-29
fix: add automationbug tag to a monitoring case that takes 1h to finish

CFSNM:


ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
```diff
@@ -80,6 +80,7 @@ Verify Alerts Are Fired When RHODS Dashboard Is Down    # robocop: disable:too-l
     [Tags]    Tier3
     ...       ODS-739
     ...       Monitoring
+    ...       AutomationBug
```
MarianMacik, line 83: I'm currently working on that to implement a keyword which will pass if the alert transitions to state pending. Because each alert has duration specified and it starts firing only after being pending for that time duration, we should check only pending state, which is triggered immediately. Waiting 1 hour for alert to transition from pending to firing does not make sense.
Timer duration is already checked when checking that alerts are present, so that is covered.

CFSNM: thank u Marian!

PR #2425, opened 2025-05-02
RHOAIENG-12886 - ODS-2170: Verify Data Science Pipelines Application …

MarianMacik:
…Alerts

**JIRA:** [RHOAIENG-12886 - ODS-2170: Verify Data Science Pipelines Application Alerts](https://issues.redhat.com/browse/RHOAIENG-12886)

I revised the test case a bit as opposed to how it was defined in Polarion. In the end I omitted these steps because of a low added value and unnecessarily prolonging the test (which already takes 8 and a half minutes):
1. "Verify alerts are firing also in Alertmanager" - I implemented this first but as it is a part of Prometheus, I then realized that it would be testing Prometheus and Alertmanager communication, not RHOAI. Also, Alertmanager config is already tested by other tests.
2. "Verify that the alert can be seen in OpenShift Cluster Monitoring Prometheus" - RHOAI Prometheus already consumes OpenShift Cluster Prometheus metrics via `/federate` endpoint, so we are effectively testing that.
3. "(Optionally) verify that the alert can be seen in Telemeter" - This would be testing that a managed cluster is able to send alerts to Red Hat Console, so again not testing RHOAI.

Test runs:
* Single test: /job/rhoai/job/2.20/job/managed/job/cli/job/aws/job/rhoai-tier2/3/
* All Monitoring test cases (failures are expected as those test cases are not stabilized yet): /job/rhoai/job/2.20/job/managed/job/cli/job/aws/job/rhoai-tier2/4/

### DIFF
--- a/ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
+++ b/ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
@@ -6,6 +6,7 @@ Resource            ../../../../Resources/ODS.robot
 Resource            ../../../../Resources/Common.robot
 Resource            ../../../../Resources/Page/OCPDashboard/Builds/Builds.robot
 Resource            ../../../../Resources/Page/ODH/JupyterHub/HighAvailability.robot
+Resource            ../../../../Resources/CLI/DataSciencePipelines/DataSciencePipelinesBackend.resource
 Library             OperatingSystem
 Library             SeleniumLibrary
 Library             JupyterLibrary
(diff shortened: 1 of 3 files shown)


asmigala (commented):
@MarianMacik  LGTM, just two concerns:

* the comment about metrics returing NaN, is that a bug?
* We now expect some alerts to be pending and other firing, this is a timing issue, would it make sense to create a keyword `Wait Until Alert Is Pending Or Firing`, so that if one alerts actually starts firing while we are waiting for another one to go into pending, the tests would still pass?

asanzgom (approved):

MarianMacik:
@asmigala 

    

> the comment about metrics returing NaN, is that a bug?

It's not a bug because the metric is written in this way: number of requests returning non-OK status / number of total requests. And if there are no requests made, the denominator is 0 -> division by zero.
    

> We now expect some alerts to be pending and other firing, this is a timing issue, would it make sense to create a keyword Wait Until Alert Is Pending Or Firing, so that if one alerts actually starts firing while we are waiting for another one to go into pending, the tests would still pass?

The first 4 alerts have rules which look 3 minutes in the past and if there is at least one success (max value is 1), it won't start firing. So the test has to wait 3 minutes for these rules to make alerts pending and alert would start firing after additional 2 minutes, which I wanted to avoid.

The burn rate alert will start pending immediately but when I was testing it I noticed that by waiting for it to firing we have to wait only 20-30 seconds max, so I decided to check firing state already.

So pending and alerting expectations were a trade-off between test duration and coverage let's say.

I can change that comment so it does not suggest it is a timing issue.

CFSNM (approved):
LGTM!

ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
```diff
@@ -199,6 +200,139 @@ Verify That MT-SRE Are Not Paged For Alerts In Clusters Used For Development Or
...
+    ...                                    Data Science Pipeline ScheduledWorkflows Unavailable
+
+    FOR    ${alert}    IN    @{alerts_to_check}
+            Prometheus.Wait Until Alert Is Pending    ${RHODS_PROMETHEUS_URL}
+            ...    ${RHODS_PROMETHEUS_TOKEN}
+            ...    RHODS Data Science Pipelines
+            ...    ${alert}
+            ...    alert-duration=120
+            ...    timeout=5 min
+    END
```
asmigala, line 284: @MarianMacik 

> The first 4 alerts have rules which look 3 minutes in the past and if there is at least one success (max value is 1), it won't start firing. So the test has to wait 3 minutes for these rules to make alerts pending and alert would start firing after additional 2 minutes, which I wanted to avoid.

this is what I meant - here we wait up to 5 minutes for each of the metrics to go into `Pending`, but it's possible that during waiting for the first one, the later ones will go from `Pending` into `Firing` and the test will fail, no?

What I'm saying is that we don't actually care whether the alert is pending or firing, they are equivalent for the purpose of this test, so it might be safer to check that the alert is in either of the states.

asmigala (approved):

PR #2427, opened 2025-05-06
Disable Update notification email test

asmigala:
It is currently failing because we are faking the addon installation, which prevents ocm from updating the email


MarianMacik (approved):

sefroberg (approved):
/LGTM

PR #2429, opened 2025-05-07
Update Get Operator Starting Version method to use global variable

ChughShilpa:


ChughShilpa:
Verified in `rhoai-test-flow/4398`

ods_ci/tests/Resources/Common.robot
```diff
@@ -589,7 +589,7 @@ Clone Git Repository
 Get Operator Starting Version
     [Documentation]    Returns the starting version of the operator in the upgrade chain
     ${rc}    ${out}=    Run And Return RC And Output
-    ...    oc get subscription rhods-operator -n ${OPERATOR_NAMESPACE} -o yaml | yq -r '.spec.startingCSV' | awk -F. '{print $2"."$3"."$4}'    # robocop: disable
+    ...    oc get subscription ${OPERATOR_SUBSCRIPTION_NAME} -n ${OPERATOR_NAMESPACE} -o yaml | yq -r '.status.installedCSV' | awk -F. '{print $2"."$3"."$4}'    # robocop: disable
```
sutaakar, line 592: If I get it right the `installedCSV` represents current CSV version installed in the cluster, right?
In this case we need to check for operator version which was originally installed, not current one.

ChughShilpa: `installedCSV` and `currentCSV` are two different parameters defined in subscription.yaml @sutaakar 
```
tatus:
  installplan:
    apiVersion: operators.coreos.com/v1alpha1
    kind: InstallPlan
    name: install-bn5kg
    uuid: cffecbaf-74df-48a2-a345-c428f865903c
  lastUpdated: '2025-05-05T11:05:07Z'
  installedCSV: rhods-operator.2.20.0
  currentCSV: rhods-operator.2.20.0
 ```

Comment (bdattoma) [reply ods_ci/tests/Resources/Common.robot:592]: correct, installedCSV cannot be used in this keyword.

We need to find a way to set the installedCSV  in the subscription at install time, more likely it will happen after the install once the code can check the installed version.

An alternative is to read the installed version in pre-upgrade and store it in a configmap which can be read by the post-upgrade suite

Comment (ChughShilpa) [reply ods_ci/tests/Resources/Common.robot:592]: There is another way to find the intial installed version by checkinh the earliest clusterserviceversion
`oc get csv -n redhat-ods-operator --sort-by=.metadata.creationTimestamp`
```
NAME                                     DISPLAY                                   VERSION         REPLACES                            PHASE
rhods-operator.2.20.0         Red Hat OpenShift AI                2.20.0             rhods-operator.2.19.0         Succeeded
```
@sutaakar @bdattoma

Comment (sutaakar) [reply ods_ci/tests/Resources/Common.robot:592]: Does cluster keep older CSVs?

Comment (ChughShilpa) [reply ods_ci/tests/Resources/Common.robot:592]: I think no, OLM keeps only active (installed) CSVs. Currently, `oc get csv` shows only `rhods-operator.2.20.0` 

I suggested this as in the cutrent CSV output, it shows the replaced (old) version
```
NAME                                      DISPLAY                           VERSION   REPLACES                                  PHASE
rhods-operator.2.20.0          Red Hat OpenShift AI        2.20.0        rhods-operator.2.19.0              Succeeded
```

bdattoma: https://github.com/red-hat-data-services/ods-ci/pull/2439 testing a solution here

PR #2434, opened 2025-05-07
post_install: Remove Tier1 and Monitoring tags from DeadManSnitch sce…

MarianMacik:
…nario

Monitoring tag was left commented as this test is a monitoring test case still but it is important to run this with every Smoke automatically rather than with Monitoring test suite.


sefroberg (approved):
/LGTM

asanzgom (approved):

CFSNM (approved):

PR #2440, opened 2025-05-12
Remove PyTorch 2.4.1 upgrade tests for Training component

sutaakar:
These tests use same namespace as PyTorch 2.5.1 upgrade tests, causing concurrency issues.
We don't need to run upgrade tests using different runtime images as they aren't relevant for upgrade functionality. It can be verified with one image only.


ChughShilpa (approved):
Thanks Karel

abhijeet-dhumal (approved):

PR #2441, opened 2025-05-12
[2.20] Remove PyTorch 2.4.1 upgrade tests for Training component

sutaakar:
These tests use same namespace as PyTorch 2.5.1 upgrade tests, causing concurrency issues.
We don't need to run upgrade tests using different runtime images as they aren't relevant for upgrade functionality. It can be verified with one image only.


sutaakar:
Cherrypick of https://github.com/red-hat-data-services/ods-ci/pull/2440

ChughShilpa (approved):

abhijeet-dhumal (approved):

PR #2448, opened 2025-05-15
[RHOAIENG-13262] Verify Data Science Pipelines Operator Alert

asmigala:
Successfull run in rhoai/2.20/managed/cli/aws/rhoai-tier2/7/

### DIFF
--- a/ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
+++ b/ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
@@ -82,10 +82,6 @@ Verify Alerts Are Fired When RHODS Dashboard Is Down    # robocop: disable:too-l
     ...       ODS-739
     ...       Monitoring
     ...       AutomationBug
-    Skip If Alert Is Already Firing    ${RHODS_PROMETHEUS_URL}
-    ...    ${RHODS_PROMETHEUS_TOKEN}
-    ...    SLOs-haproxy_backend_http_responses_dashboard
-    ...    RHODS Dashboard Route Error Burn Rate
 
     ODS.Scale Deployment    ${OPERATOR_NAMESPACE}    rhods-operator    replicas=0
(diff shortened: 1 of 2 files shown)


CFSNM (approved):

ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
```diff
@@ -334,6 +334,37 @@ Verify Data Science Pipelines Application Alerts
...
     [Teardown]    Delete Project Via CLI    ${PROJECT}
 
+
+Verify Data Science Pipelines Operator Alert Fires When Operator Is Down
+    [Documentation]     Verifies that alert "Data Science Pipelines Operator Probe Success Burn Rate (for 5m)" is fired
+    [Tags]    Tier3
+    ...       ODS-2166
+    ...       RHOAIENG-13262
+    ...       Monitoring
+    Skip If Alert Is Already Firing    ${RHODS_PROMETHEUS_URL}
```
MarianMacik, line 344: I know we already have this in other test cases but do we want to skip this gracefully if an alert is already firing? I would let it run and possibly fail so we know the cluster wasn't clean/RHOAI wasn't correctly installed.

MarianMacik (commented):
Just one comment to consider, otherwise looks good.

ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
asmigala: Good point, removed here. Should I also remove it from the other tests?

CFSNM (dismissed):

ods_ci/tests/Tests/0100__platform/0102__monitor_and_manage/0102__alerts/0102__alerts.robot
MarianMacik: I would say so. Thanks!

MarianMacik (approved):
Looks good to me now!

CFSNM (approved):

PR #2449, opened 2025-05-16
[RHOAIENG-12811] Verify allowlist for component customization

asmigala:
Passing test run in rhoai-test-flow/5186

### DIFF
--- a/ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0113__dsc_components.robot
+++ b/ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0113__dsc_components.robot
@@ -58,10 +58,18 @@ ${IS_NOT_PRESENT}                                       1
 ...                                                     WORKBENCHES=${EMPTY}
 ...                                                     FEASTOPERATOR=${EMPTY}
 
-@{CONTROLLERS_LIST}
-...    kserve-controller-manager
-...    odh-model-controller
-...    modelmesh-controller
+@{CONTROLLERS_LIST}                                     # dashboard added in Suite Setup, since it's different in RHOAI vs ODH
+...                                                     codeflare-operator-manager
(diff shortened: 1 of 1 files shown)


ods_ci/tests/Tests/0100__platform/0101__deploy/0104__operators/0104__rhods_operator/0113__dsc_components.robot
```diff
@@ -301,28 +313,52 @@ Validate Feastoperator Removed State
...
+        ${new_image}=  Set Variable  registry.invalid/test:latest
+        ${new_replicas}  Set Variable  0
+
+        # overwrite some fields
+        Patch Controller Deployment  ${controller}  /spec/template/spec/containers/0/resources/limits/cpu  '${new_cpu_limit}'
+        Patch Controller Deployment  ${controller}  /spec/template/spec/containers/0/resources/limits/memory  '${new_memory_limit}'
+        Patch Controller Deployment  ${controller}  /spec/template/spec/containers/0/image  '${new_image}'
+        Patch Controller Deployment  ${controller}  /spec/replicas  ${new_replicas}
+
+        Sleep    10s  Give time for operator to potentially reconcile our changes
```
sefroberg, line 329: Is this enough time? You have a 1 min 10 seconds wait for the reverted portion but only 10 seconds here.

asmigala: Good point, what I'm seeing is that the non-allowlisted values are reverted immediately, but it's true that's just the current implementation and it might change in the future (though probably unlikely). I originally also had 1 min wait here, the problem with that is that we iterate over all the controllers, so this test would take ~15 minutes.

One other option I considered is doing all the patches at once, then just a single wait, then verify if things got reverted. That makes the code somewhat ugly and less clear, but it's probably the safer approach, wdyt?

asanzgom (approved):

asanzgom (approved):

PR #2455, opened 2025-05-20
Remove odhdashboardconfig checks from the Verify Auth CRD Test Case

asanzgom:
Remove odhdashboardconfig checks from the Verify Auth CRD Test Case, as the logic of copying user groups from the odhdashboardconfig into the auth CR has been deprecated

### DIFF
--- a/ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource
+++ b/ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource
@@ -23,11 +23,11 @@ ${FMS_HF_TUNING_IMAGE}                   quay.io/modh/fms-hf-tuning@sha256:1ad46
 # Corresponds to quay.io/modh/training:py311-cuda121-torch241
 ${CUDA_TRAINING_IMAGE_TORCH241}          quay.io/modh/training@sha256:f64f7bba3f1020d39491ac84d40d362a52e4822bdc11a33cfff021178b7c4097
 # Corresponds to quay.io/modh/training:py311-rocm62-torch241
-${ROCM_TRAINING_IMAGE_TORCH241}          quay.io/modh/training@sha256:1104eb83d7ad9cfb63c1437be38d1615a669473c5b1c2da5a9a01ddd90daceb9
+${ROCM_TRAINING_IMAGE_TORCH241}          quay.io/modh/training@sha256:883739b576b485d79966d8c894fdd9ebebd226605a2abe8b33593ca67c87a394
 # Corresponds to quay.io/modh/training:py311-cuda124-torch251
 ${CUDA_TRAINING_IMAGE_TORCH251}          quay.io/modh/training@sha256:1d0caea3e5d56ff7d672954b1ad511e661df9bdb364d56879961169a4ca8dae0
 # Corresponds to quay.io/modh/training:py311-rocm62-torch251
-${ROCM_TRAINING_IMAGE_TORCH251}          quay.io/modh/training@sha256:f98d70edc49203537d4fec005aaff68a4830b9c9ce2390bb4d7910b89c312b97
+${ROCM_TRAINING_IMAGE_TORCH251}          quay.io/modh/training@sha256:6cdae840fa029da33cccab620367e82404d24ddf67762eb4537a9bffe1af306d
 # Corresponds to quay.io/modh/odh-generic-data-science-notebook:v3-20250519
 ${NOTEBOOK_IMAGE_3.11}                   quay.io/modh/odh-generic-data-science-notebook@sha256:1c585c849cc466fce6443f5c186fb9b642d21b5f3581515ada08bba2fbc3591f
 ${NOTEBOOK_USER_NAME}                    ${TEST_USER_3.USERNAME}


ods_ci/tests/Tests/0100__platform/0101__deploy/0102__auth_providers_and_rbac/0103__auth_crd.robot
```diff
@@ -15,18 +15,10 @@ Verify Auth Crd
     [Tags]      Smoke
     ...         Operator
     ...         RHOAIENG-18846
-    ${rc}       ${odhdashboardconfig_groups}=    Run And Return Rc And Output
-    ...    oc get odhdashboardconfig odh-dashboard-config -n ${APPLICATIONS_NAMESPACE} -o jsonpath='{.spec.groupsConfig}'
-    Should Be Equal As Integers     ${rc}       0
     ${rc}       ${auth_cr_groups}=      Run And Return Rc And Output
     ...                     oc get auth auth -o jsonpath='{.spec}'
     Should Be Equal As Integers     ${rc}       0
     ${auth_cr_groups_str}=      Remove String       ${auth_cr_groups}   [   ]
```
CFSNM, line 21: this variable is no longer used, so we can delete it

ods_ci/tests/Tests/0100__platform/0101__deploy/0102__auth_providers_and_rbac/0103__auth_crd.robot
```diff
@@ -15,18 +15,10 @@ Verify Auth Crd
     [Tags]      Smoke
     ...         Operator
     ...         RHOAIENG-18846
-    ${rc}       ${odhdashboardconfig_groups}=    Run And Return Rc And Output
-    ...    oc get odhdashboardconfig odh-dashboard-config -n ${APPLICATIONS_NAMESPACE} -o jsonpath='{.spec.groupsConfig}'
-    Should Be Equal As Integers     ${rc}       0
     ${rc}       ${auth_cr_groups}=      Run And Return Rc And Output
```
CFSNM, line 18: this variable is no longer used, so we can delete it

CFSNM (changes requested):
some suggestions

sefroberg (dismissed):
/LGTM

ods_ci/tests/Tests/0100__platform/0101__deploy/0102__auth_providers_and_rbac/0103__auth_crd.robot
asanzgom: removed

ods_ci/tests/Tests/0100__platform/0101__deploy/0102__auth_providers_and_rbac/0103__auth_crd.robot
asanzgom: removed

CFSNM (approved):

ods_ci/tests/Tests/0100__platform/0101__deploy/0102__auth_providers_and_rbac/0103__auth_crd.robot
```diff
@@ -15,18 +15,6 @@ Verify Auth Crd
...
-    Should Be Equal As Integers     ${rc}       0
-    ${auth_cr_groups_str}=      Remove String       ${auth_cr_groups}   [   ]
-    IF    """${odhdashboardconfig_groups}""" != """${auth_cr_groups_str}"""
-        Fail        User Groups on Auth Cr and OdhDashboardConfig DO NOT match
-    ELSE
-        Log To Console      User Groups on Auth Cr and OdhDashboardConfig match
-    END
     Get User Groups From Auth Cr And Check Rolebinding Exists       adminGroups     rolebinding     admingroup-rolebinding
     Get User Groups From Auth Cr And Check Rolebinding Exists       adminGroups     clusterrolebinding     admingroupcluster-rolebinding
     Get User Groups From Auth Cr And Check Rolebinding Exists       allowedGroups   rolebinding     allowedgroup-rolebinding
```
MarianMacik, line 20: In case we are still checking these, it would be good to have the Auth CR content as well as these rolebindings should exist only if the respective groups are configured in the Auth CR.

asanzgom: Do you want me to discard the last commit then?

MarianMacik: I think just the dashboardconfig check should be removed. And we should check the content of Auth CR that it contains `{"adminGroups":"rhods-admins","allowedGroups":"system:authenticated"}` which is the default.

asanzgom: I will include that

asanzgom: Wait, I think that check is already being done on the Get User Groups From Auth Cr And Check Rolebinding Exists Keyword

MarianMacik: OK, makes sense. We could check the expected default values because now we don't check the content, just that whatever is in Auth CR is referenced in these rolebindings.
However, that will fail post-upgrade smokes I think because during upgrade testing we change the defaults and then when we run smokes after it would be failing as there will be other values than defaults in Auth CR. So I think it is fine as of now probably.

bdattoma: >However, that will fail post-upgrade smokes I think because during upgrade testing we change the defaults and then when we run smokes after it would be failing as there will be other values than defaults in Auth CR.


post upgrade suite teardown should restore the default values, shouldn't it?

bdattoma (approved):

ods_ci/tests/Tests/0100__platform/0101__deploy/0102__auth_providers_and_rbac/0103__auth_crd.robot
MarianMacik: @bdattoma @asanzgom yep, it should, but as of now it doesn't since we ran into issues with that. That when Smokes were run after upgrade test suite, it was failing on this.
BUT now when I am looking at it, it is probably restored:
https://github.com/red-hat-data-services/ods-ci/blob/561872afffc24caaa145987d6fa80fe6a5ea74a8/ods_ci/tests/Tests/0200__rhoai_upgrade/0203__post_upgrade.robot#L353-L352
Which [patches OdhDashboardConfig](https://github.com/red-hat-data-services/ods-ci/blob/561872afffc24caaa145987d6fa80fe6a5ea74a8/ods_ci/tests/Resources/ODS.robot#L104) back. However, now it should be patching Auth CR probably, wdyt @asanzgom ?

bdattoma: indeed, it is using https://github.com/red-hat-data-services/ods-ci/blob/1c7713fce60cb5fe488df2e97447e5f0218bc466/ods_ci/tests/Resources/ODS.robot#L105 which patches the OdhDashboardConfig

PR #2465, opened 2025-05-26
RHOAIENG-26136: Automatically clean dependent operators

apodhrad:


ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
```diff
@@ -51,7 +51,7 @@ RHODS Operator Should Be Uninstalled
 Uninstall RHODS In Self Managed Cluster Using CLI
   [Documentation]   UnInstall rhods on self-managedcluster using cli
   Clone OLM Install Repo
-  ${return_code}    Run and Watch Command    cd ${EXECDIR}/${OLM_DIR} && ./cleanup.sh -t operator    timeout=10 min
+  ${return_code}    Run and Watch Command    cd ${EXECDIR}/${OLM_DIR} && ./cleanup.sh -t operator -a "authorino serverless servicemesh"    timeout=10 min
```
mwaykole, line 54: will it also clean crd ?

CFSNM, line 54: in the line 33 we have the same option for the addon for managed clusters. Do we want to add the -a option there too?

CFSNM (changes requested):

ods_ci/tasks/Resources/RHODS_OLM/uninstall/uninstall.robot
jgarciao: Looking at the changes at olminstall it seems that yes, the crd's are deleted

apodhrad: @CFSNM not at the moment because
* this change is primarily for supporting kserve raw deployments which is for self-managed
* addons are for managed clusters which we do not reuse so often
* I did not test it with an addon (but it should be working in the same way)

So, there is such a requirement then I will test it and extend also the addon cleanup.
But I would rather do it as a separate task and separate PR

PR #2484, opened 2025-06-04
RHOAIENG-13073 - ODS-558: Verify alerts have links to the triage guide

MarianMacik:
**JIRA:** [RHOAIENG-13073](https://issues.redhat.com/browse/RHOAIENG-13073)

This will fail if there is a new alert added which does not have triage URL mapped, also if triage URL is missing (except for DeadManSnitch). I used regex for mapping because a lot of alerts have the same triage URL, to save space and time.

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/Prometheus/Triage.resource
+++ b/ods_ci/tests/Resources/Page/ODH/Prometheus/Triage.resource
@@ -1,0 +1,45 @@
+*** Settings ***
+Documentation       Contains Triage URLs Mappings
+
+
+*** Variables ***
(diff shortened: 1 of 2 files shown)


asanzgom:
@MarianMacik Can you provide a successful run

CFSNM (approved):
/lgtm

MarianMacik:
@asanzgom I had to reprovision the cluster as it was deleted already.

Here is the run: rhoai-test-flow/4846
It currently fails because one triage link is not correct, it is fixed by https://github.com/opendatahub-io/opendatahub-operator/pull/2011

PR #2487, opened 2025-06-05
fix: Remove wait for NFD in nvidia ns

bdattoma:
https://github.com/red-hat-data-services/ods-ci/pull/2477 was not enough.

PR validation:
- `rhoai-test-flow/4868` PASS


amalykhi (approved):
Passed test run for me

PR #2483, opened 2025-06-03
create different upgrade configmap for user groups to avoid collisions

asanzgom:
create different upgrade configmap for user groups to avoid collisions

PR #2502, opened 2025-06-12
[2.19] create different upgrade configmap for user groups to avoid collisions

jstourac:
(cherry picked from commit f59e1020980720ff85f39af89a4bdf05dd1afb81)

This is a backport of #2483


CFSNM (approved):

MarianMacik (approved):

MarianMacik:
@jstourac post_upgrade file needs to be changed as well, look at the original PR.

jstourac:
> @jstourac post_upgrade file needs to be changed as well, look at the original PR. I think it makes to change the post_upgrade as well as there may be upgrades to patches of 2.19 because 2.19 is stable.

If I'm not missing anything, then - If I change the post_upgrade file, then I would have to backport this even further into the past. Because when you do upgrade testing, then `pre_upgrade` is used from branch X but `post_upgrade` is used from version X+Y.

I don't want to do backport these changes more into the past myself, but I am okay if somebody else will do so.

PR #2510, opened 2025-06-23
Add user defined tags or label for cloud resources when provisioning self-managed AWS, GCP and Azure clusters

bdattoma:


kobihk (approved):

aloganat (approved):
lgtm

PR #2547, opened 2025-07-15
RHOAIENG-26686 - [QE] Enable the Monitoring tests to run in the conne…

MarianMacik:
…cted scenarios as well

**JIRA:** [RHOAIENG-26686](https://issues.redhat.com/browse/RHOAIENG-26686)

Preparation for when the source will be merged in RHOAI, so creating as a draft. We still need to discuss some things to better understand which parts will be still applicable only to managed clusters, hence the comments.

### DIFF
--- a/ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
+++ b/ods_ci/tests/Tests/0100__platform/0101__deploy/0101__installation/0101__post_install.robot
@@ -86,7 +86,7 @@ Verify That Prometheus Image Is A CPaaS Built Image
     [Tags]    Sanity
     ...       Tier1
     ...       ODS-734
-    Skip If RHODS Is Self-Managed
+    Skip If RHODS Is Self-Managed    # TODO: Do we still control Prometheus image to deploy?
     Wait For Pods To Be Ready    label_selector=deployment=prometheus
     ...    namespace=${MONITORING_NAMESPACE}    timeout=60s
     ${pod} =    Find First Pod By Name    namespace=${MONITORING_NAMESPACE}    pod_regex=prometheus-
@@ -100,7 +100,7 @@ Verify That Blackbox-exporter Image Is A CPaaS Built Image
     [Tags]    Sanity
     ...       Tier1
     ...       ODS-735
-    Skip If RHODS Is Self-Managed
+    Skip If RHODS Is Self-Managed    # TODO: Will we deploy blackbox-exporter also on self-managed?
     Wait For Pods To Be Ready    label_selector=deployment=blackbox-exporter
     ...    namespace=${MONITORING_NAMESPACE}    timeout=60s
     ${pod} =    Find First Pod By Name    namespace=${MONITORING_NAMESPACE}    pod_regex=blackbox-exporter-
@@ -112,7 +112,7 @@ Verify That Alert Manager Image Is A CPaaS Built Image
     [Tags]    Sanity
     ...       Tier1
     ...       ODS-733
-    Skip If RHODS Is Self-Managed
+    Skip If RHODS Is Self-Managed    # TODO: Will we deploy alertmanager also on self-managed?
     Wait For Pods To Be Ready    label_selector=deployment=prometheus
     ...    namespace=${MONITORING_NAMESPACE}    timeout=60s
     ${pod} =    Find First Pod By Name    namespace=${MONITORING_NAMESPACE}    pod_regex=prometheus-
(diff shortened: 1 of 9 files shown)


CFSNM:
I wonder if we also need to modify the logic to retrieve the Prometheus URL and auth params. Need to check if that needs to be done in ods-ci or in jenkins repo

ods_ci/tests/Resources/Common.robot
```diff
@@ -615,3 +626,12 @@ Delete All ${resource_type} In Namespace By Name
...
     ...    shell=true    stderr=STDOUT
     Log    ${result.stdout}    console=yes
+
+Is New Observability Stack Enabled
+    [Documentation]    Checks whether new observability stack is enabled, controlled by ENABLE_NEW_OBSERVABILITY_STACK test variable.
+    ...    Returns ${TRUE} if new observability stack is enabled by the variable, i.e. ENABLE_NEW_OBSERVABILITY_STACK
+    ...    has a value which can be converted to boolean True (True|true|"True"|"true")
+    ...    Returns ${FALSE} otherwise, i.e. the variable is not defined or has a false value
+    ${enable_new_observability_stack} =    Get Variable Value    ${ENABLE_NEW_OBSERVABILITY_STACK}    false
+    ${enable_new_observability_stack} =    Convert To Boolean    ${enable_new_observability_stack}
```
CFSNM, line 636: @MarianMacik did you face any issue with variable type? I dont remember seeing something like that when I tested this code change

ods_ci/tests/Resources/Common.robot
```diff
@@ -314,6 +314,17 @@ Skip If RHODS Is Managed
        Skip If    condition=${is_self_managed}==False    msg=This test is skipped for Managed RHODS
     END
 
+Skip If RHODS Is Self-Managed And New Observability Stack Is Disabled
+    [Documentation]    Skips test if RHODS is installed as Self-managed and ${ENABLE_NEW_OBSERVABILITY_STACK} == false
+    [Arguments]    ${msg}=${EMPTY}
+    ${is_self_managed}=    Is RHODS Self-Managed
+    ${enable_new_observability_stack} =    Is New Observability Stack Enabled   # Replace with this also other occurrences
```
asmigala, line 321: What does the comment mean? Is this a TODO that needs to be done once the rest of the functionality is implemented? I see this being used everywhere that I can tell

ods_ci/tests/Resources/Page/ODH/Prometheus/Prometheus.robot
```diff
@@ -382,23 +382,37 @@ Run Query Range
...
         IF    "${RHODS_PROMETHEUS_URL}" == "" or "${RHODS_PROMETHEUS_URL}" == "PROMETHEUS_URL"
-            ${result} 	Run Process 	oc get route prometheus -n ${prom_namespace} -o jsonpath\='{.spec.host}'    shell=yes
-            Should Be True    ${result.rc} == 0    Error getting Prometheus host: ${result.stderr}
-            ${RHODS_PROMETHEUS_URL}=    Set Variable    https://${result.stdout}
-            Log    Prometheus URL: ${RHODS_PROMETHEUS_URL}    console=True
+            ${RHODS_PROMETHEUS_URL}=    Set Variable    ${protocol}://${prometheus_url}
+            Log    Setting RHODS_PROMETHEUS_URL to: ${RHODS_PROMETHEUS_URL}    console=True
             Set Global Variable   ${RHODS_PROMETHEUS_URL}
+        ELSE
+            Log    Skipping setting RHODS_PROMETHEUS_URL, it already has value: ${RHODS_PROMETHEUS_URL}    console=True
```
asmigala, line 405: What would be the use case for setting the prometheus url to something other than the auto-detected value?

MarianMacik: That is a good question, but probably for Noam. I just wanted to leave the functionality unchanged. In theory you may deploy your own Prometheus for testing etc. But AFAIK we don't do that. I just left this option to preserve backward functionality.

ods_ci/tests/Resources/Common.robot
MarianMacik: Unfortunately yes. As I described in the PR description: 

> The reason was that when the variable is provided through the test-variables.yml file it is parsed as Boolean, while when it is provided via command line, it is parsed as String.

When it is read from `test-variables.yml` its type is `Boolean` and it looks like `False`. When it is provided using command line like `ENABLE_NEW_OBSERVABILITY_STACK:false`, which we use in installation scenarios, then its type is `String` and it looks like `false`.
The keyword I added works with both options.

MarianMacik:
> the changes look good to me. I am now wondering if we should start running the Monitoring tests using the new stack as well, as we are running them using the old stack

@CFSNM I would still wait a bit until we have scraping of metrics and alerts in place. Now there are no metrics to scrape yet. And no alerts are defined too.

asanzgom (approved):

PR #2583, opened 2025-07-30
add logic to enable disable the monitoring stuff in the dsci when the dsci is created by the operator

CFSNM:


ods_ci/tasks/Resources/Files/monitoring-patch-payload.json
```diff
@@ -1,7 +1,7 @@
 {
     "spec": {
         "monitoring": {
-            "managementState": "Managed",
+            "managementState": "<monitoring_namespace>",
```
asmigala, line 4: Should be `<monitoring_value>`

asmigala (changes requested):

ods_ci/tasks/Resources/Files/monitoring-patch-payload.json
CFSNM: changed, copy paste issue :(

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -372,6 +372,18 @@ Create DSCInitialization CustomResource Using Test Variables
 Patch DSCInitialization With Monitoring Info
     [Documentation]  Patches the DSCInitialization with the Monitoring info in case is created by the operator
     ${file_path} =    Set Variable    tasks/Resources/Files/
+    ${ENABLE_NEW_OBSERVABILITY_STACK} =    Get Variable Value    ${ENABLE_NEW_OBSERVABILITY_STACK}    true
+        IF    "${ENABLE_NEW_OBSERVABILITY_STACK}" == "true"
```
asmigala, line 376: AFAICT we only call `Patch DSCInitialization With Monitoring Info` on line 333 which is already inside of `IF "${enable_new_observability_stack}" == "true"`, so the whole `ELSE` branch here is dead code, no?

CFSNM: true, fixed that

asmigala (approved):

PR #2588, opened 2025-07-31
enhancement: simplifying the logic behing the new observability stack

CFSNM:
…


asanzgom (dismissed):

asmigala (dismissed):

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -327,7 +327,6 @@ Apply DSCInitialization CustomResource
     Log To Console    output : ${output}, return_code : ${return_code}
     IF  ${output} != 0
         Log to Console    Skip creation of DSCInitialization
-        Patch DSCInitialization With Monitoring Info
```
MarianMacik, line 330: In this case, if DSCI is already created by the operator and user selected ENABLE_NEW_OBSERVABILITY_STACK=true, it won't be patched.

I think the flow should be like this:
If DSCI is already created and if ENABLE_NEW_OBSERVABILITY_STACK=true -> patch it

If DSCI is not created yet -> create DSCI
     If ENABLE_NEW_OBSERVABILITY_STACK=true -> patch it

ods_ci/tasks/Resources/Files/dsci_template.yml
```diff
@@ -11,21 +11,6 @@ metadata:
...
-                memoryrequest: 256Mi
-            storage:
-                retention: 1d
-                size: 5Gi
-        traces:
-            sampleRatio: "0.1"
-            storage:
-                backend: "pv"
-                size: "5Gi"
         managementState: <monitoring_value>
```
MarianMacik, line 14: We can leave that Managed at all times like [it used to be](https://github.com/red-hat-data-services/ods-ci/blob/release-2.22/ods_ci/tasks/Resources/Files/dsci_template.yml#L14) because this is [the default](https://github.com/opendatahub-io/opendatahub-operator/blob/be9b0ccabc00cdffc72dac15d5feb421ec555c85/pkg/upgrade/upgrade.go#L227-L226) for both selfmanaged and managed. But just on managed cluster the monitoring will actually be deployed.

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -353,36 +353,27 @@ Create DSCInitialization CustomResource Using Test Variables
     Run    sed -i'' -e 's/<monitoring_namespace>/${MONITORING_NAMESPACE}/' ${file_path}dsci_apply.yml
     Run    sed -i'' -e 's/<operator_yaml_label>/${OPERATOR_YAML_LABEL}/' ${file_path}dsci_apply.yml
     ${ENABLE_NEW_OBSERVABILITY_STACK} =    Get Variable Value    ${ENABLE_NEW_OBSERVABILITY_STACK}    true
-        IF    "${ENABLE_NEW_OBSERVABILITY_STACK}" == "true"
-                Run    sed -i'' -e 's/<monitoring_value>/Managed/' ${file_path}dsci_apply.yml
+    IF    "${ENABLE_NEW_OBSERVABILITY_STACK}" == "true"
+            Run    sed -i'' -e 's/<monitoring_value>/Managed/' ${file_path}dsci_apply.yml
+    ELSE
+        IF     "${cluster_type}" == "selfmanaged"
```
MarianMacik, line 359: We don't need these changes if we let managementState Managed for monitoring at all times.

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
CFSNM: done

ods_ci/tasks/Resources/Files/dsci_template.yml
CFSNM: done

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
CFSNM: done

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -352,37 +359,16 @@ Create DSCInitialization CustomResource Using Test Variables
...
-                    Run    sed -i'' -e 's/<monitoring_value>/Managed/' ${file_path}dsci_apply.yml
-            END
-        END
 
 Patch DSCInitialization With Monitoring Info
-    [Documentation]  Patches the DSCInitialization with the Monitoring info in case is created by the operator
+    [Documentation]  Patches the DSCInitialization with the Monitoring info if the new obs stack is being used
     ${file_path} =    Set Variable    tasks/Resources/Files/
     ${ENABLE_NEW_OBSERVABILITY_STACK} =    Get Variable Value    ${ENABLE_NEW_OBSERVABILITY_STACK}    true
         IF    "${ENABLE_NEW_OBSERVABILITY_STACK}" == "true"
```
MarianMacik, line 367: This check is already done before the keyword is called, we can either remove it from there or from the places where the keyword is called.

MarianMacik (approved):
Thanks!

mlassak (approved):

PR #2597, opened 2025-08-14
fix: Pass ENABLE_NEW_OBSERVABILITY_STACK to tests

MarianMacik:
After [!1301 (merged)](https://github.com/ods/jenkins/-/merge_requests/1301), which added ENABLE_NEW_OBSERVABILITY_STACK to the step when RHOAI operator is being deployed, we need to add it also to the test execution step. Otherwise if a DSCI is created during the tests, it will contain new observability stack every time.


MarianMacik:
Successful test run: http://10.0.98.56:8080/job/devops/job/rhoai-test-flow/3/robot/report/log.html#s1-s2-s1-s1-t1-k4-k1
You can see that the `false` value was correctly read from the job param.

CFSNM (approved):

PR #2624, opened 2025-09-08
Add ARM installation support for AWS OCP deployments.

sefroberg:


ods_ci/tasks/Resources/Provisioning/Hive/AWS/aws-cluster.yaml
```diff
@@ -34,12 +34,14 @@ items:
       baseDomain: {{ base_domain }}
       compute:
       - name: worker
+        architecture: {{ aws_worker_architecture|default("amd64") }}
```
CFSNM, line 37: I would pass the x86_64 arch as the default value

CFSNM (changes requested):

ods_ci/utils/scripts/arch_utils.py
```diff
@@ -0,0 +1,229 @@
...
+def validate_architecture_support(arch: str, provider: str = "aws") -> bool:
+    """Validate that architecture is supported"""
+    supported_archs = ["x86_64", "arm64", "aarch64"]
+    return arch.lower() in [a.lower() for a in supported_archs]
+
+
+# Architecture Detection Functions
+
+@lru_cache(maxsize=1)
+def get_node_architecture() -> str:
```
bdattoma, line 50: are you sure we would need both this function and `get_cluster_architecture_info` ? there's overlap

ods_ci/utils/scripts/ocm/templates/create-cluster.jinja
```diff
@@ -23,12 +23,20 @@
       "compute_machine_type": {
         "id": "{{ COMPUTE_MACHINE_TYPE }}"
       }
+      {% if ARCHITECTURE == "arm64" %}
```
bdattoma, line 26: what about aarch64? it is listed as a supported one earlier

CFSNM:
@sefroberg whats the purpose of having these 2 python scripts? what do we need them for?

ods_ci/utils/scripts/arch_utils.py
```diff
@@ -0,0 +1,229 @@
...
+    "arm64": {
+        "general": ["m6g.large", "m6g.xlarge", "m6g.2xlarge", "m6g.4xlarge"], 
+        "memory": ["r6g.large", "r6g.xlarge", "r6g.2xlarge"],
+        "gpu": ["g5g.xlarge", "g5g.2xlarge"],
+        "compute": ["c6g.large", "c6g.xlarge", "c6g.2xlarge"]
+    }
+}
+
+def get_compatible_instance_type(instance_type: str, target_arch: str) -> str:
+    """Map x86_64 instance types to ARM64 equivalents"""
```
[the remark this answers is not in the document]
sefroberg: @bdattoma The basic idea is for an x86_64 compute node, what would an ARM node be? This was generated as part of the vibe coding for this so take the GPU ARM with a grain of salt as I don't think we technical have access to them yet.

ods_ci/utils/scripts/arch_utils.py
sefroberg: While we have a 1:1 for cluster to node architecture today, OCP does support multi architecture clusters and this may come in handy in the future. I can remove it to only push up the bare min code at this time if you would like. Please let me know.

ods_ci/utils/scripts/arch_utils.py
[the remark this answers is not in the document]
bdattoma: > so take the GPU ARM with a grain of salt as I don't think we technical have access to them yet.


Why merging a such code then?

ods_ci/utils/scripts/arch_utils.py
bdattoma: If you want to have the support for multi arch in, it should be fine, but I'd say choose one of the 2 and use that one only. As of now, the code is using both functions in sequence basically, and one function's return already include the information returned by the other. Correct?

ods_ci/infrastructure_configuration.yaml.example
```diff
@@ -3,4 +3,16 @@ infrastructure_configurations:
...
+    # ARM64-compatible instance type mappings
+    aws_arch_instance_mappings:
+        x86_64:
+            worker_default: "m5.2xlarge"
+            master_default: "m5.2xlarge" 
+            gpu_default: "g4dn.xlarge"
+        arm64:
+            worker_default: "m6g.2xlarge"  # Graviton2
+            master_default: "m6g.2xlarge"  # Graviton2  
+            gpu_default: "g5g.xlarge"     # Graviton2 + T4G GPU
```
bdattoma, line 18: does the nvidia gpu work with this flavor as well?

bdattoma (approved):

ods_ci/infrastructure_configuration.yaml.example
bdattoma, line 18: this mapping is not being used anywhere right? If not, let's remove it from the example of the configuration file. 

If it is still worth having  a mapping, then we can put it in some readme or comment in the code. Wdyt?

ods_ci/tasks/Resources/Provisioning/Hive/AWS/aws-cluster.yaml
```diff
@@ -79,4 +81,4 @@ items:
     name: {{ image_set|default("rhods-openshift") }}
     namespace: {{ hive_claim_ns }}
   spec:
-    releaseImage: {{ openshift_image|default("quay.io/openshift-release-dev/ocp-release:4.15.25-x86_64") }}
+    releaseImage: {{ openshift_image|default("quay.io/openshift-release-dev/ocp-release:4.15.25-multi") }}
```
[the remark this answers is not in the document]
sefroberg: > Is there any disadvantage/advantage in using directly the corresponding arch image tag?

@bdattoma I think the OCP installer is more forgiving with the multi vs a specific tag. Plus in the future we will not need to hardcode anything for ppc or s390x

bdattoma (dismissed):

MarianMacik (dismissed):

CFSNM (dismissed):

ods_ci/tasks/Resources/Provisioning/Hive/AWS/aws-cluster.yaml
```diff
@@ -34,15 +34,17 @@ items:
...
-            type: {{ worker_node_instance_type|default("m5.2xlarge") }}
+            type: {{ worker_node_instance_type }}
         replicas: {{ worker_node_replicas|default("2") }}
       controlPlane:
         name: master
+        architecture: {{ aws_master_architecture|default("amd64") }}
         platform:
           aws:
-            type: {{ master_node_instance_type|default("m5.2xlarge") }}
+            type: {{ master_node_instance_type }}
```
MarianMacik, line 47: [infrastructure_configuration_variables.md](https://github.com/red-hat-data-services/ods-ci/pull/2624/files#diff-4bdf8f96d980a99f14a4d43a44a37387dca385ca2feef04f9048f9adf998b1bc) lists that `master_node_instance_type` and `worker_node_instance_type` have m5.xlarge as defaults. I think we can leave the defaults as they were before, or?

sefroberg: Thanks for catching that.

CFSNM (approved):

MarianMacik (approved):

PR #2636, opened 2025-09-19
conf: deprecate e2e kuberay tests

pawelpaszki:
## WHAT
conf: deprecate e2e kuberay tests

## WHY
kuberay e2e tests moved to shift left


pawelpaszki:
@sutaakar do I need to create a PR against any other branch? the containerised tests should work against 2.24+ (probably even 2.23)

sutaakar (approved):

sutaakar:
@pawelpaszki it is ok to have it on master branch only, older branches won't be ran

PR #2863, opened 2026-02-24
Add custom options to Helm installation method for ODH/RHOAI

mlassak:
JIRA ref: [RHOAIENG-49772](https://issues.redhat.com/browse/RHOAIENG-49772)

Includes:
- add support for custom catalog source for Helm install method
- add support for custom MODEL_REGISTRY namespace (operator, app, workbenches namespaces already covered)

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/catalogsource_template.yaml
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/catalogsource_template.yaml
@@ -1,0 +1,10 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: ${catalog_source_name}
+  namespace: ${namespace}
+spec:
+  sourceType: grpc
+  image: ${image_url}
+  displayName: ${catalog_source_name}
+  publisher: Custom Catalog
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -76,6 +76,8 @@ ${DEFAULT_APPLICATIONS_NAMESPACE_RHOAI}=    redhat-ods-applications
 ${DEFAULT_APPLICATIONS_NAMESPACE_ODH}=    opendatahub
 ${DEFAULT_WORKBENCHES_NAMESPACE_RHOAI}=    rhods-notebooks
 ${DEFAULT_WORKBENCHES_NAMESPACE_ODH}=    opendatahub
+${DEFAULT_MODEL_REGISTRY_NAMESPACE_RHOAI}=    rhoai-model-registries
+${DEFAULT_MODEL_REGISTRY_NAMESPACE_ODH}=    odh-model-registries
 ${CUSTOM_MANIFESTS}=    ${EMPTY}
 ${IS_NOT_PRESENT}=      1
 ${DSC_TEMPLATE}=    dsc_template.yml
@@ -146,7 +148,7 @@ Install RHODS
           END
       ELSE IF  "${TEST_ENV}" in "${SUPPORTED_TEST_ENV}" and "${INSTALL_TYPE}" == "Helm"
              Install RHOAI In Self Managed Cluster Using Helm  ${enable_new_observability_stack}
-             ...    ${GITOPS_REPO_BRANCH}    ${GITOPS_REPO_URL}
+             ...    ${GITOPS_REPO_BRANCH}    ${GITOPS_REPO_URL}    ${image_url}
       ELSE IF  "${TEST_ENV}" in "${SUPPORTED_TEST_ENV}" and "${INSTALL_TYPE}" == "OperatorHub"
           IF  "${is_upgrade}" == "False"
               ${file_path} =    Set Variable    tasks/Resources/RHODS_OLM/install/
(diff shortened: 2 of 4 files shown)


asanzgom:
## Issues to Address

  ### 1. Unused `${output}` variable (flagged by static analysis + CodeRabbit)                                                                       
   
  In `Create Custom CatalogSource If Not Present`, line ~537:                                                                                        
                  
  ```robot                                                                                                                                           
  ${rc}    ${output} =    Run And Return Rc And Output
  ...    oc get catalogsource ${catalog_source_name} -n ${namespace} --no-headers 2>/dev/null
```                                                                                                                                                                
`  ${output}` is never used and gets overwritten later. Should be `${_}` to signal it's intentionally discarded. 

### 2. Missing empty-value guard in Configure Custom Namespaces                                                                                        
   
  The new block:                                                                                                                                     
                  
  ```robot 
  IF  "${MODEL_REGISTRY_NAMESPACE}" != "${DEFAULT_MODEL_REGISTRY_NAMESPACE_RHOAI}" and "${MODEL_REGISTRY_NAMESPACE}" !=
  "${DEFAULT_MODEL_REGISTRY_NAMESPACE_ODH}"                                                                                                          
      Configure Custom Model Registry Namespace    ${MODEL_REGISTRY_NAMESPACE}
  END
```                                                                                                                                     
If `MODEL_REGISTRY_NAMESPACE` is defined but empty, both inequality checks pass, and it will try to create a namespace with an empty name. Should add `"${MODEL_REGISTRY_NAMESPACE}" != "${EMPTY}"` to the condition.

### 3. `image_url` parameter passed through Install RHODS, but origin unclear                                                                             

In the Install RHODS keyword (line ~150), ${image_url} is passed to Install RHOAI in Self Managed Cluster Using Helm:                              
```robot               
  Install RHOAI In Self Managed Cluster Using Helm  ${enable_new_observability_stack}                                                                
  ...    ${GITOPS_REPO_BRANCH}    ${GITOPS_REPO_URL}    ${image_url}                                                                                 
```                                                                                                      
Where does `${image_url}` come from in this scope? It's not in the arguments of Install RHODS, and I don't see it defined as a suite-level variable.  
If it's expected to come from the test variables file, it should be documented and have a default, or be fetched with Get Variable Value, like `CATALOG_SOURCE` is handled inside the Helm keyword.

### 4. Duplicate `catalog_source` guards could be merged                                                                                                 

The two `IF "${catalog_source}" != "${EMPTY}"` blocks (one for creating the CatalogSource, one for appending -s operator.*.olm.source) could be merged into a single block for readability. Inside that block, the `image_url` check would gate only the creation part.

### 5. RHOSi.resource hardcodes `MODEL_REGISTRY_NAMESPACE` for RHOAI only                                                                                
   
At line 128, Set Suite Variable ${MODEL_REGISTRY_NAMESPACE} rhoai-model-registries only runs in the RHOAI/odh-stable branch. The ODH branch (line 131+) doesn't set MODEL_REGISTRY_NAMESPACE, but the PR adds it as a required variable. This means ODH test runs will fail unless `MODEL_REGISTRY_NAMESPACE` is set externally. Need to add a Set Suite Variable ${MODEL_REGISTRY_NAMESPACE} odh-model-registries in the ODH branch of `Assign Vars According To Product`.

The two main issues to resolve before approving are:                                                                                                       
                  
  1. The ODH path in `RHOSi.resource` needs to set `MODEL_REGISTRY_NAMESPACE` — this will break ODH test runs as-is.                          
  2. The `image_url` origin needs to be clarified — it appears to be an undefined variable in the Install RHODS scope.

mlassak:
> 5. RHOSi.resource hardcodes MODEL_REGISTRY_NAMESPACE for RHOAI only
At line 128, Set Suite Variable ${MODEL_REGISTRY_NAMESPACE} rhoai-model-registries only runs in the RHOAI/odh-stable branch. The ODH branch (line 131+) doesn't set MODEL_REGISTRY_NAMESPACE, but the PR adds it as a required variable. This means ODH test runs will fail unless MODEL_REGISTRY_NAMESPACE is set externally. Need to add a Set Suite Variable ${MODEL_REGISTRY_NAMESPACE} odh-model-registries in the ODH branch of Assign Vars According To Product.

I don't think this is a concern, as this is already covered in `master` [here](https://github.com/mlassak/ods-ci/blob/a7a3507b1e1928a2ac85f6babbd65ba3f1a9311f/ods_ci/tests/Resources/RHOSi.resource#L140), both IF branches already do have `Set Suite Variable ${MODEL_REGISTRY_NAMESPACE} <ns>`

asanzgom:
> > 5. RHOSi.resource hardcodes MODEL_REGISTRY_NAMESPACE for RHOAI only
> >    At line 128, Set Suite Variable ${MODEL_REGISTRY_NAMESPACE} rhoai-model-registries only runs in the RHOAI/odh-stable branch. The ODH branch (line 131+) doesn't set MODEL_REGISTRY_NAMESPACE, but the PR adds it as a required variable. This means ODH test runs will fail unless MODEL_REGISTRY_NAMESPACE is set externally. Need to add a Set Suite Variable ${MODEL_REGISTRY_NAMESPACE} odh-model-registries in the ODH branch of Assign Vars According To Product.
> 
> I don't think this is a concern, as this is already covered in `master` [here](https://github.com/mlassak/ods-ci/blob/a7a3507b1e1928a2ac85f6babbd65ba3f1a9311f/ods_ci/tests/Resources/RHOSi.resource#L140), both IF branches already do have `Set Suite Variable ${MODEL_REGISTRY_NAMESPACE} <ns>`

Ok @mlassak, but the `image_url` parameter origin unclear issue is not addressed. 
Install RHODS receives `${image_url}` as a positional argument (line ~105), and the CLI path passes it through (line ~147). But the Helm path (line ~150) does not pass `${image_url}` — instead, the Helm keyword reads the global `${IMAGE_URL}` via Get Variable Value. This inconsistency remains.

asanzgom (approved):

asanzgom (approved):

CFSNM (approved):

PR #2908, opened 2026-04-09
fix: use RF7-compatible ELSE IF expression in RHOSi.resource

asanzgom:
## Problem

The `Assign Vars According To Product` keyword in `RHOSi.resource` uses an `ELSE IF` condition that is incompatible with Robot Framework 7:

```robot
ELSE IF    "${PRODUCT}" == "ODH" AND "${OPERATOR_NAME}" == "opendatahub-operator"
```

This breaks because:

1. **`AND` (uppercase)** is not a valid Python operator — RF7 evaluates `ELSE IF` conditions as Python expressions, which require lowercase `and`.
2. **`${VARIABLE}` syntax** causes RF7 to resolve variables *before* expression evaluation, producing invalid Python like `"ODH" == "ODH" AND "opendatahub-operator" == "opendatahub-operator"`.

Since this keyword runs in the **top-level suite setup**, the failure cascades to every test — causing 61 suite setup failures across the entire job.

## Fix

Switch to `$variable` syntax (deferred resolution) and lowercase `and`:

```diff
- ELSE IF    "${PRODUCT}" == "ODH" AND "${OPERATOR_NAME}" == "opendatahub-operator"
+ ELSE IF    $PRODUCT == "ODH" and $OPERATOR_NAME == "opendatahub-operator"
```

This follows the [RF7 expression evaluation guidelines](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#evaluating-expressions).

## Impact

Resolves all 61 `"Parent suite setup failed: Invalid ELSE IF condition"` errors seen in CI for ODH runs.


sefroberg (approved):
Claude analysis
  Approve this PR. The fix is:
  - Technically correct
  - Follows Robot Framework 7 best practices
  - Solves the 53 suite setup failures mentioned
  - Aligns with Python's expression evaluation requirements

ods_ci/tests/Resources/RHOSi.resource
```diff
@@ -128,7 +128,7 @@ Assign Vars According To Product
         Set Suite Variable  ${MODEL_REGISTRY_NAMESPACE}    rhoai-model-registries
         ${CSV_NAME} =   Catenate    SEPARATOR=   ${OPERATOR_NAME}.     ${RHODS_VERSION}
         Set Suite Variable      ${CSV_NAME}
-    ELSE IF    "${PRODUCT}" == "ODH" AND "${OPERATOR_NAME}" == "opendatahub-operator"
+    ELSE IF    $PRODUCT == "ODH" and $OPERATOR_NAME == "opendatahub-operator"
```
MarianMacik, line 132:
```suggestion
    ELSE IF    ${PRODUCT} == "ODH" and ${OPERATOR_NAME} == "opendatahub-operator"
```
just to be aligned with the first IF.

asanzgom: syntax must be `$PRODUCT` instead of `${PRODUCT}`, these changes are working:
rhoai-test-flow/11913/

MarianMacik: This part works as well:
https://github.com/red-hat-data-services/ods-ci/blob/4d657842ed08b68b338a5f4985c662abc3bf3e6f/ods_ci/tests/Resources/RHOSi.resource#L110

So maybe the only change needed is the `AND` operator?

MarianMacik (approved):

PR #2912, opened 2026-04-13
Fix SIGPIPE errors by piping /dev/urandom through cat instead of input redirection

cgoodfred:
The script was encountering a sigpipe failure due to the way the random credentials were being generated. This change matches the way random generation is done within `install_idp.sh`.


CFSNM (approved):

sefroberg (changes requested):
Still some issues. Shared with @CFSNM in slack

sefroberg (approved):

CFSNM:
@sefroberg without this PR, the master branch is broken, so what about merging this and handle the changes in a follow-up one?

PR #2929, opened 2026-04-23
fix: GPU operator channel sed pattern to match unquoted PLACEHOLDER

san7ket:
The sed pattern `s|channel: ".*"|` only matched quoted channel values, but gpu_install.yaml uses unquoted `channel: PLACEHOLDER`. This caused the subscription to be created with a non-existent channel, failing GPU operator installation on every run.

Changes:
- Fix sed to match any channel value (quoted or unquoted)
- Add fallback to 'stable' if defaultChannel lookup returns empty
- Add CSV name validation with available channels output on failure
- Add post-sed verification to catch PLACEHOLDER left behind
- Add logging for channel and CSV selection


yash2189 (approved):
LGTM and after verifying the logs can see `Using GPU Operator channel: v26.3`

crackcodecamp (approved):

PR #2937, opened 2026-05-01
add OpenShift Route for MaaS Gateway

SB159:
## Problem
On GCP/OpenStack clusters, the MaaS Gateway endpoint (maas.apps.<cluster>) is unreachable externally. The Gateway's backing Service only receives an internal ClusterIP, unlike AWS where an ELB automatically assigns an external IP.

## Fix
Add an OpenShift Route (TLS passthrough) to `configure_maas_gateway.sh` that exposes `maas.<cluster-domain>` through the cluster's default router. Safe on all platforms — on AWS it simply provides a redundant ingress path alongside the ELB.

## Validation
Tested on two OpenStack clusters:
- ods-qe-psi-10: Route created, `curl` returns 404 (reachable, no timeout)
- ods-qe-psi-11: Route created, `curl` returns 404 (reachable, no timeout)

cc -@aloganat


ods_ci/tasks/Resources/Gateway/configure_maas_gateway.sh
```diff
@@ -75,4 +75,35 @@ oc wait gateway -n "${INGRESS_NS}" maas-default-gateway \
...
+    kind: Service
+    name: ${SVC_NAME}
+    weight: 100
+  tls:
+    termination: passthrough
+    insecureEdgeTerminationPolicy: Redirect
+  wildcardPolicy: None
+EOF
+
 echo "MaaS Gateway configured successfully with domain: maas.${CLUSTER_DOMAIN}"
```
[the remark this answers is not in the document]
SB159: Done

ods_ci/tasks/Resources/Gateway/configure_maas_gateway.sh
```diff
@@ -75,4 +75,41 @@ oc wait gateway -n "${INGRESS_NS}" maas-default-gateway \
 --for=condition=Accepted --timeout=5m || \
 echo "Gateway may not expose 'Accepted' condition; check with: oc get gateway -n ${INGRESS_NS} maas-default-gateway"
 
+# On non-AWS platforms (GCP, OpenStack, etc.) the Gateway service only receives an
+# internal IP.  Create an OpenShift Route so maas.<cluster-domain> is reachable
+# externally through the cluster's default router.  The route is safe to apply on
+# all platforms — on AWS it simply provides an additional ingress path.
+SVC_NAME="maas-default-gateway-openshift-default"
```
asanzgom: **Hardcoded service name may be fragile (Medium)**

This service name is derived from the Gateway name (`maas-default-gateway`) and the GatewayClass (`openshift-default`), but it's a naming convention of the OpenShift Gateway controller — not a contract. If the controller's naming scheme changes in a future OpenShift version, this will silently break.

**Suggestion**: Consider deriving the service name dynamically, e.g.:                                                                                                                                                                                         `SVC_NAME=$(oc get svc -n "${INGRESS_NS}" -l gateway.networking.k8s.io/gateway-name=maas-default-gateway -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)`

This is non-blocking — the hardcoded name works today, and the `oc wait` guard will catch any mismatch — but worth considering for robustness.

ods_ci/tasks/Resources/Gateway/configure_maas_gateway.sh
```diff
@@ -75,4 +75,41 @@ oc wait gateway -n "${INGRESS_NS}" maas-default-gateway \
 --for=condition=Accepted --timeout=5m || \
 echo "Gateway may not expose 'Accepted' condition; check with: oc get gateway -n ${INGRESS_NS} maas-default-gateway"
```
asanzgom, line 76: **Existing script error handling is inconsistent (Pre-existing, not introduced by this PR):**

The existing code at lines 4-76 uses || echo "warning" patterns and bare oc apply calls without exit-code checks. The new code at lines 83-115 properly uses `if !` guards. This inconsistency is pre-existing and not this PR's responsibility, but worth noting for a future follow-up.

asanzgom (changes requested):

ods_ci/tasks/Resources/Gateway/configure_maas_gateway.sh
CFSNM: agree with this, I would address this suggestion even if its non blocking

CFSNM (changes requested):

ods_ci/tasks/Resources/Gateway/configure_maas_gateway.sh
```diff
@@ -75,4 +75,41 @@ oc wait gateway -n "${INGRESS_NS}" maas-default-gateway \
...
 
+# On non-AWS platforms (GCP, OpenStack, etc.) the Gateway service only receives an
```
bdattoma, line 78: is it ok to create the Route for all the providers, included AWS?

ods_ci/tasks/Resources/Gateway/configure_maas_gateway.sh
SB159: Done.

ods_ci/tasks/Resources/Gateway/configure_maas_gateway.sh
SB159: Yes it is ok. it just adds another way to reach the same service.

ods_ci/tasks/Resources/Gateway/configure_maas_gateway.sh
```diff
@@ -75,4 +75,50 @@ oc wait gateway -n "${INGRESS_NS}" maas-default-gateway \
...
+# all platforms — on AWS it simply provides an additional ingress path.
+echo "Looking up service for gateway maas-default-gateway in ${INGRESS_NS}..."
+SVC_NAME=$(oc get svc -n "${INGRESS_NS}" \
+  -l gateway.networking.k8s.io/gateway-name=maas-default-gateway \
+  -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)
+
+if [ -z "${SVC_NAME}" ]; then
+  echo "ERROR: No service found for gateway maas-default-gateway in ${INGRESS_NS}"
+  exit 1
+fi
```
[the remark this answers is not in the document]
SB159: Done

asanzgom (approved):

CFSNM (approved):

PR #2955, opened 2026-05-25
Add modelCache configuration support to DSC setup

mwaykole:
Configure modelCache in the kserve component of DataScienceCluster during RHOAI installation. Worker node names are dynamically fetched at runtime. The feature is controlled by ENABLE_MODEL_CACHE variable.

- Self-managed: inject modelCache via yq before oc apply
- Managed: patch DSC after operator creates it
- Add Is Model Cache Enabled keyword in Common.robot

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -386,6 +386,10 @@ Verify RHODS Installation
   ${kserve} =    Is Component Enabled    kserve    ${DSC_NAME}
   IF    "${kserve}" == "true"
     Configure Gateway API
+    ${enable_model_cache} =    Is Model Cache Enabled
+    IF    ${enable_model_cache} and "${cluster_type}" == "managed"
+        Patch DSC With Model Cache Config
+    END
     Wait For Deployment Replica To Be Ready    namespace=${APPLICATIONS_NAMESPACE}
     ...    label_selector=app=odh-model-controller    timeout=400s
     Wait For Deployment Replica To Be Ready    namespace=${APPLICATIONS_NAMESPACE}
@@ -855,6 +859,10 @@ Create DataScienceCluster CustomResource Using Test Variables
             IF    '${cmp}' == 'workbenches'
                 Run    sed -i'' -e 's/<workbenches_namespace>/${NOTEBOOKS_NAMESPACE}/' ${file_path}dsc_apply.yml
             END
+    END
+    ${enable_model_cache} =    Is Model Cache Enabled
+    IF    ${enable_model_cache}
+        Add Model Cache Config To DSC Yaml    ${file_path}dsc_apply.yml
     END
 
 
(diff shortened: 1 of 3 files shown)


ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -1539,6 +1547,56 @@ Configure Gateway API
...
+    [Documentation]    Patch the DataScienceCluster with modelCache config
+    ...    using dynamically fetched worker node names (used for managed clusters only)
+    [Arguments]    ${dsc_name}=${DSC_NAME}
+    ${node_names_json} =    Get Worker Node Names As JSON Array
+    Log To Console    Patching DSC ${dsc_name} with modelCache config (nodes: ${node_names_json})
+    ${rc}    ${output} =    Run And Return Rc And Output
+    ...    oc patch DataScienceCluster/${dsc_name} --type merge -p '{"spec":{"components":{"kserve":{"modelCache":{"cacheSize":"20Gi","managementState":"Managed","nodeNames":${node_names_json}}}}}}'    #robocop:disable
+    Log To Console    ${output}
+    Should Be Equal As Integers    ${rc}    0    msg=Error patching DSC with modelCache config: ${output}
+
```
bdattoma: could you try unifying the function to update the node names between Self-managed and Managed clusters? 
Also, what's the difference between the two which is making you using 2 different functions?

mwaykole: Good point! Unified into a single `Patch DSC With Model Cache Config` keyword that works for both self-managed and managed clusters.

The earlier split was because:
- **Self-managed**: DSC is created from a template via `oc apply`, so I was injecting modelCache into the YAML file (using `yq`) before apply
- **Managed**: Operator auto-creates the DSC, so the only option was `oc patch` after it exists

But since `Verify RHODS Installation` runs after the DSC exists in both paths, the `oc patch` approach works uniformly. Simplified to a single keyword now — removed the yq-based injection entirely.

bdattoma (dismissed):

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -1539,6 +1543,20 @@ Configure Gateway API
...
     Should Be Equal As Integers    ${rc}    0    msg=Error configuring Gateway for KServe
 
+Patch DSC With Model Cache Config
+    [Documentation]    Patch the DataScienceCluster with modelCache config using dynamically fetched worker node names
+    [Arguments]    ${dsc_name}=${DSC_NAME}
+    ${cmd} =    Set Variable
+    ...    oc get nodes -l node-role.kubernetes.io/worker= --no-headers -o custom-columns=":metadata.name" | jq -R . | jq -sc .    #robocop:disable
+    ${rc}    ${node_names_json} =    Run And Return Rc And Output    ${cmd}
+    Should Be Equal As Integers    ${rc}    0    msg=Failed to fetch worker node names
+    Should Not Be Empty    ${node_names_json}    msg=No worker nodes found in the cluster
```
sefroberg: Can this not work on Single Node Openshift? I think this might fail on a single node.

sefroberg:
@mwaykole you need an attached Jira for this PR.

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
```diff
@@ -1539,6 +1543,20 @@ Configure Gateway API
...
+    Log To Console    Patching DSC ${dsc_name} with modelCache config (nodes: ${node_names_json})
+    ${rc}    ${output} =    Run And Return Rc And Output
+    ...    oc patch DataScienceCluster/${dsc_name} --type merge -p '{"spec":{"components":{"kserve":{"modelCache":{"cacheSize":"20Gi","managementState":"Managed","nodeNames":${node_names_json}}}}}}'    #robocop:disable
```
sefroberg: Why hardcode to 20Gi? Could you use an environment variable?

sefroberg (dismissed):
Left some notes regarding lack of SNO support and hardcoded value.

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
mwaykole: not its for kserve feature  where we require to have some node for caching the model

ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
mwaykole: i think i can dec it to 10 but cannot use  environment variable because the size is given considering to model size

bdattoma (approved):

manosnoam (approved):

PR #2986, opened 2026-07-08
Provision MaaS postgres in new infrastructure namespace

jrhyness:
Changed default destination for the postgres database and maas-db-config secret.


jrhyness:
Implemented by https://github.com/red-hat-data-services/ods-ci/pull/2987 . Closing.

PR #2987, opened 2026-07-09
Fix MaaS prerequisites: add metrics config to DSCI and support infra namespace separation

FedeAlonso:
https://redhat.atlassian.net/browse/RHOAIENG-75591
https://redhat.atlassian.net/browse/RHOAIENG-75592

Fix MaaS prerequisites: add metrics config to DSCI and support infra namespace separation

  - Add metrics: {} to dsci_template.yml so the monitoring stack deploys
    and MonitoringStackAvailable becomes True (required by MaaS controller
    since PR https://github.com/opendatahub-io/models-as-a-service/pull/1088)
  - Update configure_maas_postgres.sh to place maas-db-config in the
    derived infra namespace (redhat-ai-gateway-infra) with cross-namespace
    DNS, matching MaaS controller logic from PR https://github.com/opendatahub-io/models-as-a-service/pull/1051
    
This has been tested in both RHOAI and ODH nightlies from July 9th:

RHOAI Before the fix - dashboard-e2e-tests/1773 
image

RHOAI with the fix - dashboard-e2e-tests/1777
image

Same with odh (runs 1774 and 1779)

### DIFF
--- a/ods_ci/tasks/Resources/Database/configure_maas_postgres.sh
+++ b/ods_ci/tasks/Resources/Database/configure_maas_postgres.sh
@@ -17,16 +17,33 @@ done
 APPS_NS="${APPS_NS:-redhat-ods-applications}"
 POSTGRES_IMAGE="registry.redhat.io/rhel9/postgresql-15@sha256:90ec347a35ab8a5d530c8d09f5347b13cc71df04f3b994bfa8b1a409b1171d59"
 
-# Ensure namespace exists
+# Derive MaaS infrastructure namespace (matches MaaS controller logic from PR #1051).
+# When INFRA_NAMESPACE=AUTO (default since 3.5), the controller expects maas-db-config
+# in a separate infra namespace. Postgres stays in APPS_NS; maas-db-config is copied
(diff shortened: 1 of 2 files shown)


ods_ci/tasks/Resources/Database/configure_maas_postgres.sh
```diff
@@ -64,13 +81,19 @@ stringData:
 EOF
 
 # 2. maas-db-config secret (DB_CONNECTION_URL key)
-DB_URL="postgresql://${PG_USER}:${PG_PASS}@postgres:5432/${PG_DB}?sslmode=disable"
+# Use cross-namespace DNS when postgres and maas-db-config are in different namespaces
+if [[ "${INFRA_NS}" != "${APPS_NS}" ]]; then
```
jrhyness, line 85: You could do this unconditionally.

jrhyness (approved):
LGTM.   

You could also deploy the db to the infra_ns if you wanted to.

jland-redhat (approved):
lgtm

PR #2996, opened 2026-07-17
Fix MaaS install verification and DSC tests to use infra namespace

jrhyness:
maas-api now deploys to a separate infrastructure namespace (e.g. redhat-ai-gateway-infra) but the Robot Framework tests were still looking for it in APPLICATIONS_NAMESPACE, causing timeouts.

Add a Detect MaaS Infra Namespace keyword that reads the controller's INFRA_NAMESPACE env var with a derivation fallback matching configure_maas_postgres.sh (PR #2987). Use it in Verify RHODS Installation and the modelsasservice DSC component lifecycle tests.

Also fix Components.resource to pass the namespace argument through in the Managed state path — previously only the Removed path forwarded it.


jrhyness:
@FedeAlonso , @jland-redhat  .  I don't know if this is an issue or not.  Just identified via code scan.

PR #3044, opened 2026-08-18
RHOAIENG-81144: Fix kale e2e test 

ada333:
This PR fixes Kale e2e robot test -> 

the Kale test failed when selecting a hardware profile because it clicked a hardware profile dropdown while it was disabled. Root cause: Kale had its own Create Kale Workbench/Select Kale Hardware Profile keywords that selected the hardware profile via a raw JS .click(), bypassing the disabled-state handling that the shared Select Hardware Profile keyword already does for Elyra.

Changes:

Removed the duplicated Create Kale Workbench / Select Kale Hardware Profile / Dismiss Open Patternfly Menus keywords and switched the test to call the shared Create Workbench keyword (same one Elyra uses), so hardware profile selection goes through the already-robust, disabled-state-aware code path.

Fixed Click Kale Compile And Run to use a real Selenium click instead of a synthetic JS .click(). The JS click doesn't reliably reach JupyterLab's React handlers, leaving the run stuck indefinitely on "Validating notebook...".

Fixed Open Kale Panel to check actual element visibility instead of mere DOM presence. JupyterLab keeps inactive sidebar panels mounted in the DOM (just hidden), so the old check always reported the panel as "open" even when another sidebar tab was covering it, leaving the Compile and Run button genuinely invisible/unclickable.

Fixed the "Unsaved Changes" dialog XPath, which looked for a <dialog> tag but JupyterLab renders it as a <div class="jp-Dialog">.
Removed now-unused variables (KALE_PANEL_ID, KALE_ENABLE_SWITCH_CLICK_XP, KALE_PIPELINE_NAME_INPUT_XP) and the now-dead Patch Workbench Notebook Image Tag Via CLI / version-skip logic in the test file, which are no longer needed.

Testing:
 Ran tests/Tests/0500__ide/0503__ide_kale.robot end-to-end against a cluster with RHOAI 3.5ea2
 confirmed that the Kale pipeline test now passes:
 
image

https://redhat.atlassian.net/browse/RHOAIENG-81144

### DIFF
--- a/ods_ci/tests/Resources/Page/ODH/JupyterHub/Kale.resource
+++ b/ods_ci/tests/Resources/Page/ODH/JupyterHub/Kale.resource
@@ -12,16 +12,12 @@ Library          Collections
 
 *** Variables ***
 ${KALE_EXTENSION_NAME} =    jupyterlab-kubeflow-kale
-${KALE_PANEL_ID} =    jupyterlab-kubeflow-kale/kubeflowDeployment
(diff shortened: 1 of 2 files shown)


jstourac (approved):
/lgtm
/approve

harshad16 (approved):

PR #3052, opened 2026-08-19
Wait for workbench image dropdown until ImageStreams finish loading

jstourac:
The dashboard renders a Skeleton until ImageStreams load, so the 5s default wait can miss the select on slow or disconnected clusters.

To avoid issues like this:
selenium-screenshot-114


harshad16 (approved):
Thanks for the fix
/lgtm

jiridanek (approved):

PR #3058, opened 2026-08-21
Revert "[release-3.4] Bump Red Hat Build of Kueue OperatorHub channel from stable-v1.3 to stable-v1.4."

san7ket:
Reverts red-hat-data-services/ods-ci#3055

---

Hold with this as it looks like the true root cause was actually a regression in the RHOAI product. See this jira for the details: https://redhat.atlassian.net/browse/RHOAIENG-87671

Marked as a draft for now.

---

This was merged as Red Hat Build of Kueue 1.4 causes problems for the trainer component: https://redhat.atlassian.net/browse/RHOAIENG-87965. In a slack discussion it was decided we want to keep tests with older version of kueue (1.3) for now as the fix won't be part of the RHOAI 3.4.4 release.


jstourac:
Hold with this as it looks like the true root cause was actually a regression in the RHOAI product. See this jira for the details: https://redhat.atlassian.net/browse/RHOAIENG-87671

Marked as a draft for now.

PR #3066, opened 2026-08-26
fix: Training Hub tests skip unnecessary Kueue setup

Fiona-Waters:
Following on from Training Hub test failures [here](https://jenkins-csb-rhods-opendatascience.dno.corp.redhat.com/job/devops/job/rhoai-test-flow/15052/#showFailuresLink) the following changes have been made.

Training Hub RayJob tests don't use Kueue scheduling, so they don't need to enable/disable the Kueue component in DataScienceCluster.

Created separate setup/teardown keywords for Training Hub tests:
- Prepare Training Hub Test Suite
- Teardown Training Hub Test Suite

These keywords download the test binary and set up RBAC, but skip the 'Set Component State kueue ...' calls that require DataScienceCluster permissions.

This fixes permission errors where test users (ldap-user2) were forbidden from accessing datascienceclusters resources.

The existing Prepare/Teardown DistributedWorkloads Integration Test Suite keywords remain unchanged for other tests (like MNIST Ray tests) that do use Kueue.

### DIFF
--- a/ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource
+++ b/ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource
@@ -138,6 +138,7 @@ Run Training Operator KFTO SDK Test
 
 Prepare DistributedWorkloads Integration Test Suite
     [Documentation]    Prepare DistributedWorkloads Integration Test Suite
+    [Arguments]    ${enable_kueue}=${TRUE}
     Log To Console    "Downloading compiled test binary ${ODH_BINARY_NAME}"
 
     ${result} =    Run Process    curl --location --silent --output ${ODH_BINARY_NAME} ${DISTRIBUTED_WORKLOADS_RELEASE_ASSETS}/${ODH_BINARY_NAME} && chmod +x ${ODH_BINARY_NAME}
@@ -151,33 +152,37 @@ Prepare DistributedWorkloads Integration Test Suite
     Log To Console    "Retrieving user tokens"
     ${common_user_token} =    Generate User Token    ${NOTEBOOK_USER_NAME}    ${NOTEBOOK_USER_PASSWORD}
     Set Suite Variable    ${NOTEBOOK_USER_TOKEN}   ${common_user_token}
-    Log To Console    "Log back as cluster admin"
-    Login To OCP Using API    ${OCP_ADMIN_USER.USERNAME}    ${OCP_ADMIN_USER.PASSWORD}
     RHOSi Setup
-    Setup RBAC for codeflare-sdk
-    # Set Kueue component to Unmanaged state as admin
-    Log To Console    "Setting Kueue component to Unmanaged state"
-    Set Component State    kueue    Unmanaged
-    Wait For DataScienceCluster CustomResource To Be Ready    timeout=600
+    IF    ${enable_kueue}
+        Setup RBAC for codeflare-sdk
+        # Set Kueue component to Unmanaged state as admin
+        Log To Console    "Setting Kueue component to Unmanaged state"
+        Set Component State    kueue    Unmanaged
+        Wait For DataScienceCluster CustomResource To Be Ready    timeout=600
+    END
 
 Teardown DistributedWorkloads Integration Test Suite
     [Documentation]    Teardown DistributedWorkloads Integration Test Suite
-    Log To Console    "Log back as cluster admin"
-    Login To OCP Using API    ${OCP_ADMIN_USER.USERNAME}    ${OCP_ADMIN_USER.PASSWORD}
+    [Arguments]    ${enable_kueue}=${TRUE}
     Log To Console    "Removing test binaries"
     Remove File        ${ODH_BINARY_NAME}
-    # Set Kueue component back to Removed state
-    Log To Console    "Setting Kueue component to Removed state"
-    Set Component State    kueue    Removed
-    Wait For DataScienceCluster CustomResource To Be Ready    timeout=600
-    Teardown RBAC for codeflare-sdk
+    IF    ${enable_kueue}
+        # Set Kueue component back to Removed state
+        Log To Console    "Setting Kueue component to Removed state"
+        Set Component State    kueue    Removed
+        Wait For DataScienceCluster CustomResource To Be Ready    timeout=600
+        Teardown RBAC for codeflare-sdk
+    END
 
 Generate User Token
     [Documentation]    Authenticate as a user and return user token.
+    ...    Uses a temporary kubeconfig to avoid clobbering the admin context.
     [Arguments]    ${username}    ${password}
-    Login To OCP Using API    ${username}    ${password}
-    ${rc}    ${out} =    Run And Return Rc And Output    oc whoami -t
+    ${temp_kubeconfig} =    Set Variable    /tmp/ods-ci-user-kubeconfig
+    Login To OCP Using API And Kubeconfig    ${username}    ${password}    ${temp_kubeconfig}
+    ${rc}    ${out} =    Run And Return Rc And Output    oc whoami -t --kubeconfig=${temp_kubeconfig}
     Should Be Equal As Integers    ${rc}    ${0}
+    Remove File    ${temp_kubeconfig}
     RETURN    ${out}
 
 Run DistributedWorkloads ODH Test
--- a/ods_ci/tests/Tests/0600__distributed_workloads/test-run-training-hub-rayjob-tests.robot
+++ b/ods_ci/tests/Tests/0600__distributed_workloads/test-run-training-hub-rayjob-tests.robot
@@ -1,7 +1,7 @@
 *** Settings ***
 Documentation     Training Hub RayJob Integration tests - https://github.com/opendatahub-io/distributed-workloads/tree/main/tests/odh
-Suite Setup       Prepare DistributedWorkloads Integration Test Suite
-Suite Teardown    Teardown DistributedWorkloads Integration Test Suite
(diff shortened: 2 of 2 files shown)


ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource
```diff
@@ -172,6 +172,34 @@ Teardown DistributedWorkloads Integration Test Suite
...
+        FAIL    Unable to retrieve odh compiled binary
+    END
+    Create Directory    %{WORKSPACE}/distributed-workloads-odh-logs
+    Log To Console    "Retrieving user tokens"
+    ${common_user_token} =    Generate User Token    ${NOTEBOOK_USER_NAME}    ${NOTEBOOK_USER_PASSWORD}
+    Set Suite Variable    ${NOTEBOOK_USER_TOKEN}   ${common_user_token}
+    Log To Console    "Log back as cluster admin"
+    Login To OCP Using API    ${OCP_ADMIN_USER.USERNAME}    ${OCP_ADMIN_USER.PASSWORD}
+    RHOSi Setup
+    Setup RBAC for codeflare-sdk
```
pawelpaszki: Please do not call `Setup RBAC for codeflare-sdk` (and the matching teardown) for Training Hub. That path is Kueue/codeflare-sdk specific.

Reuse `Prepare DistributedWorkloads Integration Test Suite` / `Teardown DistributedWorkloads Integration Test Suite` with Kueue gated off (e.g. `${enable_kueue}=${FALSE}`) instead of a copied keyword that still applies the codeflare-sdk RBAC bundle.

Suggestion only — not verified on a cluster.

ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource
```diff
@@ -172,6 +172,34 @@ Teardown DistributedWorkloads Integration Test Suite
...
+    Log To Console    ${result.stdout}
+    IF    ${result.rc} != 0
+        FAIL    Unable to retrieve odh compiled binary
+    END
+    Create Directory    %{WORKSPACE}/distributed-workloads-odh-logs
+    Log To Console    "Retrieving user tokens"
+    ${common_user_token} =    Generate User Token    ${NOTEBOOK_USER_NAME}    ${NOTEBOOK_USER_PASSWORD}
+    Set Suite Variable    ${NOTEBOOK_USER_TOKEN}   ${common_user_token}
+    Log To Console    "Log back as cluster admin"
+    Login To OCP Using API    ${OCP_ADMIN_USER.USERNAME}    ${OCP_ADMIN_USER.PASSWORD}
```
pawelpaszki: This is the 15052 failure: after `Generate User Token` overwrites the default kubeconfig as `ldap-user2`, `Login To OCP Using API` with an empty `${OCP_ADMIN_USER.USERNAME}` becomes `oc login … -u -p …` (`Username: -p`, 401). Job 15052 is kubeconfig/`kube:admin`, not admin user/password.

Please mint the notebook token against a side kubeconfig (`Login To OCP Using API And Kubeconfig`) so the admin context is never replaced. Then this re-login can go away.

Suggestion only — not verified on a cluster.

ods_ci/tests/Resources/Page/DistributedWorkloads/DistributedWorkloads.resource
```diff
@@ -172,6 +172,34 @@ Teardown DistributedWorkloads Integration Test Suite
     Wait For DataScienceCluster CustomResource To Be Ready    timeout=600
     Teardown RBAC for codeflare-sdk
 
+Prepare Training Hub Test Suite
```
pawelpaszki: This is a near-copy of `Prepare DistributedWorkloads Integration Test Suite`. Prefer an `${enable_kueue}` argument on the existing DW keywords over a second suite setup. Training Hub can then keep using the DW setup/teardown entrypoints with Kueue skipped, without duplicating the binary download / token / RHOSi steps (and without the Robocop too-many-keywords / line-too-long hits on the copied `curl` line).

Suggestion only — not verified on a cluster.

ods_ci/tests/Tests/0600__distributed_workloads/test-run-training-hub-rayjob-tests.robot
```diff
@@ -1,7 +1,7 @@
 *** Settings ***
 Documentation     Training Hub RayJob Integration tests - https://github.com/opendatahub-io/distributed-workloads/tree/main/tests/odh
-Suite Setup       Prepare DistributedWorkloads Integration Test Suite
-Suite Teardown    Teardown DistributedWorkloads Integration Test Suite
+Suite Setup       Prepare Training Hub Test Suite
```
pawelpaszki: Please keep the existing DW suite setup/teardown here, with Kueue disabled, rather than switching to new Training Hub keywords that still call codeflare-sdk RBAC.

For example:
```robot
Suite Setup       Prepare DistributedWorkloads Integration Test Suite    enable_kueue=${FALSE}
Suite Teardown    Teardown DistributedWorkloads Integration Test Suite    enable_kueue=${FALSE}
```

Suggestion only — not verified on a cluster.

pawelpaszki (changes requested):
## Request changes

Skipping Kueue for Training Hub is the right call and should stay in this PR. On its own it does not fix [job 15052](https://jenkins-csb-rhods-opendatascience.dno.corp.redhat.com/job/devops/job/rhoai-test-flow/15052/#showFailuresLink).

### What 15052 actually failed on

The four GPU tests never reached `Set Component State kueue`. Suite setup died earlier:

```
"Retrieving user tokens"          # oc login as ldap-user2 succeeded
"Log back as cluster admin"
Username: -p
Password: Login failed (401 Unauthorized)
Parent suite setup failed: 1 != 0
```

`Login To OCP Using API` runs `oc login ${OCP_API_URL} -u ${username} -p ${password}`. `Username: -p` means `${OCP_ADMIN_USER.USERNAME}` was empty, so `oc` treated `-p` as the username.

That matches this job: external cluster, kubeconfig as `kube:admin`, not admin user/password. `Generate User Token` does `oc login` into the default kubeconfig, which overwrites `kube:admin`. The re-login cannot restore it. Teardown then runs as `ldap-user2` (Kueue DSC patch forbidden, ~17 min DSC wait, RBAC deletes forbidden) and later suites (Feature Store, etc.) fail the same way.

The new Training Hub keywords still call `Generate User Token` + `Login To OCP Using API` as admin, so on this cluster they will still fail with `1 != 0` whether Kueue is skipped or not.

### Suggested direction (not tested)

Please treat the following as review suggestions only. I have not run these tests against a cluster. They need to be verified on a kubeconfig-based GPU cluster (same shape as 15052 / llmd) before merging.

1. **Keep skipping Kueue** for Training Hub: no `Set Component State kueue Unmanaged/Removed`, no DSC wait.

2. **Do not use `Setup RBAC for codeflare-sdk` / `Teardown RBAC for codeflare-sdk` on the Training Hub path.** That keyword is Kueue/codeflare-sdk oriented (`kueue-batch-user-role`, `test-user-kueue-admin`, etc.). Training Hub should reuse the existing Distributed Workloads prepare/teardown (or another existing DW helper), with Kueue gated off. Prefer an argument on the current DW keywords over cloning them:

```robot
Prepare DistributedWorkloads Integration Test Suite
    [Arguments]    ${enable_kueue}=${TRUE}
    ... download binary, generate token, RHOSi Setup ...
    IF    ${enable_kueue}
        Setup RBAC for codeflare-sdk
        Set Component State    kueue    Unmanaged
        Wait For DataScienceCluster CustomResource To Be Ready    timeout=600
    END
```

Training Hub:

```robot
Suite Setup       Prepare DistributedWorkloads Integration Test Suite    enable_kueue=${FALSE}
Suite Teardown    Teardown DistributedWorkloads Integration Test Suite    enable_kueue=${FALSE}
```

That avoids the duplicated `curl` keyword (and the Robocop hits) and keeps MNIST/Kueue tests unchanged.

3. **Fix `Generate User Token` so it does not clobber the admin kubeconfig.** Use a side kubeconfig (`Login To OCP Using API And Kubeconfig` already exists). After that, the “log back as cluster admin” step should not be required. This is the 15052 blocker.

4. Confirm on a real run that Training Hub still has whatever non-Kueue DW setup it needs (binary download, user token, RHOSi). Do not pull in the codeflare-sdk RBAC bundle to get there.

pawelpaszki:
Should be OK to merge from a code-review perspective, with a few nits (leftover admin `oc login` on the `enable_kueue=${TRUE}` path is now unnecessary and would still fail on a kubeconfig/`kube:admin` cluster if MNIST is un-skipped; hardcoded `/tmp/ods-ci-user-kubeconfig`).

This still needs verification on a GPU cluster of the same shape as [job 15052](https://jenkins-csb-rhods-opendatascience.dno.corp.redhat.com/job/devops/job/rhoai-test-flow/15052/) — these comments are review-only, the tests have not been run here. Please confirm suite setup no longer clobbers the admin kubeconfig and that the four Training Hub RayJob tests can still create RayJobs as the notebook user without the codeflare-sdk RBAC.

ChughShilpa:
@Fiona-Waters Did you run the tests locally or in jenkins after the fix ?

Fiona-Waters:
> @Fiona-Waters Did you run the tests locally or in jenkins after the fix ?

No I haven't had a chance to run them. I am hoping maybe your team can take over this fix. I've asked in the slack thread.

pawelpaszki (approved):

ChughShilpa (approved):

PR #2495, opened 2025-06-10
enhancement: add option to install new Observability dependent operators

CFSNM:
Related JIRA: https://issues.redhat.com/browse/RHOAIENG-26685

PR #3067, opened 2026-08-26
fix: wait for model-registry-operator using control-plane label

antowaddle:
## Summary
- RHOAI 3.6 install verification waits for pods with `app.kubernetes.io/part-of=model-registry-operator`. That label is on Deployment metadata only; the ODH overlay stamps `control-plane=model-registry-operator` on the controller Deployment, ReplicaSet, and pods.
- Dashboard E2E (e.g. `dashboard-e2e-tests` #2492) then fails with `Check If Pod Exists` after 400s even when DSC is Ready and `model-registry-operator-controller-manager` is running.
- Switch the install wait (and the ODH post-install MR checks) to `control-plane=model-registry-operator`, matching `0113__dsc_components.robot`.

## Test plan
- [x] Rebuild `components/dashboard/dashboard-e2e-tests` with `ODS_CI_GIT_REPO_BRANCH=fix/model-registry-operator-wait-label`
- [x] Validated on https://jenkins-csb-rhods-opendatascience.dno.corp.redhat.com/job/components/job/dashboard/job/dashboard-e2e-tests/2526/
- [x] Confirm install passes the model-registry wait instead of 6m40s of `status:FAIL`
- [x] Confirm later component waits (feast, ogx, dashboard) still run
- [x] Optional: `oc get pods -n redhat-ods-applications -l control-plane=model-registry-operator` on the test cluster

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
+++ b/ods_ci/tasks/Resources/RHODS_OLM/install/oc_install.robot
@@ -430,14 +430,14 @@ Verify RHODS Installation
   IF    "${modelregistry}" == "true"
     # 3.6+ deploys AI Hub (and catalog on some overlays) instead of model-registry-operator.
     # Wait for whichever controllers the operator actually created.
-    ${legacy_mr}=    Wait For Deployment Replica If Present    namespace=${APPLICATIONS_NAMESPACE}
+    ${legacy_mr} =    Wait For Deployment Replica If Present    namespace=${APPLICATIONS_NAMESPACE}
     ...    label_selector=control-plane=model-registry-operator    timeout=400s
-    ${aihub}=    Wait For Deployment Replica If Present    namespace=${APPLICATIONS_NAMESPACE}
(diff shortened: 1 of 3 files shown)


lugi0 (dismissed):
Looks ok to me, Model Registry / AI Hub now also deploys the catalog controller and ai hub controller pods, so maybe we want to add those checks as well here?
```
control-plane=catalog-controller-manager
control-plane=aihub-controller-manager
```

antowaddle:
@lugi0 Good catch — that matches what we saw on dashboard-e2e-tests #2495 after a clean 3.6 EA1 install.

3.6 `modelregistry:Managed` deploys AI Hub (`aihub-controller-manager`) instead of the old `model-registry-operator-controller-manager`. Catalog is present on some overlays and missing on others.

Pushed a follow-up that waits for whichever of these actually exist:
- `control-plane=model-registry-operator` (pre-3.6)
- `control-plane=aihub-controller-manager` (3.6+)
- `control-plane=catalog-controller-manager` (when the overlay creates it)

and fails only if none of them are found.

mwaykole (approved):

kobihk (approved):

PR #3072, opened 2026-09-02
fix: retry Hive webConsoleURL after cluster claim is running

kobihk:
## Summary
- Wait for Hive provision/claim conditions before checking ClusterDeployment webConsoleURL
- Retry the console URL for up to 5 minutes so a briefly empty status field does not fail a successful Hive install

## Test plan
- [x] Re-run Hive GCP cluster-pool provision (e.g. autotrigger-smoke bvt-rhoai-gcp-1)
- [x] Confirm Wait For Cluster To Be Ready succeeds when install completed and ClusterRunning=True
- [x] Confirm it still fails and dumps ClusterDeployment/ClusterClaim if webConsoleURL never appears

Related Jira: [RHOAIENG-90655](https://redhat.atlassian.net/browse/RHOAIENG-90655)

Signed-off-by: Kobi Hakimi <khakimi@redhat.com>
Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
+++ b/ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
@@ -170,6 +170,16 @@ Watch Hive Install Log
     END
     Sleep   10s    reason=Let's wait some seconds before proceeding with next checks
 
+Cluster Web Console URL Should Be Set
+    [Documentation]    Fails unless ClusterDeployment status.webConsoleURL is populated.
+    [Arguments]    ${namespace}    ${cd_name}
+    ${web_access} =    Run Process
+    ...    oc -n ${namespace} get cd ${cd_name} -o json | jq -r --exit-status '.status.webConsoleURL | select(type\=\="string" and length > 0)'    # robocop: disable:line-too-long
+    ...    shell=yes
+    Should Be Equal As Integers    ${web_access.rc}    ${0}
+    Should Not Be Empty    ${web_access.stdout}
+    RETURN    ${web_access}
+
 Wait For Cluster To Be Ready
     IF    ${use_cluster_pool}
         ${pool_namespace} =    Get Cluster Pool Namespace    ${pool_name}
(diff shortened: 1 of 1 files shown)


ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -204,7 +211,11 @@ Wait For Cluster To Be Ready
         ${custer_status} =    Run Process
         ...	oc -n ${hive_namespace} get clusterclaim ${claim_name} -o json | jq '.status.conditions[] | select(.type\=\="ClusterRunning" and (.reason\=\="Resuming" or .reason\=\="Running"))' --exit-status    shell=yes    # robocop: disable:line-too-long
     END
-    IF    ${provision_status.rc} != 0 or ${web_access.rc} != 0 or ${custer_status.rc} != 0
+    # Hive can take a short time after install/claim to populate status.webConsoleURL.
+    ${web_status}    ${web_access} =    Run Keyword And Ignore Error
```
bdattoma: why ignoring the error?

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
```diff
@@ -170,6 +170,16 @@ Watch Hive Install Log
...
     Sleep   10s    reason=Let's wait some seconds before proceeding with next checks
 
+Cluster Web Console URL Should Be Set
+    [Documentation]    Fails unless ClusterDeployment status.webConsoleURL is populated.
+    [Arguments]    ${namespace}    ${cd_name}
+    ${web_access} =    Run Process
+    ...    oc -n ${namespace} get cd ${cd_name} -o json | jq -r '.status.webConsoleURL' --exit-status
+    ...    shell=yes
+    Should Be Equal As Integers    ${web_access.rc}    ${0}
+    Should Not Be Empty    ${web_access.stdout}
```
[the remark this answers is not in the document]
manosnoam: @kobihk can you fix this as well ?

ods_ci/tasks/Resources/Provisioning/Hive/provision.robot
kobihk: It is only there so that a missing console URL does not fail the keyword immediately.

Wait Until Keyword Succeeds raises if webConsoleURL is still empty after 5 minutes. Without Run Keyword And Ignore Error, that timeout would abort here, and you would never reach the dump of ClusterDeployment and ClusterClaim that the original check used.

The ignore turns the wait into a status:

PASS → ${web_access} is the process result, including the URL
FAIL → fall through to the same IF as a failed provision/claim, dump the CRs, then FAIL with the existing message
So the error is not discarded. It is deferred so all three checks share one failure path. Wait Until Keyword Succeeds still retries the real check every 10s; Ignore Error only wraps that wait so a timeout is handled with the other Hive failures instead of as a raw Robot timeout.

Anyway, changed to:
    ${web_ready} =    Run Keyword And Return Status

bdattoma (approved):

manosnoam (approved):

PR #3079, opened 2026-09-15
RHOAIENG-94949: Fix Kale e2e test failing in disconnected environment

ada333:
PR for bug described in https://redhat.atlassian.net/browse/RHOAIENG-94949
This PR brings changes to the Kale test, adding an env var option of the github link - meant for disconnected cluster environment, where normal github cannot be reached.

This PR was tested on disconnected cluster, due to other issues with Kale in disconnected env (https://redhat.atlassian.net/browse/RHOAIENG-95600
https://redhat.atlassian.net/browse/RHOAIENG-95601) the test did not pass, but failed when running the pipeline, which verifies that this change works.


jesuino (approved):
LGTM! We still need to work on the image for the disconnected environment.

jstourac (approved):
/lgtm
/approve

thank you for the fix!

